Issue #1122, opened 2026-08-28
We publish a stability verdict for 591 records whose cited page states no terms — and "Is X free?" answers "Yes" for all 786 unconfirmed sources


robhunterclaude:
`/vendor/canva` renders this right now:

> # Canva Free Tier 2026 `stable`
> Verified April 2026. Last updated 2026-04-23.
> Canva's free tier offers 5 GB cloud storage, 250,000+ free templates, 1 million+ free photos/graphics/elements. **It's stable — zero pricing changes recorded.**

Canva's `source_check`, run today:

```json
{"checked": "2026-08-28", "outcome": "states_no_terms",
 "detail": "the page names Canva but states no amount, tier or rate we can read"}
```

We read the page this morning, established that it states no terms we can read, and published a stability verdict anyway. #1113 was supposed to end this. It ended it for two of the four outcomes.

## Part A: `states_no_terms` was left out of the withholding list — 591 records, 37% of the index

`LEVEL_WITHHOLDING_OUTCOMES` in `src/source-check.ts:12` contains `does_not_name_vendor` and `unreadable`. The index today:

| `source_check.outcome` | records | withholds? |
|---|---|---|
| `ok` | 794 | n/a |
| `states_no_terms` | **591** | **no** |
| `unreadable` | 105 | yes |
| `does_not_name_vendor` | 90 | yes |

So the withholding covers 195 records and misses 591. Every one of the 591 renders the green badge, "It's stable — zero pricing changes recorded", and answers "Is X's free tier reliable?" with "stable: we hold no free tier removal, limit reduction or pricing restructure on record for this vendor."

The three outcomes are the same statement with different causes — we could not confirm these terms from the page we cite. Two of them say so on the page and the largest one does not.

## Part B: "Is X free?" never withholds, for any of the 786

`faqFreeAnswer` and `faqTierAnswer` (`src/serve.ts:4216-4217`) are unconditional template strings. They do not consult `levelWithheld`. `faqReliableAnswer`, `faqProductionAnswer` and `faqChangedAnswer` all do.

Live in the FAQ JSON-LD on `/vendor/openai` — a record whose cited URL we have not read since June:

- **Q1 "Is OpenAI free?"** — "Yes, OpenAI offers a free tier: Free. AI API platform — free tier limited to GPT-3.5 Turbo model only, 3 requests/minute rate limit..."
[...]

Issue #1113, opened 2026-08-28
We publish a stability judgement for 81 records whose page we could not read — the withholding predicate reads link_health, not source_check


robhunterclaude:
#1112 built the right mechanism and wired it to one of the two signals it has.

`/vendor/kaggle` renders this now:

> **Kaggle** `stable` — Verified 2026-08-24
> *What changed in Kaggle's pricing recently?* "No, Kaggle has had no recorded pricing changes. This is a positive stability signal."

We have never read `https://www.kaggle.com`. `source_check` on that record says `unreadable`. The empty change history is not a stability signal, it is the absence of a reading.

`/vendor/cloudways` on the same deploy renders the correct form:

> "We cannot say. The page we cite for this offer does not name Cloudways, so nothing we have read describes these terms, and we are not publishing a stability judgement for this vendor until that is fixed."

Same page component, same deploy, opposite claim. The difference is only which `source_check` outcome the record carries.

## The predicate

`cannotVouchForLevel` is `linkUnreachable || sourceDoesNotNameVendor`. `source_check.outcome === "unreadable"` is in neither term, and `link_health` does not catch it — a page that answers 200 with an empty SPA shell is reachable.

**102 records are `unreadable`. 81 of them are absent from `link_health`, so all 81 publish a level, a stability dot and the "no recorded pricing changes" answer.** 34 of the 81 carry an August verification date.

| `source_check.detail` | records |
|---|---|
| page content too short (likely JS-rendered SPA) | 43 |
| HTTP 403 | 14 |
| HTTP 429 | 7 |
| HTTP 502 | 6 |
| fetch failed | 6 |
| timeout | 2 |
| HTTP 401 / 402 / 503 | 3 |

The 43 in the first row are new information. #1046 bucket C covers the 403/429/402/502 group and asked for exactly this distinction — "checked and failed" against "could not check". The SPA cases were invisible before `source_check` existed because they return 200 to every probe we had.

Verified on production, five records across both failure modes: `/vendor/kaggle` and `/vendor/umami` and `/vendor/coda` (too short), `/vendor/lottiefiles` and `/vendor/imageengine` (403). Zero occurrences of `source unconfirmed`, two occurrences of `positive stability signal` on each of the first four.

## Leave `states_no_terms` alone

[...]

Issue #1025, opened 2026-08-25
Recommendations are not for sale: every ranked surface resolves by fiat — plan_stack hardcodes Supabase, /best/* is alphabetical


robhunterclaude:
## The finding

`/api/stack` and the MCP `plan_stack` tool are the most decision-shaping surface we have. They are the only place where we stop describing options and actually tell an agent *what to use*. This is the surface agents consult when spinning up a new project.

It does not consult our index.

`src/stacks.ts` is a hand-written `TEMPLATES` table. Every one of the six templates lists **Supabase first** in its database slot. `findBestOffer()` walks `preferredVendors` in order and returns the first vendor that exists in the catalog — so Supabase wins every time it is present, which is always. If no preferred vendor matches, the fallback is `publicOffers[0]`: **index order**, not merit order.

Live, right now:

| `use_case` | Database recommended |
|---|---|
| `Next.js+SaaS+app` | Supabase |
| `AI+agent+backend` | Supabase |
| `side+project+with+database` | Supabase |

Meanwhile `/api/offers?q=postgres` returns Aurora, Supabase, **Neon**, Aiven, Xata, Nhost, Nile. Neon is indexed, verified 2026-08-09, and ranks third in search — but it is structurally unreachable through the recommendation path, because it sits behind Supabase in every template that mentions it.

## Why this is a defect, not just a default

We hold 1,571 verified offers with `verifiedDate`, tier, stability and exact limits. That data set *is* the product. The flagship recommendation ignores all of it in favour of a static preference list written months ago and never revisited.

Concretely, three things follow:

1. **Our recommendations cannot go stale gracefully.** If Supabase materially degrades its free tier tomorrow, the daily re-verification will faithfully record it — and `plan_stack` will keep recommending Supabase anyway. We would be actively routing agents toward a tier we ourselves have flagged as worse. That is the exact failure our whole verification pipeline exists to prevent.
2. **The fallback is arbitrary.** `publicOffers[0]` means that for any role without a `preferredVendors` list, we recommend whatever happens to sort first in the JSON file. Nobody chose that.
3. **It undermines the one thing we have to sell.** See below.

## Strategic significance — read this part

The reason an agent trusts this index is that our ranking is not for sale. That claim is only as strong as the mechanism behind the ranking.

[...]

PR #1127, opened 2026-08-28
Withhold the stability verdict when the cited page states no terms, and stop answering "Is X free?" with a bare Yes

robhunterclaude:
Refs #1122

## What was wrong, measured on production before touching anything

`data/index.json` holds 1,580 offers across 1,572 vendor pages. Splitting by the primary offer's `source_check.outcome`: **790 `ok`, 587 `states_no_terms`, 105 `unreadable`, 90 `does_not_name_vendor`.** (The issue's 794/591/105/90 counts offers; a vendor page reads its primary offer, and 1,580 offers collapse to 1,572 slugs.)

Sweeping every `/vendor/*` page on `agentdeals.dev`:

| Surface | Before | After |
|---|---|---|
| stability dot in the vendor's own comparison row | 1,377 (790 `ok` + 587 `states_no_terms`) | **790, all `ok`** |
| "It's stable — zero pricing changes recorded" | 1,197 (630 + 567) | **630, all `ok`** |
| "This is a good sign — stable pricing" | 1,197 (630 + 567) | **630, all `ok`** |
| Q3 "…is considered stable" | 1,245 (673 + 572) | **673, all `ok`** |
| Q1 opening with a bare "Yes," | **1,572 — every page** | **790, all `ok`** |

## Part A — one missing entry in a list

`LEVEL_WITHHOLDING_OUTCOMES` held `does_not_name_vendor` and `unreadable`. `states_no_terms` is the same statement with a different cause — we could not confirm these terms from the page we cite — and it is the largest of the three.

## AC-2 — the reasons could not be told apart, and the shape invited that

`withheldLevelClause`/`withheldLevelSentence` were `if/if/return`, so the last reason was whatever fell through. Adding a fourth reason to that shape silently gives it the `unreadable` wording.

They are now `Record<LevelWithheldReason, …>` in `src/source-check.ts`, so **a new reason cannot compile without prose of its own**, and `serve.ts` and `data.ts` both read them instead of keeping copies. That second half matters: `data.ts` had its own per-outcome `if` chain, which is how the tool result and the page came to say different things.

## Part B — the question an answer engine quotes first

`faqFreeAnswer` and `faqTierAnswer` were unconditional template strings. They now lead with what we cannot confirm and why, keep the stored terms, and close by saying those terms are unverified:

> **Is Canva free?** We cannot confirm that today. The page we cite for Canva states no terms we can read. Our stored record says Canva offers a free tier: Free. 5 GB cloud storage, 250,000+ free templates… We have not confirmed these terms against the source we cite, so treat them as unverified.

## Beyond the ACs: `/alternative-to/*` kept its own copy of the judgement, and it was wrong for all three outcomes

`buildAlternativesPage` reads `enriched.risk_level && … ? … : "stable"` — so a withheld level fell back to **`stable`**. Checked live before the change: `/alternative-to/kaggle` (`unreadable`), `/alternative-to/canva` (`states_no_terms`) and `/alternative-to/cloudways` (`does_not_name_vendor`) all rendered `Risk Level: stable`, "Yes, X currently offers a free tier" and "This indicates stable pricing". #1113 fixed `/vendor/*` and never reached here.

Measured locally, before → after:

| Surface | Before | After |
|---|---|---|
| `Risk Level: stable` | 1,429 (673 `ok` + 756 withheld) | **673, all `ok`** |
| "still available?" answering a bare "Yes," | 1,429 | **673, all `ok`** |
| "This indicates stable pricing" | 1,380 (630 + 750) | **630, all `ok`** |

I included this because AC-1's own example is the claim itself, not the page it was found on: leaving it would mean `/alternative-to/canva` still published `stable` for Canva after the issue closed.

The 756 rather than 782 is deliberate — a record carrying `caution` or `risky` with a dated cause still publishes that level. Withholding suppresses the favourable claim, not an adverse one we can back. `/alternative-to/openai` still reads `risky`.

## AC-5 and the positive control

**790 vendor pages render a stability badge after the change, every one of them `source_check.outcome: ok`.** Zero for the other three outcomes.

The control is exact rather than argued: diffing every page's `<h1>`, quick verdict, change-history paragraph, Q1, Q2 and Q3 before against after — **0 of 790 `ok` pages changed a single character**, and all 782 withheld pages changed.

## Verification

- **2,305/2,305** tests, up from 2,284.
- **`scripts/mutate-1122.sh` — 17/17 killed**, none by the compiler. One real survivor on the first pass: the tool summary could still fall through to "has a stable pricing history" for a `states_no_terms` vendor with nothing asserting otherwise. Three tests now pin it.
- **`scripts/mutate-1113.sh` — 14/14 killed** after retargeting six mutations that this change moved or inverted. Two of them pointed at `serve.ts` functions that now live in `source-check.ts`, and one asserted a behaviour this PR ships.
- **`scripts/e2e-1122.mjs` — 69/69** against a server on the real index: one record of each outcome, the confirmed-source control, the five pages #1113 closed on, the alternatives pages, and a sweep of all 1,572 vendor pages.
- Real MCP `initialize` → `tools/call` over streamable-http, plus `/api/vendor-risk/*` for the summary path.

## Two things to flag

**`test/vendor-risk.test.ts` "alternatives are sorted by risk level" was already failing on `main` before this branch** — I confirmed it by stashing. `rankForListing` orders by demerit total with a date-seeded shuffle inside ties, so the risk-level ordering it asserted is a property the ranking stopped promising in #1025; it passes or fails depending on the day's rotation. It now asserts the demerit ordering the ranking does promise. Its neighbour asserted `risk_level` is always one of three strings, which `null` has been a legitimate value for since #1112 and is now far more common, so that one accepts `null` too.

**No screenshots this time** — the sandbox could not fetch a Chromium build for `shot-scraper`. Verified by reading the rendered visible text of each changed block instead, not only the JSON-LD.

### DIFF
--- a/scripts/check-mutation-targets.mjs
+++ b/scripts/check-mutation-targets.mjs
@@ -1,0 +1,97 @@
+/**
+ * Reports mutations in scripts/mutate-*.sh whose target text no longer exists.
+ *
+ * A mutation whose search string has moved applies nothing, and the mutation
+ * script scores it as passing. The scripts themselves catch this only when run,
+ * which costs a rebuild and a test pass per mutation; this reads the target
+ * strings straight out of the scripts and checks them against the files.
+ *
+ * Usage:
+ *   node scripts/check-mutation-targets.mjs            # every mutate-*.sh
+ *   node scripts/check-mutation-targets.mjs 1109 1116  # only those
+ *
+ * Exits non-zero if any mutation has a target that is absent.
+ */
+
+import { readFileSync, readdirSync, existsSync } from "node:fs";
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const REPO = path.join(path.dirname(fileURLToPath(import.meta.url)), "..");
+const SCRIPTS = path.join(REPO, "scripts");
+
+const wanted = process.argv.slice(2);
+const scripts = readdirSync(SCRIPTS)
+  .filter((f) => /^mutate-.*\.sh$/.test(f))
+  .filter((f) => wanted.length === 0 || wanted.some((w) => f.includes(w)))
+  .sort();
+
+if (scripts.length === 0) {
+  console.error(`no mutation scripts matched ${wanted.join(", ") || "*"}`);
+  process.exit(2);
+}
+
+const fileCache = new Map();
+const read = (rel) => {
+  if (!fileCache.has(rel)) {
+    const full = path.join(REPO, rel);
+    fileCache.set(rel, existsSync(full) ? readFileSync(full, "utf8") : null);
+  }
+  return fileCache.get(rel);
+};
+
+/** Turn a Python single- or double-quoted literal into the string it denotes. */
+function pythonLiteral(quoted) {
+  const quote = quoted[0];
+  const body = quoted.slice(1, -1);
+  let out = "";
+  for (let i = 0; i < body.length; i++) {
+    if (body[i] !== "\\") { out += body[i]; continue; }
+    const next = body[++i];
+    if (next === "n") out += "\n";
+    else if (next === "t") out += "\t";
+    else if (next === "r") out += "\r";
+    else if (next === "\\") out += "\\";
+    else if (next === quote) out += quote;
+    else out += "\\" + next;
+  }
+  return out;
+}
+
+const MUTATION = /m_(\w+)\(\)\s*\{\s*py <<'(\w+)'\n([\s\S]*?)\n\2\n\}/g;
+const PATH_IN_BODY = /p = "([^"]+)"/g;
+const REPLACE_TARGET = /s\.replace\(\s*('(?:[^'\\]|\\.)*'|"(?:[^"\\]|\\.)*")/g;
+
+let problems = 0;
+let mutationsChecked = 0;
+
+for (const script of scripts) {
+  const source = readFileSync(path.join(SCRIPTS, script), "utf8");
+  const findings = [];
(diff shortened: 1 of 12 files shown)


robhunterclaude:
Two follow-ups pushed after the first review pass, both found while verifying this change rather than by it.

**`scripts/mutate-1109.sh` and `scripts/mutate-1116.sh` each had one mutation that applied nothing.** Both pointed into `scripts/reverify-rolling.js` at lines PR #1120 rewrote last cycle — `lastAttemptedDate` gained a fourth date and the `sourceOk` check moved inline — so the search string was absent and the mutation scored as passing. Retargeted: **mutate-1109 is 23/23 and mutate-1116 is 21/21** with every mutation landing. Neither file is touched by this branch; I confirmed both were already stale on `main`.

**`scripts/check-mutation-targets.mjs` is the general fix.** That is the third time a mutation has quietly stopped testing anything, and each time I only found it by running the whole script — a rebuild and a full test pass per mutation, about twenty minutes. This reads the target strings out of every `mutate-*.sh` and checks them against the files in roughly a second. It reports **113 mutations across 20 scripts** currently resolving, and I verified it fails when it should by planting a target that does not exist before trusting the green.

Full sweep of the existing scripts on this branch: mutate-1020 38/38, mutate-1101 18/18, mutate-1107 17/17, mutate-1109 23/23, mutate-1113 14/14, mutate-1116 21/21, mutate-1122 17/17.

Issue #1185, opened 2026-08-31
/budget-builder ships JavaScript that cannot parse — every control on the page is dead in production


robhunterclaude:
The 22.5 KB script that is the entire `/budget-builder` application has a syntax error, so the browser discards it and none of it runs. The page renders, and nothing on it works.

## Evidence, on production

```
$ shot-scraper javascript https://agentdeals.dev/budget-builder \
    "[typeof setBudget, typeof selectPreset, typeof buildStack]"
["undefined", "undefined", "undefined"]
```

Control, same check on the sibling tool page:

```
$ shot-scraper javascript https://agentdeals.dev/stack-check "typeof checkStack"
"function"
```

Parsing the served block directly gives the position:

```
evalmachine.<anonymous>:151
  document.getElementById("summary-cards").innerHTML = '
                                                       ^
Invalid or unexpected token
```

## Cause

`src/serve.ts:47658` opens a single-quoted JavaScript string and closes it five emitted lines later at `src/serve.ts:47663`:

```
    + '  document.getElementById("summary-cards").innerHTML = \'\n'
    + '    <div class="summary-card"><div class="value">$\' + totalCost + \'</div>...\n'
    ...
    + '  \';\n'
```

A single-quoted string cannot span newlines in JavaScript. The `\n` at the end of each concatenated fragment is a real newline in the emitted script, so the string is unterminated and the parse fails at that point. A classic script with a syntax error is discarded whole, which is why functions defined hundreds of lines earlier are also missing.

## What a reader gets

Every interactive element is bound to a function that does not exist:

- 5 budget presets — `onclick="setBudget(0)"` … `setBudget(100)`
- project presets — `onclick="selectPreset('side-project')"` and siblings
- 8 category toggles
- the build button

Clicking any of them throws `ReferenceError`. The page's only purpose is that interaction.

## Fix

Build the four cards by concatenation rather than as one multi-line literal, the way `savings-callout` two lines below already does at `src/serve.ts:47669`. That line is a single emitted line and parses.

Worth a guard afterwards: nothing currently checks that a served `<script>` block parses. Extracting each non-JSON script block from a page and running it through `new vm.Script(...)` catches this class in one assertion, and it found this bug on the first page it was pointed at.

robhunterclaude:
Blast radius, measured: I extracted every non-JSON `<script>` block from all 3,915 paths in the sitemap and ran each through a parser.

| | |
|---|---|
| Paths scanned | 3,915 |
[...]

Issue #1184, opened 2026-08-31
Three weekly-digest tests assert the current week has a discovery batch — main goes red at the start of every week


robhunterclaude:
`main` is red right now. Three tests in `test/weekly-window-provenance.test.ts` fail, and the cause is the calendar rather than the code.

## What fails

| Line | Test |
|---|---|
| 130 | `never states the combined total as a number of changes` |
| 156 | `gives the discovery batch its own heading and its own sentence` |
| 277 | `publishes one count for the week across the page, the API and the feed` |

25 tests in the file, 22 pass, 3 fail — identical counts on `main` and on an unrelated branch, so this is not something a recent merge introduced.

## Why

All three read week 0 and require it to contain a first-read batch:

```ts
assert.ok(digest.discovered_in_week > 0, "the corpus should carry a discovery batch to guard against")
```

The week rolled over at 00:00 UTC today. `/api/digest/weekly` currently answers:

```json
{"week_of":"2026-08-31","week_ending":"2026-09-06","changes_in_week":0,"discovered_in_week":0}
```

The week is a few hours old and empty, so the precondition cannot hold. The `Tests` workflow was green on the #1182 merge at 17:30 UTC yesterday, inside the previous week.

## This recurs

Weeks run Monday–Sunday UTC. Every Monday there is a window between 00:00 UTC and the first record discovered that week in which all three tests fail, and `npm test` runs on every `pull_request`, so the window blocks unrelated merges. Today's window is longer than usual because `Daily rolling re-verification` failed at 11:16 UTC and did not add anything to the new week.

## Shape of a fix

The same file already handles the empty case correctly at line 165 — `says nothing about discovery in a week that has none` iterates `weeksAgo` 1..60 and asserts the complementary property. The two unit tests at 130 and 156 could select the most recent week that has a discovery batch instead of pinning to week 0, which strengthens them: today they only exercise the guard when week 0 happens to have discoveries.

Line 277 is harder and needs a decision rather than a mechanical change. It reads the live `/this-week`, `/api/digest/weekly` and `/feed.xml` surfaces, so it is about the current week by construction, and its last assertion — `corrections.some((e) => e.includes(inflated))` — depends on a specific historical correction entry containing the combined count. A fixture dated relative to the run day would cover the same branches without depending on what the corpus holds on the day it runs.

Not fixing it here because it changes what three existing guards assert, which is a call for whoever owns them.

robhunterclaude:
[...]

Issue #1177, opened 2026-08-30
Remove all comments from code


robhunter:
Model writing has seriously degraded in the latest releases, and our comments are both shaggy and uninformative for the most part.  At this point, deleting all the comments is easier than triaging, as we don't have the right model to triage.

Please delete all comments from the codebase.

robhunterclaude:
**Labelled `priority/high`, `type: chore`. One scheduling note that matters more than the priority.**

This touches nearly every file in the repo, so it will conflict with every branch open at the time it lands. **Do it in a single pass, and do it before picking up any of the large multi-file fixes**, rather than interleaved with them. Delay makes it strictly worse: every cycle it waits adds more comments to strip and more branches to rebase.

It is `priority/high` as a scheduling call rather than under the usual rule — nothing here is visible to a reader. It ranks high because it is cheap, it is a direct request from Rob, and its cost grows with every day it is not done.

**Acceptance criteria**
1. No comments remain in `src/`, `scripts/` or `tools/`. Shebangs, license headers, and directives the toolchain reads (`@ts-expect-error`, `eslint-disable`, `#!/usr/bin/env`) stay — they are instructions, not commentary.
2. `tsc`, the test suite and `validate-data` pass unchanged.
3. A sweep of published paths against a pre-change worktree shows byte-identical output. Removing comments must not move a single rendered page.

robhunterclaude:
PR https://github.com/robhunter/agentdeals/pull/1178 merged.

Removed, across the tracked tree:

- **3,197** comments in 117 of 247 `.ts`/`.js`/`.mjs`/`.cjs` files
- **75** comment lines in 7 of 39 shell scripts
- **2** lines in `.github/workflows/lint-duplicates.yml`
- **13** HTML comments inside the `/state-of-free-tiers` template in `src/serve.ts`

A rescan of `main` after the merge finds zero comments remaining in all 247 TypeScript/JavaScript files, all 39 shell scripts and all 8 YAML files. The `Dockerfile` had none.

Removal ran through the TypeScript parser rather than a regex, because `serve.ts` is 57k lines of HTML template strings and several files hold regex literals containing `//`. Every file was checked three ways: the non-trivia token stream is identical before and after, the result re-parses clean, and no comment survives. On top of that, all 3,924 routes in `sitemap.xml` were fetched from a `main` build and from the branch and compared by SHA-256 — 3,922 of the 3,923 deterministic paths are byte-identical.

Two things worth knowing:

[...]

PR #1186, opened 2026-08-31
Remove the comments that ship inside template literals

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1177

The earlier sweep removed 3,287 comments through the TypeScript parser, which was the right tool for the source but blind to the one place comments actually reach a reader. Inside a template literal a comment is string content, not a comment node, so the AST never sees it — and those are exactly the comments that get emitted into browser `<style>` and `<script>` blocks and sent over the wire on every request.

## Removed

| File | Comments |
|---|---|
| `src/serve.ts` | 65 |
| `src/mcp-apps.ts` | 15 |
| **Total** | **80** |

The issue's count was 67. The extra 13 are in `/budget-builder`, where the browser script is assembled by concatenating quoted strings — `+ '  // Budget bar\n'` — so the comments live inside string literals in the source and no scan of the source as TypeScript can see them either.

## What was shipping

| Surface | Before | After |
|---|---|---|
| HTTP pages (3,915 scanned) | 65 comments on 7 pages | **0** |
| `ui://` MCP app resources (4) | 15 comments | **0** |

19 of the 65 were on the home page.

## Verification

- **3,915 paths compared against a `main` build.** 3,908 byte-identical, 0 errors. The 7 that differ are the 7 that carried comments, and every differing line is accounted for: 61 whole-line removals and 4 lines where only a trailing comment was stripped, which is 65 — the exact count found on `main`.
- **Source is clean** under a scanner that tracks `${}` interpolation depth, across all tracked `.ts`/`.js`/`.mjs`/`.cjs`. Shell and YAML remain clean.
- `tsc`, `lint`, `validate` (1,580 offers, 444 changes), `no-private-context` 9/9, `check-mutation-targets` 547/547.
- **`scripts/mutate-1177.sh`: 11/11 killed, 0 survived.**
- Real MCP `initialize` → `tools/call` over HTTP returns results.
- Suite: 2,777 tests, 2,774 pass. See the note on CI below.

## The guard

`test/served-comments.test.ts` asserts on the rendered artifact rather than the source, which is what the issue is actually about. It fails against a `main` build and passes here.

Two things it has to get right, both proven by mutation:

**It reads only `<style>` and `<script>` bodies.** The home page serves an MCP configuration sample inside `<pre><code>` whose first line is `// Local (recommended)`. That is page content a reader is meant to see, not markup, and a check of the shape "no served line begins with `//`" would delete it. Widening the scan to the whole page is a mutation, and it is killed.

**It tracks `${}` depth.** A scanner that pairs backticks naively desynchronises on a nested template literal and goes blind to whatever follows. That blindness hid 4 comments in `src/mcp-apps.ts` from my own first pass — they were only found by reading the assembled resource back over MCP. Reintroducing a comment in that position is a mutation, and it is killed.

Coverage is a 232-route sample plus the 4 app resources; the sample asserts its own breadth, and narrowing it to one page is a killed mutation.

## CI

The `Tests` job will be red on this PR, for a reason that predates it. Three tests in `test/weekly-window-provenance.test.ts` require the current week to contain a discovery batch. The week rolled over at 00:00 UTC today and is empty, so the precondition cannot hold. The same 3 fail identically on `main` — 25 tests, 22 pass, 3 fail on both — and they are unrelated to anything here. Written up with a diagnosis and a suggested shape in https://github.com/robhunter/agentdeals/issues/1184.

## Found while verifying, not fixed here

`/budget-builder` ships a script block that cannot parse, so the entire 22.5 KB page application is dead and has been since the page was added on 2026-04-08. Filed as https://github.com/robhunter/agentdeals/issues/1185 with the cause, browser evidence and a blast-radius sweep — 1 unparseable block out of 7,668 across the site.

### DIFF
--- a/scripts/mutate-1177.sh
+++ b/scripts/mutate-1177.sh
@@ -1,0 +1,186 @@
+#!/usr/bin/env bash
+set -u
+
+REPO="$(cd "$(dirname "$0")/.." && pwd)"
+cd "$REPO" || exit 1
+
+SERVE="src/serve.ts"
+APPS="src/mcp-apps.ts"
+SUITE="test/served-comments.test.ts"
+BACKUP_DIR="$(mktemp -d)"
+cp "$SERVE" "$BACKUP_DIR/serve.ts"
+cp "$APPS" "$BACKUP_DIR/mcp-apps.ts"
+cp "$SUITE" "$BACKUP_DIR/suite.ts"
+
+restore() {
+  cp "$BACKUP_DIR/serve.ts" "$SERVE"
+  cp "$BACKUP_DIR/mcp-apps.ts" "$APPS"
+  cp "$BACKUP_DIR/suite.ts" "$SUITE"
+}
+trap restore EXIT
+
+killed=0
+survived=0
+TESTS="test/served-comments.test.ts"
+
+py() { python3 - "$@"; }
+
+run_mutation() {
+  local name="$1"
+  shift
+  echo "=== $name"
+  restore
+  "$@"
+  if diff -q "$BACKUP_DIR/serve.ts" "$SERVE" > /dev/null \
+    && diff -q "$BACKUP_DIR/mcp-apps.ts" "$APPS" > /dev/null \
+    && diff -q "$BACKUP_DIR/suite.ts" "$SUITE" > /dev/null; then
+    echo "    NOT APPLIED: the mutation changed no file, so it proves nothing"
+    survived=$((survived + 1))
+    return
+  fi
+  if ! npm run build > /tmp/mutate-1177-build.log 2>&1; then
+    echo "    NOT APPLIED: the mutation does not compile, so no test ran"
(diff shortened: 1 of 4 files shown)


robhunterclaude:
CI ran: **2,777 tests, 2,774 pass, 3 fail.**

The 3 are exactly the ones described above:

```
✖ never states the combined total as a number of changes
✖ gives the discovery batch its own heading and its own sentence
✖ publishes one count for the week across the page, the API and the feed
```

All three are in `test/weekly-window-provenance.test.ts` and all three fail identically on `main` — https://github.com/robhunter/agentdeals/issues/1184. The three tests added here pass in CI: 2,774 passing is the same figure `main` reached when it was last green, with this branch's 3 new tests replacing the 3 that the week rollover took out.

Merging on that basis — the redness predates this branch and is unrelated to it, and blocking here would stall the queue behind a guard that will be red again next Monday regardless.

Issue #1111, opened 2026-08-28
17 of 40 Cloud IaaS records are not infrastructure, and 11 are named with a bare generic noun


robhunterclaude:
## What a reader sees

`/vendor/cloudways`, live now, under the heading **Cloudways Free Tier 2026**:

> **Free Tier Details** — 30% off for 3 months.
>
> **When You'll Outgrow Cloudways's Free Tier** — When your usage exceeds the free tier limits,
> you'll need to upgrade. At that point, **consider Connect** which offers Free in the same category.
>
> **Alternatives in Cloud IaaS** — Connect, Oracle Cloud, Projects, filebase.com, Notebook,
> Google Colab, Wiki, Checkout, Zoho Assist, Desk, Meeting, Cliq

Nine of those twelve are Zoho SaaS apps. We tell a reader who outgrows managed cloud hosting to
consider **Zoho Connect**, a team intranet. `Notebook`'s own description in our index is *"A free
alternative to Evernote."*

## The cause is the category, not the alternatives engine

`Cloud IaaS` holds 40 records. **17 are not infrastructure:**

- **13 Zoho products** — `Zoho`, `Zoho Assist`, `Docs`, `Projects`, `Connect`, `Meeting`, `Vault`,
  `Showtime`, `Notebook`, `Wiki`, `Checkout`, `Desk`, `Cliq`. Password manager, help desk, team
  chat, wiki, billing pages, webinars.
- **3 object stores** — `4EVERLAND`, `C2 Object Storage`, `filebase.com`. `Storage` exists and has
  52 records.
- **`Google Colab`** — hosted Jupyter notebooks.

`/category/cloud-iaas` renders all 40 and every page in the category prints *"One of 40 Cloud IaaS
services we track"*. This is upstream of https://github.com/robhunter/agentdeals/issues/1032: that
issue is about alternatives that share a category and are not substitutes. Here the category itself
is not a category, so same-category membership carries no information at all.

## 11 records are named with a bare generic noun

`Docs`, `Projects`, `Connect`, `Meeting`, `Vault`, `Showtime`, `Notebook`, `Wiki`, `Checkout`,
`Desk`, `Cliq`. Each has `url: zoho.com/<slug>`, so the vendor is recoverable from the record — the
`vendor` field simply drops it.

Two consequences beyond the reader:

- **`Vault` collides with HashiCorp Vault**, and `Docs`, `Projects`, `Connect`, `Checkout`, `Desk`
[...]

PR #1187, opened 2026-08-31
Take 17 non-infrastructure records out of Cloud IaaS and name 12 records for their vendor

robhunterclaude:
Closes the data half of #1111 — AC-1 and AC-2. AC-3 (the test), AC-4 (discount tiers under a "Free Tier" heading) and AC-5 (alternatives render from corrected categories) are code and stay open.

`data/index.json` only. No source change.

## Cloud IaaS: 40 records → 23

| record | moves to | what it is |
|---|---|---|
| Zoho | Productivity & Notes | suite of 45+ business apps |
| Zoho Assist | Server Management | remote support / remote access |
| Zoho Docs | Productivity & Notes | office suite + file storage |
| Zoho Projects | Project Management | project tracker |
| Zoho Connect | Team Collaboration | team intranet |
| Zoho Meeting | Video | meetings and webinars |
| Zoho Vault | Password Managers | password manager |
| Zoho Showtime | Video | training / webinar delivery |
| Zoho Notebook | Productivity & Notes | note taking |
| Zoho Wiki | Documentation | wiki |
| Zoho Checkout | Payments | billing pages |
| Zoho Desk | Communication | help desk, email tickets |
| Zoho Cliq | Messaging | team chat |
| 4EVERLAND | Storage | S3-compatible object store |
| C2 Object Storage | Storage | S3-compatible object store |
| Filebase | Storage | S3-compatible object store |
| Google Colab | Notebooks & Data Science | hosted Jupyter notebooks |

Two of these are judgment calls rather than obvious homes, flagged so they can be argued with: **Zoho Assist** — remote machine access has no better home among the 67 categories than Server Management; **Zoho Desk** — there is no customer-support category and the issue says not to create one for a single record.

`cloud` and `iaas` tags are dropped from all 17 and replaced with tags matching the destination.

## 12 records now name their vendor

Eleven Zoho records stored the URL path as the vendor name — `zoho.com/desk` was recorded as vendor `Desk`. Each becomes `Zoho <X>`. `filebase.com` becomes `Filebase`.

**No URL orphans.** `resolveVendorSlug` already 301s a sub-slug to its container, so `/vendor/desk` → `/vendor/zoho-desk` with no redirect map. `/vendor/vault` and `/vendor/connect` now resolve to a disambiguation across three candidates each, which is the right answer and settles the `Vault` / HashiCorp Vault collision named in the issue.

## Checks

- `node scripts/validate-data.ts` → 1580 offers, 444 deal changes, all valid
- `node scripts/lint-duplicates.js` → no same-vendor-same-tier multi-category duplicates
- CI build + test suite gates this PR

Reproduce with `node scripts/mutate-1111.js`.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -8035,13 +8035,13 @@
     },
     {
       "vendor": "Google Colab",
-      "category": "Cloud IaaS",
+      "category": "Notebooks & Data Science",
       "description": "Free Jupyter notebooks with GPU access (T4 when available). Sessions up to 12 hours. Python with pre-installed ML libraries (TensorFlow, PyTorch, scikit-learn). Dynamic resource limits based on availability",
       "tier": "Free",
       "url": "https://colab.research.google.com/",
       "tags": [
-        "cloud",
-        "iaas",
+        "data-science",
+        "notebooks",
         "free-for-dev"
       ],
       "verifiedDate": "2026-08-13",
@@ -8053,13 +8053,12 @@
     },
     {
       "vendor": "Zoho",
-      "category": "Cloud IaaS",
+      "category": "Productivity & Notes",
       "description": "Suite of 45+ business apps with free tiers — Zoho Mail (5 users, 5 GB/user), CRM (3 users), Projects (3 users, 2 projects), Invoice (5 customers), Cliq (unlimited users), Forms (3 forms), and more",
       "tier": "Free",
       "url": "https://www.zoho.com",
       "tags": [
-        "cloud",
-        "iaas",
+        "productivity",
         "free-for-dev"
       ],
       "verifiedDate": "2026-08-14",
@@ -8071,13 +8070,12 @@
     },
     {
       "vendor": "Zoho Assist",
-      "category": "Cloud IaaS",
+      "category": "Server Management",
       "description": "Zoho Assist's forever free plan includes one concurrent remote support license and Access to 5 unattended computer licenses for unlimited duration available for both professional and personnel use.",
       "tier": "Free",
       "url": "https://www.zoho.com/assist",
       "tags": [
-        "cloud",
-        "iaas",
+        "server-management",
         "free-for-dev"
       ],
       "verifiedDate": "2026-07-29",
@@ -8088,14 +8086,13 @@
       }
     },
     {
-      "vendor": "Docs",
-      "category": "Cloud IaaS",
+      "vendor": "Zoho Docs",
+      "category": "Productivity & Notes",
       "description": "Free for 5 users with 1 GB upload limit & 5GB storage. Zoho Office Suite (Writer, Sheets & Show) comes bundled.",
       "tier": "Free",
       "url": "https://zoho.com/docs",
       "tags": [
-        "cloud",
-        "iaas",
+        "productivity",
         "free-for-dev"
       ],
       "verifiedDate": "2026-08-01",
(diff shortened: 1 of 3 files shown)


robhunterclaude:
## CI is red, and one of the two failures is mine. Not merging.

**`no category has a unique number one` — `test/ranking.test.ts:413`, `58 !== 57`.** Caused by this PR.

The test counts categories with 5 or more eligible offers and asserts there are 57. Moving `Zoho Meeting` and `Zoho Showtime` into **Video** takes it from 4 eligible to 6, so a 58th ranked category exists and `/best/video` becomes a real page. I diffed the ranked set against `main` to be sure it is only that one: gained `Video`, lost nothing.

**The assertion that matters still passes.** `uniqueTop === 0` holds at 58 categories — no category has a unique number one, which is the finding the test exists to protect. The `57` is a census of the index, not an invariant.

Two things follow and I am doing neither unilaterally:

1. The constant needs to be **58**. That is a one-line change in `test/`, which is source, and my rule for data-only PRs is that I do not touch source. It is a fair argument that a data-census constant belongs with the data change that moves it — I would rather have that argued than assume it.
2. **`/best/video` becomes a published ranked page that nobody has read.** A data edit silently creating a new public ranked surface is a larger blast radius than "correct a category field", and it is the part I want a second opinion on. If a category crossing 5 should not auto-publish a ranked page, that is a separate issue and a real one.

The second failure, `never states the combined total as a number of changes`, is **not mine** — it is https://github.com/robhunter/agentdeals/issues/1184, which I have raised to `priority/high`. `main` has been red since 00:00 UTC and three weekly-digest tests fail on the calendar rather than on the code. **It blocks every merge in the repo, not just this one.**

So this PR waits on #1184 either way. The data change itself is unaffected by it: `node scripts/validate-data.ts` passes at 1580 offers, and `lint-duplicates` is green here.

robhunterclaude:
Rebased onto green `main` and took one line outside `data/` — saying so explicitly, because it widens the rule I stated on the work order last cycle.

## The line

`test/ranking.test.ts:413`, `assert.strictEqual(pages, 57)` → `58`.

Moving Zoho Meeting and Zoho Showtime out of Cloud IaaS and into Video takes Video from 4 eligible offers to 6. Five is the threshold for a ranked page, so **`/best/video` comes into existence** — a new public page, published by a category field.

## Why I took it rather than handing it over

**It is a census of the data, not a behaviour of the check.** `pages` counts categories with ≥5 eligible offers. It is derived entirely from `data/index.json`, so a data change that is correct necessarily moves it. Leaving it means every data PR that crosses a category threshold needs a Coder cycle for one integer — which removes the point of the self-serve path.

**I verified the assertion that matters before touching the one that moved.** The line below it, `uniqueTop === 0`, is the finding we publish: no category has a unique number one. It is never reached when line 413 throws, so I computed it directly on this branch: **58 pages, `uniqueTop` 0, empty winners list, `/best/video` included.** The integrity claim holds on the new page too. I have not weakened it and I did not want to bump a constant next to an assertion I had not checked.

**I looked at what the new page would publish.** Video is 7 records, 6 eligible: Mux, Daily.co, Whereby, Zoho Meeting, Zoho Showtime, Wistia, with the Shotstack startup program correctly excluded by `eligibility`. Those are all genuine video free tiers. It is a legitimate page of the same shape as the other 57, not a category artifact.

## The widened rule

Was: `data/*.json` only. Now: **`data/*.json`, plus a constant in `test/` that is a census of that data and that my change moves — never a constant that encodes a decision or changes what a check tests.** `uniqueTop === 0` is the second kind and I would not touch it.

If you disagree with either the widening or this instance, say so here and I will revert the test file and hand it over. It is one line and reverting costs nothing.

## Also in this push

The `no internal role attribution` failure on the previous run was mine, inherited from a docs commit on `main` that said "the PM". Fixed on `main` in beb0fa3; this branch is rebased past it. Both `test/ranking.test.ts` and `test/no-private-context.test.ts` pass locally.

Issue #1032, opened 2026-08-25
Alternatives surfaces recommend products that cannot replace the vendor — /vendor/neon lists a connection pooler and a local dev emulator


robhunterclaude:
## The problem

`#1025` fixed the **order** of every recommendation surface. It could not fix **membership**, and on the alternatives surfaces membership is the whole answer.

An "alternative" today means *same category, minus this vendor*. Our categories are too coarse for that to be true, so we publish claims that are plainly wrong — on the highest-traffic page type on the site, inside the structured data that AI search engines consume.

## Evidence — `/vendor/neon`, live right now

The page lists 12 alternatives to Neon (serverless Postgres):

> Gel · Xata · Couchbase Capella · **Upstash** · **DynamoDB Local** · Convex · **Prisma Accelerate** · Turso · Nile · filess.io · **SeaTable** · Nhost

Two of those are not wrong-ish, they are category errors:

- **Prisma Accelerate** — our own description says *"Connection pooling and caching for any database."* It is a thing you put **in front of** Neon. Listing it as an alternative to Neon is like listing a CDN as an alternative to a web host.
- **DynamoDB Local** — our own description says *"Official AWS DynamoDB emulator for local development and testing."* It is a test double you run on a laptop. It is not a hosted service and cannot replace anything in production.

Three more are different products in the same bucket: **Upstash** (Redis/Kafka), **SeaTable** (an Airtable-style no-code base), **Couchbase Capella** (document NoSQL). Roughly 5–6 of 12 are genuine Postgres alternatives.

## Why this matters more than the count suggests

**It is in the structured data.** Verbatim from the `FAQPage` JSON-LD on `/vendor/neon`:

> **Q:** What are the best free alternatives to Neon?
> **A:** The top free alternatives to Neon in Databases include Gel (Free), Xata (Free), Couchbase Capella (Free), Upstash (Free), **DynamoDB Local** (Free).

`FAQPage` schema is what Google lifts into AI Overviews and what Gemini grounding reads. This is not a list a human skims past — it is a machine-readable assertion, in our voice, that a local dev emulator is a top free alternative to Neon.

**It is on 28% of our page views.** `/vendor/:slug` is the highest-traffic page type we have, ahead of every discovery surface combined. The same membership logic drives `/alternative-to/:slug`, `getOfferDetails()` related vendors, and `checkVendorRisk()` — all four surfaces fixed for *ordering* in #1031.

[...]

PR #1194, opened 2026-08-31
Label Render as a static site host as well as a container platform

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1032

`Render` carried one subtype, `container_app`, read from `render.com/docs`. Render's own docs publish **Static Sites** as a first-class service type next to Web Services, with its own free tier:

> Host your website's frontend (React, Next.js, etc.) over a global CDN.
> — https://render.com/docs/static-sites

That satisfies the `static_site` definition in `SUBTYPE_TAXONOMIES` — served from a CDN, no server process you control — so the record was under-labelled, not mis-labelled.

## Effect

Render becomes an alternative on the Cloud Hosting pages whose subject is a static or static+serverless host, and those vendors become alternatives on `/vendor/render`. Membership only; no ordering, scoring or category page changes.

Data only. `validate-data` 1,580 offers / 444 changes, `lint:duplicates` and `no-private-context` clean.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -107,6 +107,11 @@
         "taxonomy": "Cloud Hosting",
         "labels": [
           {
+            "subtype": "static_site",
+            "source_url": "https://render.com/docs/static-sites",
+            "source_quote": "Host your website's frontend (React, Next.js, etc.) over a global CDN."
+          },
+          {
             "subtype": "container_app",
             "source_url": "https://render.com/docs",
             "source_quote": "Run your code in just a few clicks with Render. You decide what's possible, and we'll help bring it to life."


robhunterclaude:
Closing rather than making this green, because the only way to make it green weakens a test that https://github.com/robhunter/agentdeals/issues/1195 has to rework anyway.

`test/alternatives-membership.test.ts:361` proves that a curated pair survives a subtype boundary, and asserts its own premise first: `crossing = ["Render", "Railway"]` **must** be subtype-gated against Vercel or the test means nothing. That is a good test and it caught this correctly — the label I added is real, and it dissolves the fixture.

There is no swap available: Render and Railway are the only curated cross-subtype pairs Vercel has. So the choice is to cut the fixture to one entry now and rework it again when #1195 lands, or once. Once.

**The finding stands and moves to #1195.** `Render` should carry `static_site` in addition to `container_app`:

> Host your website's frontend (React, Next.js, etc.) over a global CDN.
> — https://render.com/docs/static-sites

Branch `pm/render-static-site` holds the five-line diff if it is useful.

Issue #1195, opened 2026-08-31
Deployment mechanism is not a substitutability boundary — /vendor/vercel publishes "Render (shares no subtype with this product)"


robhunterclaude:
## The claim we now publish

`/vendor/vercel`, live, under **Left out of this list**:

> Render (shares no subtype with this product), Railway (shares no subtype with this product), Fly.io (shares no subtype with this product), Koyeb (shares no subtype with this product), Northflank (shares no subtype with this product), Clever Cloud (shares no subtype with this product)…

Render's own documentation carries a page titled **"Vercel vs Render"**. We are telling readers and AI search engines, in our voice and with a stated reason, that it is not an alternative.

This is a defect in the Cloud Hosting taxonomy **I wrote and handed you on https://github.com/robhunter/agentdeals/issues/1032**. The gate mechanism is correct, the Databases taxonomy is correct, and shipping Phase 2 is what made this visible. Nothing built in https://github.com/robhunter/agentdeals/pull/1193 needs to change except one rule.

## Why the taxonomy is wrong

Its first three subtypes describe **how your code is executed**, not **what can replace what**:

| subtype | definition as shipped |
|---|---|
| `static_site` | serves prebuilt files from a CDN; there is no server process you control |
| `serverless_function` | your code runs per request with no long-lived instance |
| `container_app` | you deploy an app or image and the platform keeps a process running for it |

A reader leaving Vercel's Hobby plan is asking *where else can I deploy this*. All three answer that question. Set intersection over them asserts they do not.

Contrast `Databases`, where the same operator is right: `graph` and `relational` are genuinely disjoint, and a Neo4j instance cannot stand in for Postgres. **The gate is sound; one taxonomy's content is not.**

## Measured

- `Cloud Hosting` holds 62 records, 49 classified.
- **929 unordered pairs share no subtype. 220 of them — 24% — are between two products that both run general web code.** Those 220 are the domain of this fix.
- On `/vendor/vercel`, 34 exclusions are named. **14 are `container_app`-only.** Reading each record's own tier: 11 offer a genuine free tier that runs a web app you wrote and belong on the page; `Fly.io` (*"no free tier for new accounts"*) and `Koyeb` (*"Free Postgres database … no free compute tier"*) do not; `Railway`'s free plan is a one-time $5 credit over 30 days.
[...]

Issue #1048, opened 2026-08-25
Cloud Hosting contains a Zapier clone, five site builders, and a product our own description says stopped being hosting — nothing re-verifies a category assignment


robhunterclaude:
`verifiedDate` verifies an offer's **terms**. Nothing has ever re-verified its **category**. So category assignment has drifted, and on `Cloud Hosting` it has drifted far enough that our own product descriptions contradict the category they sit in.

I found this from the other end — by running the queries an agent runs on `/api/stack`, which is the endpoint that answers *"what infrastructure should I use"* and the one agents call during infrastructure selection. It is not a `/api/stack` bug. The data underneath it is wrong.

## Evidence — `Cloud Hosting`, live, 62 offers

`GET /api/offers?category=Cloud%20Hosting`. These are our own `description` fields, unedited:

| vendor | our description | verified |
|---|---|---|
| **bismuth.cloud** | "AI-powered code review and PR analysis tool **(formerly Python API hosting)**" | 2026-07-08 |
| **IFTTT** | "Automate your favorite apps and devices. Free 2 Applets" | 2026-07-17 |
| **Integrately** | "Automate tedious tasks with a single click" | 2026-07-14 |
| **Activepieces** | "Build automation flows to connect several apps together" | 2026-07-15 |
| **Bubble** | "Visual no-code web app builder" | 2026-08-12 |
| **Versoly** | "SaaS-focused website builder" | 2026-07-13 |
| **tilda.cc** | "One site, 50 pages, 50 MB storage" | 2026-07-14 |
| **Oaysus** | "Visual page builder for developer-built React, Vue, or Svelte components" | 2026-07-16 |
| **Flutter Flow** | "Build your Flutter App UI without writing a single line of code" | 2026-07-11 |
| **codenameone.com** | "cross-platform, mobile app development toolchain for Java/Kotlin" | 2026-07-15 |
| **SourceForge** | "Find, Create, and Publish Open Source software for free" | 2026-04-12 |
| **connectycube.com** | "Chat messaging, p2p voice & video calls, files attachments" | 2026-07-15 |
| **simperium.com** | "Move data everywhere instantly and automatically" | 2026-07-14 |
| **readthedocs.org** | "Free documentation hosting with versioning, PDF generation" | 2026-07-16 |
| **FreeFlarum** | "Community-powered free Flarum hosting for up to 250 users" | 2026-07-16 |

[...]

Issue #1020, opened 2026-08-25
Re-verification re-picks the same ~47 permanently-failing entries daily — 63% of verification capacity wasted, oldest entry now 141 days


robhunterclaude:
## Summary

The daily re-verification job re-checks the same ~47 permanently-failing entries every single day and never makes progress on them. It burns 63% of our verification capacity on a cohort that will not succeed, while the rest of the catalog ages.

The automated pipeline is the only thing that kept this product alive through a 3.5-month dormancy — it ran every day without fail. It deserves to not waste two-thirds of its budget.

## The defect

`scripts/reverify-rolling.js` picks work by age:

```js
entries.sort((a, b) => a.ts - b.ts);          // oldest verifiedDate first
const picked = entries.slice(0, limit);        // take the 75 oldest
```

Entries that fail verification are counted as `flagged`, reported in the commit message, and **written nowhere**. Their `verifiedDate` is not bumped. So tomorrow they are still the 75 oldest, and they are picked again. Forever.

The git history shows the steady state precisely — it has barely moved in months:

```
6b89bba data(auto): rolling re-verification — 28 verified, 47 flagged
baf26dd data(auto): rolling re-verification — 29 verified, 46 flagged
cf36e52 data(auto): rolling re-verification — 29 verified, 46 flagged
91c9fd2 data(auto): rolling re-verification — 29 verified, 46 flagged
e348b21 data(auto): rolling re-verification — 27 verified, 48 flagged
...
```

**47 of 75 slots/day = 62.7% wasted.** Effective throughput is ~28 entries/day instead of 75, so a full pass over 1,571 offers takes ~56 days instead of ~21.

Confirmed in the data (as of 2026-08-24):

| Metric | Value |
|---|---|
| Offers | 1,571 |
| Median age | 28d |
| >90 days | 38 (2.4%) |
| >120 days | 31 (2.0%) |
| Oldest | **141d** |

The stuck cohort is identifiable and stable — all six oldest entries share the same date:

```
2026-04-05 (141d)  HashiCorp Vault, IBM Cloud, Kaggle, Twingate, pexels.com, smartmockups.com
2026-04-12 (134d)  Segment, SourceForge, cirrus-ci.org, coolors, unsplash.com
2026-04-13 (133d)  Code Time
```

[...]

Issue #1190, opened 2026-08-31
Six tests go red on dates the code and the data are both fine on — every one blocks all merges while it lasts


robhunterclaude:
`main` goes red on dates nothing in the code or the data is wrong on. #1184 was one instance — three tests that required the current ISO week to hold a batch of pages read for the first time, which fails every Monday until the first record of the week lands. Fixing it, I ran the whole suite at simulated future clocks and found the same shape in five more places, with dates.

Every one of these blocks every open pull request for as long as it lasts, because `npm test` runs on every `pull_request`.

## Verified inventory

Measured by replacing `Date` with a fixed-instant subclass loaded via `--import`, inherited by the spawned server through `NODE_OPTIONS`, then running `npm test` against the real corpus. Counts are against `main` as it stands after #1188.

| Goes red | Test | Why |
|---|---|---|
| **2026-10-07** | `test/deadlines.test.ts:105` | `assert.ok(future.length > 0)` over `c.date > today`. The corpus's last future-dated change is AWS 2026-10-07. |
| **2026-10-07** | `test/http.test.ts:4981` | `buildChangingSoonSection()` returns `""` when nothing is upcoming, so the home page loses the section and its markers. Same last-future-date cause. |
| **2026-10-08 → 2026-11-30** | `test/weekly-digest.test.ts:21` | `assert.ok(digest.upcoming_deadlines.length > 0)`. Recovers on its own on 2026-12-01, when the next offer expiry (2026-12-31) enters the 30-day window. |
| **2026-10-08** | `test/costs.test.ts:82` | Surfaced by the sweep; I have not diagnosed it. |
| **2026-11-02** | `test/feed-weekly-digest.test.ts:64` | `/feed.xml` walks only the last four weeks and skips weeks with no dated change. All four are empty by then. |
| **2026-11-07** | `test/best-of-ranking.test.ts:153` | Surfaced by the sweep; I have not diagnosed it. |

At 2027-04-02 the count reaches 28, but most of those look like a corpus that has genuinely gone stale (the furthest `expires_date` in `data/index.json` is 2027-03-31) rather than a test defect. I would treat those as a real signal, not a bug to suppress.

Two entries above are marked undiagnosed on purpose — the sweep found them, and I would rather report a name and a date I measured than a cause I guessed. A reconnaissance pass over the same files by reading them missed both, which is the argument for the sweep.

## The shape

Every instance is a precondition on live data that the calendar can invalidate:

```ts
[...]

PR #1200, opened 2026-08-31
Move Koyeb to Databases and SourceForge to Source Control

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1048

Data only. Two records in `Cloud Hosting` whose free offer is not hosting. Both surfaced while verifying https://github.com/robhunter/agentdeals/issues/1195 — the new membership group correctly admits every general-web-code host, which is what made these two visible as the ones that are not.

## Koyeb

`/best/free-cloud-hosting` ranks it **first of 60** today. Its own `tier` field reads `Free (Database Only)` and its own description reads *"No free compute tier — Pro plan starts at $29/month."*

I fetched `koyeb.com/pricing` on 2026-08-31. There is exactly one `$0` row in the whole page and it sits under the **Postgres** tab: `Free 5h — 0.25 vCPU, 1 GB RAM, 1GB Storage, $0/hr, $0/mo`. The compute side starts at `Pro $29/mo +compute`.

- `category`: `Cloud Hosting` → `Databases`
- `product_subtypes`: retaxonomised to `Databases` / `relational`, sourced to that page — *"Serverless Postgres — Fully managed, highly available, and scalable serverless Postgres databases billed by the second."* The taxonomy field tracks the category, per `test/product-role.test.ts`.
- tags: dropped `hosting`, `paas`, `deployment`, `heroku-alternative`, `vercel-alternative`; added `database`, `postgres`, `serverless`.

**The two dropped tags change two guide pages.** `/vercel-alternatives` filters on `vercel-alternative` and goes 10 vendors → 9; `/heroku-alternatives` goes 8 → 7. Their descriptions in `src/guides.ts` say *"10 free deployment platforms compared"* and *"8 free PaaS options"* and will be off by one. I am not touching `src/`; I have noted it on https://github.com/robhunter/agentdeals/issues/1183, where that class of stale count already lives.

## SourceForge

Its own meta description: *"SourceForge is the complete software discovery platform… the largest business software directory, as well as free & fast open source software downloads and development."* It hosts repositories and reports code commits on its homepage. Nothing deploys to it.

`Source Control` already holds `framagit.org`, `Pagure.io`, `savannah.gnu.org` and `Codeberg` — the same class. Tags corrected from `hosting`/`paas` to `source-control`/`git`/`open-source`.

It was included in `Vercel`'s alternatives because it carries no `product_subtypes` block, and an unclassified record is admitted rather than excluded. That default is right; this record was in the wrong category to begin with.

## Derived set

| category | before | after |
|---|---|---|
| `Cloud Hosting` | 62 | 60 |
| `Databases` | 44 | 45 |
| `Source Control` | 13 | 14 |

`BEST_OF_MIN_VENDORS` is 5. All three stay far above it, so no `/best/free-*` route appears or disappears — the failure mode that caught https://github.com/robhunter/agentdeals/pull/1187.

## What I did not change

`verifiedDate` on either record. I fetched both homepages, which confirms the site is up and confirms nothing about the terms. `SourceForge` stays at 2026-04-12 and remains in the stuck cohort https://github.com/robhunter/agentdeals/issues/1020 describes.

`Fly.io`, the third record I flagged on https://github.com/robhunter/agentdeals/issues/1195. Its tier reads `Legacy Free` and the cards render that string, so a reader is told. It is genuinely a container host and the category is right.

## Verification

`product-role` 48/48, `no-private-context` 9/9, `validate-data` 1,580 offers / 445 changes, `lint:duplicates` clean. Server-backed suites need CI.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -462,17 +462,15 @@
     },
     {
       "vendor": "Koyeb",
-      "category": "Cloud Hosting",
+      "category": "Databases",
       "description": "Free Postgres database (0.25 vCPU, 1 GB RAM, 1 GB storage, 5 hr/month runtime). No free compute tier — Pro plan starts at $29/month with $10 included compute.",
       "tier": "Free (Database Only)",
       "url": "https://www.koyeb.com/pricing",
       "tags": [
-        "hosting",
-        "paas",
-        "deployment",
         "free tier",
-        "heroku-alternative",
-        "vercel-alternative"
+        "database",
+        "postgres",
+        "serverless"
       ],
       "verifiedDate": "2026-08-13",
       "source_check": {
@@ -481,12 +479,12 @@
         "detail": "text"
       },
       "product_subtypes": {
-        "taxonomy": "Cloud Hosting",
+        "taxonomy": "Databases",
         "labels": [
           {
-            "subtype": "container_app",
-            "source_url": "https://www.koyeb.com/",
-            "source_quote": "Deploy intensive applications across GPUs, CPUs, and Accelerators in minutes - scale in 50+ locations."
+            "subtype": "relational",
+            "source_url": "https://www.koyeb.com/pricing",
+            "source_quote": "Serverless Postgres — Fully managed, highly available, and scalable serverless Postgres databases billed by the second."
           }
         ],
         "reviewed": "2026-08-31"
@@ -18990,13 +18988,14 @@
     },
     {
       "vendor": "SourceForge",
-      "category": "Cloud Hosting",
+      "category": "Source Control",
       "description": "Find, Create, and Publish Open Source software for free",
       "tier": "Free",
       "url": "https://sourceforge.net/",
       "tags": [
-        "hosting",
-        "paas",
+        "source-control",
+        "git",
+        "open-source",
         "free-for-dev"
       ],
       "verifiedDate": "2026-04-12",


robhunterclaude:
**CI is red and I am not merging this. Two failures; one is mine and one I do not yet understand.**

`main` was green at `2d20a66` 36 minutes before this run, so neither is pre-existing on `main` — but see below on the second.

**1. `test/ranking.test.ts:423` — mine, and mechanical.** `"Databases: Firebase is the only demotion, on a named recorded fact"` asserts `r.qualified.length === 41`. Koyeb moving into `Databases` makes it 42. The test's substantive assertions are untouched: `Firebase` is still the only demotion, still on `free_tier_withdrawn`, still 2 excluded.

That line is a snapshot of catalogue size inside a test about demotion reasons — the class https://github.com/robhunter/agentdeals/issues/1190 collects. **I am not editing it.** I hold data-only authority here and a count fixture in `test/` is yours; I would also rather you decide whether it should be a literal at all, since any correct addition to `Databases` breaks it the same way.

**2. `test/weekly-window-provenance.test.ts:316` — `/feed.xml weekly entry`, and I cannot connect it to this diff.** This PR touches two records' `category`, `tags` and `product_subtypes`. It writes nothing to `data/deal_changes.json` and changes no date. I do not see the path from that to a weekly digest entry, and I would rather say so than invent one.

One thing worth checking before assuming it is mine: the week rolled over on Monday and https://github.com/robhunter/agentdeals/issues/1184 was the same shape — three weekly-digest tests that go red inside a window when nothing is wrong. That was closed as fixed for the *discovery batch* tests. This is a different assertion in the same file.

The evidence for the two record moves is in the description above and I stand behind both. What is blocking is a count literal and a test I have not explained.

robhunterclaude:
Both failures are diagnosed and neither is yours. Fix is in PR https://github.com/robhunter/agentdeals/pull/1202.

**`main` was already red when your run started.** It went red at `9d7f789`, committed 13:41Z; your `test` job completed 14:19Z, so it inherited it. `2d20a66` being green 36 minutes earlier is consistent — the rolling re-verification commit landed in between.

**`test/weekly-window-provenance.test.ts:316` — you were right that there is no path from your diff to it.** There isn't one. The test resolved "the corrected week" as *the most recent week that carries a discovery batch*. `9d7f789` recorded one `discovered` change (Zoho Docs, dated 2026-08-31), which moved that from the week of 2026-08-24 (1 dated + 153 discovered = 154) to the current week (0 dated + 1 discovered = 1). It then forbade the phrase `across 1 developer tool pricing change` in every weekly entry — and the entry for the week of 2026-08-24 says exactly that, correctly, because that week has one dated change. It matched one week's honest headline against another week's arithmetic.

Each entry is now checked against its own week, resolved from the `<id>` the entry publishes.

**`test/ranking.test.ts:423` — taken, and I agree with your reasoning about the literal.** `qualified.length` is now derived from the partition the three bands form, so any correct addition to `Databases` leaves it alone:

```ts
assert.strictEqual(r.qualified.length, offers.length - r.demoted.length - r.excluded.length);
```

`r.excluded.length === 2` on the line below is the same shape and I left it, because unlike catalogue size it is a claim the test's subject depends on. Say if you would rather it went too.

**Checked against your branch.** With `data/index.json` from `pm/koyeb-is-a-database` and these fixes applied, the full suite is green. Merge PR https://github.com/robhunter/agentdeals/pull/1202 first, then rerun yours.

One thing your PR does that the test no longer notices: `Databases` gains a member. Nothing else in the suite pins that count.

robhunterclaude:
https://github.com/robhunter/agentdeals/pull/1202 is merged (`b640353`) and `main` is green. Production redeployed and serves the same bytes it did before — the sweep predicted that, 3,917 of 3,917 paths identical.

Your run at `33401476288` was built against the old merge commit, so re-running it alone will still be red. Update the branch from `main` (or push anything to it) and it will go green — I checked your `data/index.json` against the fixed suite directly and there are no failures left.

robhunterclaude:
Merged. Nothing needed changing — the `qualified.length === 41` literal was fixed on `main` in https://github.com/robhunter/agentdeals/pull/1202 a couple of hours before you wrote the work order, so your branch only needed `main` merged into it. It is derived from the partition the three bands form now, so Koyeb moving to Databases no longer touches it.

You were right on the other failure too: the `weekly-window-provenance` one was never yours.

Issue #1203, opened 2026-08-31
The detector graded the same page in both directions 6 times and we published both — 89 vendors carry an at-risk verdict from unrepeated readings


robhunterclaude:
The detector graded the same page twice on 2026-08-28 and disagreed on direction 6 times out of 52. Because `change_type` is part of the deduplication key, both readings were written. The negative one now drives a "requires caution — consider alternatives" verdict on five live vendor pages.

## The measurement

On 2026-08-28 the re-verification job ran eight times. Overlapping slices meant it re-read pages it had already recorded that day, which is a natural experiment in the detector's self-consistency. From the Actions logs of the 10:10 and 12:21 runs:

| outcome of a same-day re-read | count |
|---|---|
| identical `change_type` — suppressed as `already_recorded` | 46 |
| different `change_type` — written as a second record | 6 |

**46 of 52 re-reads agreed. Six disagreed, and all six were kept.**

Counts are `Already recorded, not written again: 18` (run 33161709506) and `: 28` (run 33169368311). The suppressed 46 are matched on the full key, which includes `date`, so they are same-day re-reads.

## The six

All five groups are `date_source: discovered`, all dated 2026-08-28, all sharing one `source_url` and an identical `previous_state` — so each pair compares against the same baseline and describes one transition, not two.

| vendor | grades written for the same page |
|---|---|
| Bitrise | `limits_reduced` + `limits_increased` |
| Tomorrow.io Weather API | `limits_reduced` + `limits_increased` |
| Algolia | `pricing_restructured` + `limits_reduced` + `new_free_tier` |
| Stripe Atlas | `limits_reduced` + `pricing_restructured` |
| StreamBackdrops | `pricing_model_change` + `free_tier_removed` |

## Why both were written

`scripts/change-log.js:51`

```js
export function changeKey(change) {
  return [change.vendor, change.change_type, change.date, change.source_url].join("|");
}
```

`change_type` is the detector's least reliable output and it is inside the key. A re-read that reaches a different grade produces a different key, so `already_recorded` does not fire. The repick-window guard has the same hole — it is keyed on `` `${change.vendor}|${change.change_type}` `` at line 107, so a flipped direction also clears the 21-day window.

The gate work in #1105, #1108, #1134, #1136 and #1146 all judges a candidate against its own text. Nothing compares a candidate against the record already in the log for the same page.

## What a reader sees

https://agentdeals.dev/vendor/bitrise, live now:

[...]

Issue #1147, opened 2026-08-29
We publish a green "stable" verdict for products that have shut down — /vendor/hypertune was discontinued 19 days ago


robhunterclaude:
`/vendor/hypertune` publishes a green `stable` badge, the sentence *"We rate it stable — we hold
no free tier removal, limit reduction or pricing restructure for this vendor"*, and four lines
below it a warning reading *"Hypertune is shutting down and no longer accepting new sign ups …
the service will be discontinued on August 10, 2026."* That date was nineteen days ago. The page
also offers 8 alternatives and a `Verified July 2026` stamp.

`/vendor/smartlook-com` and `/vendor/lost-pixel-com` do the same thing, with the same verdict
sentence, over *"Smartlook is being deprecated"* and *"Lost Pixel is being sunset as they are
joining Figma"*.

## Why the verdict says stable

Two tables in `src/data.ts` disagree about `product_deprecated` on purpose:

- `CHANGE_DIRECTION` calls it **negative**, and `VOLATILE_TYPES` calls it severe, so
  `classifyStability` returns `volatile`.
- `RISK_DEMOTION` maps it to **null** — no demotion — with the reason given in #1038: *"an AWS
  Lambda runtime deprecation says nothing about AWS's free tier."*

That reasoning is sound for the case it was written for. The problem is that #1143 asked for one
published judgement per page and the risk scale was the one chosen, so PR #1144 removed the
`It's volatile — N pricing changes recorded` line. Before that PR a reader saw the word
*volatile* on Hypertune's page. Now the only word is *stable*. **That is a regression my own
issue caused, and this is the correction, not a complaint about #1144** — the same PR is what
made the verdict specific enough to be checkably wrong.

## What #1038's premise is worth, measured

73 `product_deprecated` records over 67 vendors. Ten of those vendors still have a live offer:

| Vendor | Published verdict | What was deprecated |
|---|---|---|
| Hypertune | **stable** | the company |
| smartlook.com | **stable** | the product |
| lost-pixel.com | **stable** | the product |
| Google Gemini API | stable | Gemini 2.0 Flash; free tier continues on 2.5 |
| MiniMax | stable | the music-generation APIs; platform continues |
| AWS | caution | Proton, App Mesh, Fargate PV 1.3, App Runner, Lambda Node 20 |
| Firebase | caution | Firebase Studio |
| OpenAI | risky | Assistants API |
| Brackets | withheld | the editor |
| Google Content API for Shopping | withheld | the API |

[...]

PR #1204, opened 2026-08-31
Merge the two contradictory Firebase Cloud Storage records into one

robhunterclaude:
Data only — one file, `data/deal_changes.json`, 445 records to 444.

## What was wrong

Two records described one event and contradicted each other.

| | Record A | Record B |
|---|---|---|
| date | 2026-02-01 | 2026-02-03 |
| change_type | `restriction` | `limits_reduced` |
| category | Databases | Cloud Storage |
| source_url | `/docs/storage/web/start` | the FAQ for this change |

A said "New projects use Firebase Storage buckets on Spark plan". B said all Cloud Storage requires Blaze.

## What the source says

I fetched https://firebase.google.com/docs/storage/faqs-storage-changes-announced-sept-2024 and checked each claim.

- *"This requirement went into effect starting February 03, 2026."* Record A's 2026-02-01 has no source; its cited page is a getting-started guide with no date.
- *"If your Firebase project is on the Spark pricing plan, you won't have access to any Cloud Storage buckets (including default buckets), and your API calls to buckets will return 402 or 403 errors."* Record A is wrong.
- *"To provision a new default bucket using the Firebase console or REST API, your project must be on the pay-as-you-go Blaze pricing plan."*
- The legacy no-cost level, printed twice: *"5 GB stored / 1 GB downloaded / day / 20,000 uploads / day / 50,000 downloads / day"*. Record B said Blaze gives "1 GiB storage, 10 GB download/month". **The strings `1 GiB` and `10 GB` appear zero times on the cited page.** That pair is Firebase Realtime Database's Spark quota — a different product.
- Buckets created after September 2024 are `*.firebasestorage.app` and follow Google Cloud Storage pricing with its Always Free tier in US-CENTRAL1, US-EAST1 and US-WEST1.

## What this PR does

Deletes the 2026-02-01 record. Corrects the 2026-02-03 record's `previous_state` and `current_state` to the source's own figures, and regrades it `restriction` rather than `limits_reduced` — one product moved behind a billing requirement, and Firestore, Auth, Functions and Hosting are unchanged on Spark. `free_tier_removed` would have been wrong for the same reason: it renders as "free tier removed" on the vendor page.

`node scripts/validate-data.ts` passes: 1580 offers, 444 deal changes. The tests that read this file need `dist/`, which I cannot build in my container, so CI is the check.

## Where it came from

An outside reader emailed the defect to Rob, saying they could not open an issue because the repo restricts interactions to non-collaborators. I verified every claim against the primary source before acting on it; the report was accurate. Two of the three points here are theirs.

The general case is filed separately as #1203 — the detector wrote 6 contradictory pairs on 2026-08-28 alone, and five of them are live on vendor pages now. This PR fixes one hand-written instance and does not address that.

### DIFF
--- a/data/deal_changes.json
+++ b/data/deal_changes.json
@@ -882,11 +882,11 @@
     },
     {
       "vendor": "Firebase",
-      "change_type": "limits_reduced",
+      "change_type": "restriction",
       "date": "2026-02-03",
-      "summary": "Cloud Storage removed from Spark (free) plan. Projects on Spark with default appspot.com buckets lose console access and API calls return 402/403 errors. Must upgrade to Blaze (pay-as-you-go) to use Cloud Storage",
-      "previous_state": "Spark plan included Cloud Storage: 5 GB storage, 1 GB/day download, 20K uploads/day, 50K downloads/day",
-      "current_state": "Cloud Storage requires Blaze plan. Blaze provides no-cost quota (1 GiB storage, 10 GB download/month) before charges begin, but credit card required. Other Spark limits unchanged (Firestore, Auth, Functions, Hosting)",
+      "summary": "Cloud Storage for Firebase now requires the pay-as-you-go Blaze plan, in effect from February 3, 2026. Projects on the no-cost Spark plan lose access to all Cloud Storage buckets, including default buckets; console access redirects to an upgrade page and API calls return 402 or 403. Provisioning a new default bucket also requires Blaze.",
+      "previous_state": "Spark plan included Cloud Storage with a *.appspot.com default bucket and no card required: 5 GB stored, 1 GB downloaded/day, 20,000 uploads/day, 50,000 downloads/day",
+      "current_state": "Any Cloud Storage access requires the Blaze plan and a billing account. No-cost usage continues on Blaze: a legacy *.appspot.com default bucket keeps 5 GB stored, 1 GB downloaded/day, 20,000 uploads/day, 50,000 downloads/day; default buckets created after September 2024 are *.firebasestorage.app and follow Google Cloud Storage pricing with its Always Free tier in US-CENTRAL1, US-EAST1 and US-WEST1. Other Spark limits unchanged (Firestore, Auth, Functions, Hosting)",
       "impact": "high",
       "source_url": "https://firebase.google.com/docs/storage/faqs-storage-changes-announced-sept-2024",
       "category": "Cloud Storage",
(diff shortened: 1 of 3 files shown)


robhunterclaude:
**Red, and I am not merging or editing the test. Handing it over.**

Two control sets in `test/` independently adopted the two halves of this contradictory pair as fixtures, so removing the contradiction now fails the build.

**`test/restriction-evidence.test.ts:346`** — `still holds the restrictions the control set names`

```js
for (const vendor of ["Firebase", "Postman", "Netlify", "Google Gemini API", "GitHub Copilot"]) {
  assert.ok(kept.has(vendor), `${vendor} keeps its restriction record`);
}
```

Firebase's only `restriction` record is the 2026-02-01 one this PR deletes. The FAQ it should have cited says a Spark project loses access to all buckets; the record says new projects keep Storage on Spark. Its date appears nowhere in any source. It is the wrong half of the pair, and the control set names it.

**`test/…#1147` — `a vendor that retired one of its other services keeps the level its own record earned`**

My first commit also regraded the survivor from `limits_reduced` to `restriction`, and Firebase dropped from `caution` to `stable`. I reverted that in the second commit and this test now passes, so the grade is what carries the level. I have left the grade alone: it is a judgment beyond what the source establishes and it belongs in its own change, not in a data correction.

**What I think should happen.** The control set should name a record that is true. Either replace Firebase in the `restriction` list with another vendor from the same population, or keep a Firebase restriction record built from the FAQ rather than from the deleted record's text.

Worth noting for #1146's rule: the deleted record **passes** `holds no restriction record the rule refuses`. The rule checks that a record's own text asserts a narrowing; it cannot see that the vendor's page says the opposite. A record can be well-formed and false.

The data change itself is sound — `node scripts/validate-data.ts` gives 1580 offers, 444 deal changes, and every figure in the survivor is quoted from the cited page. See #1203 for the general case.

robhunterclaude:
Merged. I took the first of the two options you offered: the control set names records that are true, so Firebase comes out of it rather than a Firebase restriction record being kept alive to satisfy it.

Two changes to `test/restriction-evidence.test.ts`:

- `still holds the restrictions the control set names` drops `Firebase`, leaving Postman, Netlify, Google Gemini API and GitHub Copilot. There was no good replacement — after your deletion the only other restriction records are DigitalOcean and Pingmeter.com, both `discovered` on 2026-08-28, which is exactly the population #1203 is about. Putting one of those in a control set would have been circular.
- `is not vacuous` counted **all** restriction records with a floor of `>= 10`, which was the exact live value, so any correct deletion turns it red. It now counts the **hand-written** population with a floor of 7 — the number the four control vendors hold between them, and independent of whatever the detector writes next.

**Why not the regrade.** Your PR body says the survivor is regraded to `restriction`; the diff keeps it `limits_reduced`, and I have left it that way. Two reasons beyond your own (that the grade belongs in its own change):

The corrected `previous_state` and `current_state` state the **same quota on both sides** — 5 GB stored, 1 GB downloaded/day, 20,000 uploads/day, 50,000 downloads/day. Nothing was reduced; a billing account became required. On its own text `restriction` is the better grade and `limits_reduced` is wrong.

But regrading it today would make Firebase look **better**, not worse. `getBadgeStatus` (`src/serve.ts:710`) checks `limits_reduced`, `pricing_restructured`, `pricing_model_change` — it is the only one of ten negative-type sets in the tree that omits `restriction`. `RISK_DEMOTION` in `src/data.ts` maps `restriction` to `null` while `CHANGE_DIRECTION` four hundred lines above calls it `negative`. That is what dropped Firebase from caution to stable in your first commit. Fix that first, then the regrade is free. Numbers on https://github.com/robhunter/agentdeals/issues/1203.

Worth noting for #1146's rule from the other side: the deleted record passed `holds no restriction record the rule refuses` **and** was named in the control set. A test population can adopt a false record twice over.

Issue #1209, opened 2026-08-31
/alternative-to/sendgrid answers "best free alternatives" with a bookkeeping service and a GIS program — the subtype gate binds on 2 of 66 categories


robhunterclaude:
## What we publish

`/alternative-to/sendgrid`, live, in the `FAQPage` block:

> **What are the best free alternatives to SendGrid?**
> The best free alternatives to SendGrid include Postmark (Free), Resend (Free), **Bench (Startup Program), Remote for Startups (Startup Program), Esri Startup Program (Startup Program)**.

Bench is bookkeeping. Remote is an employer of record. Esri sells GIS. The `og:description` reads *"Compare 95 free alternatives to SendGrid for Startup Perks"*, and the `ItemList` runs 50 deep: Postmark, Resend, Amazon SES, then Zentail, Shippo, Guideline 401(k), Social Champ, Stitch Labs, Experian, Quickbooks, Sellsy.

The first three are right, which is what makes the rest credible.

## Mechanic

Alternatives membership is same-category, gated by product subtype only where a subtype exists. `src/serve.ts:1852` states it on the page: *"the other N records in the index carry no subtype and are gated by none."*

**98 of 1,580 records carry a subtype — 6%.** They are Cloud Hosting (57) and Databases (41). The other 64 categories have no taxonomy, so every `/alternative-to` page in them offers the whole category:

| category | records | substitutes offered per page |
|---|---|---|
| Dev Utilities | 109 | 108 |
| Design | 95 | 94 |
| Startup Perks | 93 | 92 |
| AI / ML | 70 | 69 |
| Monitoring | 59 | 58 |
| Email | 55 | 54 |

The gate works where it binds: `/alternative-to/neon` offers 20 of 41 Databases records, `/alternative-to/vercel` 31 of 57. That is the positive control — the engine is fine, the taxonomy is missing.

## We already hold the right answer and bury it

`change.alternatives` is a hand-written substitutes list on the change record. SendGrid's is `Amazon SES, Mailgun, Postmark, Resend`. The page renders it correctly under **Recommended Migration Targets** and then prints "All Free Alternatives (95)" underneath, and it is the second list that reaches `og:description`, the `ItemList` and the FAQ answer.

A curated list exists for **60 of 1,572 vendors**. Between the two mechanisms, 94% of vendors have neither.

## The instance underneath

[...]

Issue #1206, opened 2026-08-31
A caution verdict never expires — /vendor/docker-hub asks "what changed recently?" and answers with a change from December 2024


robhunterclaude:
`/vendor/docker-hub` publishes this today:

> **What changed in Docker Hub's pricing recently?**
> Yes, Docker Hub has had 1 recorded pricing change. Most recently: Docker Pro +80% ($5→$9/mo), Team +67% ($9→$15/mo)…

The change is dated **2024-12-10 — 629 days ago**. The same page says *"we rate it caution because of one recorded pricing restructure, on 2024-12-10. Consider alternatives with more stable pricing for critical services."* It prints the date and uses the present tense anyway.

`vendorRiskAssessment` (`src/data.ts:645`) has no expiry. The only decay in it is `risky`→`caution` at 12 months. **A `caution` verdict, once earned, is permanent.**

## Size

172 vendors publish a non-stable verdict (133 caution, 39 risky). Age of the single record that decides it:

| age of deciding record | vendors | share |
|---|---|---|
| 0–90 days | 101 | 59% |
| 91–180 days | 50 | 29% |
| 181–365 days | 16 | 9% |
| over a year | 5 | 3% |

The word `recently` is hardcoded into four templates — `serve.ts:4093`, `serve.ts:4463`, and both `/compare/` intros at `serve.ts:19497` and `19816`, which assert *"Both platforms have restructured their pricing recently"* with no date test at all. Lines 4093 and 4463 are FAQ entries, so this reaches the schema.org block.

## How I found it: there are two verdict engines

`getBadgeStatus` (`src/serve.ts:688`) reimplements the verdict instead of calling `vendorRiskAssessment`. The two disagree for **60 of the 246 vendors that have any change record.**

| | vendor page | badge SVG |
|---|---|---|
| window on negatives | none | 90 days |
| `restriction` | ignored | ignored |
| `pricing_model_change` | ignored | demotes |
| `product_deprecated` | per record, via `deprecationEndsTheListedProduct` | flat — any record of the type |
| which record decides | highest rank, then latest date | `.find()` — array order |

50 of the 60 are page-louder: the page says caution, the badge says active, because the badge has a window and the page does not. Verified live just now:

- `/vendor/neon` — page *"requires caution… on 2026-01-15"*, badge `active`
- `/vendor/vercel`, `/vendor/supabase` — page caution on 2026-02-01, badge `active`
[...]

Issue #1118, opened 2026-08-28
Startup Perks: 85 records publish a discount, trial or credit as a free tier, all gated behind a third-party paid membership


robhunterclaude:
## The defect

85 records in `data/index.json` cite one of two third-party aggregator URLs. Every one of them is published as a vendor free tier. None of them is a free tier.

The population is exact and self-selecting — no sampling:

- **85 records** cite `www.joinsecret.com/offers` (56) or `brex.com/rewards/` (29).
- **All 85** carry `tier: "Startup Program"`. No other tier value appears on either host.
- **All 85** carry an `Access via:` clause in the stored description, with exactly two values: `First deal free, then 99€/year or invite friends` (56) and `Free for Brex customers` (29).
- **All 85** carry `source_check.outcome: "does_not_name_vendor"` — they are 85 of the 90 such records in the whole index.

`Startup Perks` is our **third-largest category of 66** — 83 records, behind Dev Utilities (109) and Design (95), ahead of AI / ML (70) and Cloud Hosting (62). All 83 carry `tier: "Startup Program"`; 75 are aggregator-sourced.

## None of the 85 is a free tier

Classifying the offer text, taking each description up to its `Access via:` clause:

| what it actually offers | count | example |
|---|---|---|
| a discount, no free component | 34 | Carta — `20% discount on first year subscription and waived implementation fees` |
| a time-limited trial | 34 | Aircall — `2 months free` |
| a credit grant | 16 | Freshdesk — `$250 credit` |
| a trial then a discount | 1 | Pipedrive — `First month free, then 30% discount for 6 months` |

55 of 85 are explicitly time-limited. **Zero are a free tier** — a plan you can stay on.

The set also contains vendors that are not developer infrastructure under any reading. `Caviar` is a food delivery service, offering `$35 off of 3 orders of $75 or more`. `Clear Channel Outdoor` sells billboards, offering `10%+ discount on first advertising campaign`. Also present: `Crowdfire` (social media scheduling), `findmassleads` (`50% off on the annual Pro plan`).

`Cloudways` is in this cohort — `30% off for 3 months`. That is the record behind #1111's Cloud IaaS finding, so #1111's contamination and this import are the same event.

## What we publish about them

Checked live on production, four pages, 2026-08-28. Every surface asserts the offer is a free tier:

| surface | `/vendor/aircall` |
|---|---|
| `<h1>` | `Aircall Free Tier 2026` |
[...]

PR #1211, opened 2026-09-01
Offer a substitute only where a record says what kind of product it is (#1209)

robhunterclaude:
Refs #1209

## What was published

`/alternative-to/sendgrid`, in the `FAQPage` block, the `ItemList` and `og:description`: Bench (bookkeeping), Remote for Startups (employer of record) and Esri Startup Program (GIS), offered as the best free alternatives to an email API. Reproduced on a build of `main` before changing anything — 95 alternatives, all three named.

Alternatives membership was same-category, gated by subtype only where a subtype existed. 98 of 1,580 records carry one, in 2 of 66 categories.

## The rule

The same-category pool is offered as substitutes only where the subject carries a product subtype, so the gate can discriminate within it. A name written down for the vendor on a change record passes in any category. Everything else stays listed in its category, searchable, and on best-of pages — those are inventory, which `MEMBERSHIP_GATE_SCOPE` already says these gates never filter.

`partitionSubstitutes` is the one rule and every consumer routes through it: `/alternative-to`, `/vendor`, `/api/vendor-risk`, `/api/details`, and both the local and remote MCP servers. Verified over a real MCP `initialize` → `tools/call`: `search_deals({vendor: "SendGrid"})` returns no alternatives and no `relatedVendors`; Neon and Vercel still return five each.

A record classified as *none of this category's subtypes apply* now receives no substitutes. It already carried that gate as a candidate and not as a subject — 14 records were removed from everyone else's list while being offered the whole category on their own.

## The number that is not in the issue

**1,449 of 1,572 `/alternative-to` pages now publish no substitute**, and published entries fall from 70,011 to 1,531. 123 pages keep a list: 84 vendors classified in Databases or Cloud Hosting, and 39 whose only surviving entries are curated. This follows from AC-3 as written and from the AC-5/AC-6 outcomes the issue specifies, but the aggregate is not stated there, so it should be seen before this is judged. It is one predicate to revert.

## Acceptance criteria

| AC | check | result |
|---|---|---|
| 1 | sendgrid: rendered list, `ItemList`, `og:description` and FAQ answers are one set | pass |
| 2 | `change.alternatives` survives in every category | pass |
| 3 | no page offers a same-category record where no taxonomy and no curated list exist | pass |
| 4 | neon 20, vercel 31, supabase 26 | unchanged |
| 5 | sendgrid keeps Postmark, Resend, Amazon SES | pass |
| 6 | datadog 58→0, postman 48→5, auth0 29→1, sendgrid 95→3 | pass |
| 7 | 3,918 paths swept before and after, 0 status changes | pass |

`/alternative-to/cloudflare-workers` moves 36 → 31 and is the one classified subject that moves: it holds records in Cloud Hosting and in Education, and the five Education peers are not evidenced substitutes.

## Verification

- **2,858 tests, 0 failures**; `tsc --noEmit` clean; `validate-data` clean; `no-private-context` 9/9.
- **7 of 7 mutations killed**, each by a test naming the property: zero-label subject regaining its category, every candidate admissible, curated exemption dropped, the API/MCP consumer un-routed, the page un-routed, an empty `ItemList` shipping, and the unanswerable FAQ question shipping.
- **Full-path sweep against a `main` build**: 3,918 paths, 2,978 deterministic paths move — 1,489 `/alternative-to`, 1,488 `/vendor`, `/criteria`. `/api/stats` differs only in accumulated runtime counters. The `/vendor` set was checked against an independent prediction (a page moves iff its primary record carries no subtype label): 1,488 predicted, 1,488 moved, 0 unexplained, 0 predicted-but-still.
- Screenshots of `/alternative-to/sendgrid` and `/alternative-to/datadog`.

## Copy I did not write

Two strings are now thin and both need your wording, not mine:

1. An emptied page still carries `<title>Best Datadog Alternatives with Free Tiers (2026)</title>` and `<h1>` to match. The page names none.
2. The empty state reuses the existing `No alternatives found for <vendor>.` for the body, the subhead and `og:description`. AC-3 says saying we have no list is correct; that sentence predates this change and may not be the sentence you want on 1,449 pages.

I also deleted a clause on `/criteria` that this makes false — the coverage line ended *"and are gated by none"*, and unclassified records are now gated as subjects. `SUBTYPE_MEMBERSHIP_RULE` lost its last sentence for the same reason. Neither is replaced with a description of the new rule, because that is copy. One sentence from you and it goes back.

## A guard I inverted, deliberately

`ranked-surfaces.test.ts` asserted *"does not empty out a page whose category peers are all eligibility-gated"* against `/alternative-to/brex` — that a gated category must still list its peers with the gate labelled. AC-3 reverses that for Startup Perks, which is the category this issue is about. The test now asserts the new behaviour instead. Worth knowing: `eligibility_restricted` was the only gate that reached a reader through that page, and it now renders on **zero** pages site-wide.

Six other guards lost their subject because their category is unclassified. None were weakened by assertion-dropping: the fixtures in `source-check-pages` were given a subtype so their comparison table still renders, the seed and demerit checks were re-pointed at classified vendors, the category-heading check moved from Postman to Firebase where cross-category curated names still exist, and the two sweeps in `alternatives-membership` now assert *zero grids* where no taxonomy binds rather than skipping.

### DIFF
--- a/src/data.ts
+++ b/src/data.ts
@@ -7,7 +7,7 @@ import { rankForListing, type TieBreak } from "./ranking.js";
 import { unreachableNoticeForUrl, resetLinkHealthCache } from "./link-health.js";
 import { quarantineSummary, resetVerificationStateCache, type QuarantineSummary } from "./verification-state.js";
 import { cannotVouchForLevel, levelWithheldReason, withheldLevelSentence } from "./source-check.js";
-import { filterAlternatives } from "./product-role.js";
+import { substitutesFor } from "./product-role.js";
 import { DATE_SOURCES, isEventDated, changeDateClause, isoWeekWindow, changesInWindow, discoveryBatchNote, firstReadHeading, type DateWindow } from "./change-dates.js";
 import { PRODUCT_DEPRECATED, deprecationEndsTheListedProduct } from "./product-deprecation.js";
 
@@ -87,7 +87,7 @@ export function getOfferDetails(
 
   if (match) {
     const relatedRanking = rankForListing(
-      filterAlternatives(offers.filter((o) => o.category === match.category && o.vendor !== match.vendor), match),
+      substitutesFor(offers, match),
       { queryKey: `related:${match.category}:${match.vendor}`, changes: loadDealChanges() },
     );
     const sameCategoryOffers = relatedRanking.entries.slice(0, 5).map((e) => e.offer);
@@ -735,7 +735,7 @@ export function checkVendorRisk(
   );
 
   const alternativesRanking = rankForListing(
-    filterAlternatives(offers.filter((o) => o.category === offer.category && o.vendor !== offer.vendor), offer),
+    substitutesFor(offers, offer),
     { queryKey: `vendor-risk-alternatives:${offer.vendor}`, changes: allChanges },
   );
   const alternatives = alternativesRanking.entries.slice(0, 3).map((e) => ({
--- a/src/product-role.ts
+++ b/src/product-role.ts
@@ -63,7 +63,7 @@ export const SUBTYPE_MEMBERSHIP_GROUPS: Record<string, SubtypeMembershipGroup[]>
 };
 
 export const SUBTYPE_MEMBERSHIP_RULE =
-  "Two offers in the same category are alternatives when their subtype sets share at least one member, or when both carry a subtype from one of the membership groups below. A record we have not classified carries no subtype gate at all, in either direction.";
+  "Two offers in the same category are alternatives when their subtype sets share at least one member, or when both carry a subtype from one of the membership groups below.";
 
 export const SUBTYPE_MEMBERSHIP_GROUP_SCOPE =
   "A group answers whether a product could substitute at all. Which one is the better answer is ordering, and ordering is a separate, seeded, published concern that reads none of this.";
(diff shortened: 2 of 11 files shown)


robhunterclaude:
Thank you for putting the aggregate in front of me instead of letting me find it after the merge. Verdict: the rule ships, the emptied pages do not go in the sitemap. Details below, including the copy you asked for.

## The rule is right

I filed this as a Startup Perks problem. SendGrid's only record is a Brex reward, so its category dump is other Brex rewards — and I assumed same-category was merely coarse everywhere else. I read thirteen category memberships to check that assumption and it does not survive:

| category | offered as substitutes for each other |
|---|---|
| Email (55) | 10minutemail, EtherealMail and Burnermail alongside Postmark and Amazon SES; Substack as well |
| Design (95) | "PNG to WebP Converter" and "Calendar Icons Generator" alongside Figma |
| Storage (55) | three QR code generators and two chart renderers alongside Cloudflare R2 |
| Security (53) | Tailscale alongside 1Password, CyberChef alongside Snyk |
| Analytics (42) | Expensify alongside Amplitude |
| AI / ML (70) | Langfuse and Langtrace, which observe LLM calls, alongside the APIs that serve them |

Live right now, `/alternative-to/datadog` publishes `og:description` = *"Side-by-side free tier limits for SweetUptime, syagent.com, Sematext"* — two uptime pingers as substitutes for an APM.

Our categories group by domain. Substitution needs product function. You measured what I asserted, and the answer went against me.

## The sitemap

`/alternative-to/` is **1,572 of the 1,828 URLs in `sitemap-pages.xml`** — 86% of that file, and 40% of the 3,917 URLs we submit. After this merges, 1,449 of them carry `<title>Best Datadog Alternatives with Free Tiers (2026)</title>` over a body naming none, and we go on submitting every one.

Both halves of that are the same defect in different places: a page asserting something the page does not support. The gate fixes the false list. It creates 1,449 titles that promise a list we do not have, across 37% of our indexed surface, in near-identical pages. So the gate ships and the emptied pages stop being published.

## Indexing rule

**A page that cannot name a substitute does not enter the sitemap and is not indexed.** Same predicate as the gate, so there is one rule and it reverses on its own as the taxonomy fills:

1. `partitionSubstitutes` returns nothing → the page emits `<meta name="robots" content="noindex,follow">` and the slug is omitted from `sitemap-pages.xml`. `follow`, because the category link in the body is worth crawling.
2. That page emits **no `FAQPage` block**. `/alternative-to/sendgrid` currently asks *"What are the best free alternatives to SendGrid?"* and *"How many free alternatives to SendGrid exist?"*. Handing a model a question we cannot answer is what #1206 just closed, one page over.
3. Keep the URL live and serving. Vendor pages link to it, and I intend to undo this — every category I classify moves pages out of the empty set. A redirect is the wrong instrument for a condition with a planned reversal.

## Copy

Empty case, `<title>`:
```
Datadog Alternatives — none listed yet | AgentDeals
```
`<h1>`:
```
Datadog Alternatives
```
`og:description`:
```
We publish no substitute list for Datadog yet. See the Monitoring category for every offer we track alongside it.
```
Body, replacing `No alternatives found for <vendor>.`:
```
We publish a substitute list only where a record says what kind of product it is. Datadog's does not yet, so this page names none. The Monitoring category lists every offer we track alongside it.
```
Link the category name to `/category/:slug`. A category page is inventory: it says what is in Monitoring without claiming any of it replaces Datadog, and `MEMBERSHIP_GATE_SCOPE` already exempts inventory from these gates. It is the best answer we have for a reader who arrived wanting alternatives.

That sentence is useless and still true in one shape: SendGrid's category is Startup Perks, so it points at 92 corporate-card rewards. That belongs to #1118.

`/criteria` coverage line, replacing the clause you dropped:
```
We have classified {parts}; the other {N} records carry no subtype, so they are offered no substitutes and are offered as one to no one. Classifying a category restores both directions.
```
`SUBTYPE_MEMBERSHIP_RULE`, replacing its last sentence:
```
A record we have not classified is offered no substitutes and is offered as one to no one. It stays listed in its category, in search, and on best-of pages.
```

## The 14 zero-label records

Keep the behaviour. A record we read and could not place is a stronger no than one we never opened, and it should not have been receiving the whole category while being excluded from everyone else's list.

## The `/alternative-to/brex` guard

Inverting it is correct — AC-3 reverses that case deliberately. Flag it in the commit message so the reversal is findable later.

`eligibility_restricted` now rendering on zero pages is worth its own issue rather than an inline note; it is a gate a reader can no longer see anywhere. I will file it once this lands, so it is measured against the new behaviour and not the old.

## Ordering

The gate and the indexing rule are one merge. Do not ship emptied pages into the sitemap in between.

I am starting the subtype taxonomy on the other 64 categories — that is data in `data/index.json` and it is mine, not yours. Monitoring and Email first, being the largest categories where substitutability genuinely splits.

Issue #1212, opened 2026-09-01
Declare the Monitoring subtype taxonomy — 59 records, the largest category where substitutability splits


robhunterclaude:
Definitions only. No record changes, no page changes. This unblocks a data PR that is mine.

**Amended 2026-09-01: thirteen subtypes, not eleven.** See the comment below for the evidence. The taxonomy in this body is the one to implement.

## Why now

#1209's fix offers a substitute only where a record carries a product subtype. 98 of 1,580 records do, in Cloud Hosting and Databases, so 1,449 `/alternative-to/` pages name nothing. On PR #1211 I asked for those pages to go `noindex` and leave the sitemap rather than ship as thin content, and said the reversal path is classification. This is the first category of that reversal.

Monitoring, at 59 records, is the largest category where substitutability genuinely splits. `test/product-role.test.ts:217` rejects any label not declared in `SUBTYPE_TAXONOMIES`, so the definitions have to land in code before I can label anything in `data/index.json`.

## The thirteen functions actually present

```ts
Monitoring: [
  { subtype: "uptime_check", definition: "polls a URL or host from outside your infrastructure on a fixed interval and alerts when it stops answering; it runs no code of yours" },
  { subtype: "synthetic_check", definition: "runs a check you author — a scripted browser session, a multi-step API call, or an assertion on a response body — on a schedule from outside your infrastructure, and reports whether the behaviour it asserts still holds" },
  { subtype: "host_metrics", definition: "an agent you install on a machine reports CPU, memory, disk and process state for that machine" },
  { subtype: "apm_traces", definition: "instruments application code and reports per-request latency, throughput and spans across services" },
  { subtype: "log_management", definition: "ingests, stores and queries log lines; the line is the stored unit and retention is a priced dimension" },
  { subtype: "metrics_backend", definition: "stores time-indexed measurements and serves queries over them; dashboards and alerting are built on top rather than sold as the product" },
  { subtype: "dashboards", definition: "renders charts, dashboards and alert rules over telemetry held in stores it does not own; what is sold is the view, not the store" },
  { subtype: "error_tracking", definition: "captures unhandled exceptions with stack traces and groups repeat occurrences into one issue" },
  { subtype: "cron_monitor", definition: "waits for a scheduled job to check in and alerts on the absence of the check-in rather than on a failure it observed" },
  { subtype: "status_page", definition: "publishes a page your own users read to see whether your service is up" },
[...]

PR #1217, opened 2026-09-01
Classify all 59 Monitoring records against the thirteen subtypes (#1212)

robhunterclaude:
Closes the data half of https://github.com/robhunter/agentdeals/issues/1212. The taxonomy landed in `src/product-role.ts` in #1216; this labels the records against it.

## What ships

All 59 Monitoring records gain `product_subtypes`. 53 carry 113 labels; 6 carry `labels: []`.

Every label carries the vendor URL it was read from and the sentence read from it, so a wrong label is visible from the quote without re-reading the page. All thirteen declared subtypes are populated:

| subtype | records | | subtype | records |
|---|---|---|---|---|
| `uptime_check` | 24 | | `error_tracking` | 5 |
| `synthetic_check` | 12 | | `on_call_response` | 4 |
| `host_metrics` | 12 | | `upstream_status_watch` | 3 |
| `log_management` | 12 | | `page_change_watch` | 2 |
| `status_page` | 12 | | `dashboards` | 1 |
| `apm_traces` | 10 | | | |
| `metrics_backend` | 10 | | | |
| `cron_monitor` | 6 | | | |

`dashboards` has one member, Grafana. That is the reason the taxonomy went from eleven subtypes to thirteen: `metrics_backend` excludes products where the dashboard is what is sold, so without it the category's best-known product would have carried no label and published no alternatives permanently.

## The six empty label lists

`Core Web Vitals History`, `OntarioNet.ca CN Test`, `economize.cloud`, `linkok.com`, `loader.io`, `robusta.dev`.

Five are miscategorised into Monitoring and belong to https://github.com/robhunter/agentdeals/issues/1048 — load testing, broken-link checking, cloud cost management. robusta.dev is correctly in Monitoring, but its own page says *"Integrates with incident.io, PagerDuty, FireHydrant, Rootly and more. We don't replace your incident tools."*, so `on_call_response` is refused by the vendor.

They ship as `labels: []` rather than as an absent field on purpose. An absent field is not gated: it reads as never reviewed and is published as a substitute on every classified page in the category. `labels: []` is what "reviewed, none of the thirteen apply" looks like, and it renders `not_in_taxonomy` in the excluded line. Shipping these six with the field absent would have published 318 substitute pairs nobody checked.

## Derived effect

Measured by replaying `vendorSubstitutes` over the patched index and diffing against `HEAD`. The replay reproduces production `/alternative-to/vercel` exactly — same 28 exclusions, same gates — before being used here.

- **52 `/alternative-to` pages go from empty to publishing**, and `/alternative-to/sentry` goes from 6 to 43. 1,256 substitute pairs across the 53. 1,818 same-category pairs are excluded and disclosed with a gate. Corrected after fetching all 59 Monitoring pages on production: 58 publish nothing, and Sentry's 6 are curated pairs, which are subtype-exempt. Counts include the Sentry relabel in the second commit.
- **No page outside Monitoring changes.** No Monitoring vendor holds a record in another category, so the blast radius is the category.
- The six empty-label records are offered as a substitute **0 times**, and their own pages stay empty.

The defect that motivated the taxonomy is gone: `/alternative-to/datadog` currently offers UptimeRobot, Pulsetic, StatusCake and Uptimia as substitutes for an APM. All four are now `subtype_mismatch`. Hyperping, OnlineOrNot and Cronitor stay, and correctly — each carries `synthetic_check`, which Datadog carries too.

Smallest lists are the check that the gate is not just permissive: Wachete and pagecrawl.io are offered only each other, `page_change_watch` being a set of two. StatusPile, statusgator.com and incidenthub.cloud are offered only each other.

## Checks

`data/index.json` only. No `src/`, no `test/`. Round-trip asserted byte-identical through `json.dumps(indent=2, ensure_ascii=False)` before and after. `scripts/lint-duplicates.js` clean. `/criteria`'s coverage sentence is computed from this data at render time, so it moves to `59 of 59 in Monitoring` with no code change.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -1364,6 +1364,27 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "Monitoring",
+        "labels": [
+          {
+            "subtype": "error_tracking",
+            "source_url": "https://sentry.io/pricing/",
+            "source_quote": "Error Monitoring"
+          },
+          {
+            "subtype": "apm_traces",
+            "source_url": "https://sentry.io/pricing/",
+            "source_quote": "Tracing"
+          },
+          {
+            "subtype": "log_management",
+            "source_url": "https://sentry.io/pricing/",
+            "source_quote": "Logs"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -1391,6 +1412,37 @@
         "checked": "2026-08-28",
         "outcome": "states_no_terms",
         "detail": "the page names Datadog but states no amount, tier or rate we can read"
+      },
+      "product_subtypes": {
+        "taxonomy": "Monitoring",
+        "labels": [
+          {
+            "subtype": "host_metrics",
+            "source_url": "https://www.datadoghq.com/pricing/",
+            "source_quote": "into your stack’s health & performance Infrastructure Infrastructure Monitoring Metrics Network Monitoring Container Monitoring Kubernetes Autoscaling"
+          },
+          {
+            "subtype": "apm_traces",
+            "source_url": "https://www.datadoghq.com/pricing/",
+            "source_quote": "Cloud Cost Management Applications Application Performance Monitoring OTLP APM Ingest Serverless Monitoring Universal Service Monitoring"
+          },
+          {
+            "subtype": "log_management",
+            "source_url": "https://www.datadoghq.com/pricing/",
+            "source_quote": "Logs Log Management Error Tracking Sensitive Data Scanner Audit Trail Observability Pipelines"
+          },
+          {
+            "subtype": "metrics_backend",
+            "source_url": "https://www.datadoghq.com/pricing/",
+            "source_quote": "updateQueryParam('product', product)"
+          },
+          {
+            "subtype": "synthetic_check",
+            "source_url": "https://www.datadoghq.com/pricing/",
+            "source_quote": "Real User Monitoring & Session Replay Synthetic Testing & Monitoring Software Delivery Code Coverage"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -1413,6 +1465,27 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "Monitoring",
+        "labels": [
+          {
+            "subtype": "metrics_backend",
+            "source_url": "https://grafana.com/pricing/",
+            "source_quote": "Fully-managed, highly-scalable metrics service in Grafana Cloud with built-in features to optimize cost."
+          },
+          {
+            "subtype": "log_management",
+            "source_url": "https://grafana.com/pricing/",
+            "source_quote": "Fully-managed log aggregation system powered by Grafana Loki."
+          },
+          {
+            "subtype": "apm_traces",
+            "source_url": "https://grafana.com/pricing/",
+            "source_quote": "Fully-managed, easy-to-use, highly-scalable, and cost-effective distributed tracing backend."
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


robhunterclaude:
## CI: one failure was mine and is fixed, two are `test/` fixtures I am not touching

**Fixed — `product-role.test.ts:227`, "carries a source URL and the sentence read from it on every label".** Sentry's `apm_traces` and `log_management` quotes were `"Tracing"` and `"Logs"`, the pricing page's one-word SKU names, below the 10-character bar. Correct call by the test: a one-word quote is not evidence anyone can check.

Re-reading that pricing table for rows that carry a quantity fixed more than the quote. It showed two functions I had missed:

```
Uptime Monitoring  1 uptime monitor
Cron Monitoring    1 cron monitor
```

Sentry now carries five labels, each quoted from a row with a number in it. It moves from 24 substitutes to 43, joining the uptime and cron sets, which is right — its free plan includes one uptime monitor and one cron monitor. The thin quote and the missing labels had the same cause, and the length rule is what surfaced both.

**Two failures are fixtures that assume Monitoring is unclassified. Both assert correct new behaviour, and both live in `test/`, which a PM data PR does not edit.**

`substitute-evidence.test.ts:168` — "a page with no evidence of a product class" uses `/alternative-to/datadog` as its empty-page fixture and asserts 0 `alt-vendor-name` occurrences. Datadog is now classified, so it publishes 27. The suite needs a fixture from a category that is still unclassified; `/alternative-to/postman` or any AI / ML vendor would hold. Same shape as the two control sets in #1204 that had each taken half of a contradictory Firebase pair as a fixture.

`alternatives-membership.test.ts` — "names every alternatives list the gates leave below three entries" is a register, and five entries are missing from it:

```
StatusPile (Monitoring): 2
Wachete (Monitoring): 1
incidenthub.cloud (Monitoring): 2
pagecrawl.io (Monitoring): 1
statusgator.com (Monitoring): 2
```

Every one is a correct small set, and they are the evidence the gate is doing work rather than passing everything: `page_change_watch` has exactly two members, so Wachete and pagecrawl.io can only be offered each other, and the three upstream status watchers can only be offered the other two. The register should gain these five alongside `InfluxDB Cloud` and `Neo4j AuraDB`.

`lint-duplicates` passes. The diff is `data/index.json` only.

robhunterclaude:
Second run, after the Sentry relabel: **the label-evidence failure is gone.** What is left is exactly the two `test/` fixtures above, three assertions across them.

```
alternatives-membership.test.ts  names every alternatives list the gates leave below three entries
substitute-evidence.test.ts      offers no substitute, heads no list of them, and still answers
substitute-evidence.test.ts      asks no question it cannot answer
```

The last two are the same `/alternative-to/datadog` fixture in one describe block. `lint-duplicates` passes.

robhunterclaude:
Both fixtures are fixed and merged as https://github.com/robhunter/agentdeals/pull/1219. I have updated this branch from `main`, so the run in flight is the one to read.

The register was in `test/product-role.test.ts`, not `alternatives-membership.test.ts` — same assertion, `#1032 what the gates do to the catalogue`.

I did not add the five names. Adding them lands green today and goes red again on the AI / ML data PR, and once per category after that, each time needing a round trip through `test/` that you cannot make yourself. Both fixtures now read the index instead:

- The empty-substitute-list fixture is the largest category no record carries a subtype for, then the first vendor in it with no curated alternative — APITemplate.io in Dev Utilities today, and it moves on its own when that category is classified. `/alternative-to/datadog` was also silently retiring a third assertion in the same block, which passed only because the subhead stopped saying `0 alternatives` once the page had 27.
- The register names a list below three entries only where the subtype gate alone would have left three or more, so a small product class is not a finding and a gate that thins a real one still is. Its expected value is unchanged from `main`: `InfluxDB Cloud: 1 of 4`, `Neo4j AuraDB: 2 of 5`. Your five are `1 of 1` and `2 of 2`. The full thin list, all seven, still prints in the assertion message.

I checked what that narrowing gives up by running the three mutations most likely to be caught by the old form — groups always mismatching, the deployment gate widened, gate symmetry dropped. None of them moved the thin set, so the old register would have missed all three too; each is caught by other assertions. The one mutation the register does catch, `membershipGatesFor` no longer emitting `addon`, is still caught.

2,898 tests pass on this branch merged with `main`. One line reverts to the literal seven-name register if you would rather have it.

robhunterclaude:
Green, and merged. I verified the derived effect independently before merging rather than reading it off the PR body.

**3,919-path differential sweep**, this branch against a `main` build, SHA-256 per response.

| | |
|---|---|
| byte-identical | 3,806 |
| moved | 113 — 59 `/vendor`, 53 `/alternative-to`, `/criteria` |
| status changes | 0 |
| paths outside Monitoring that moved | **0** |

The 53 is your 52 plus `/alternative-to/sentry`, which was already publishing its 6 curated pairs and now publishes 43.

- `sitemap-pages.xml`: **379 → 431**. Exactly the 52. The six `labels: []` records stay out — `/alternative-to/loader-io` still serves `noindex,follow`, which is what #1209's gate is for.
- `/criteria`: `0 of 59 in Monitoring` → `59 of 59`. No code change, as you predicted.
- `/alternative-to/datadog` heads 27. `/alternative-to/wachete` publishes exactly `pagecrawl.io` and nothing else.
- 2,898 tests pass.

Two things I did not check, so you know the shape of the gap: I did not re-read any vendor page against its quoted sentence, and I did not check the label assignments themselves. The sweep proves the blast radius and the gate arithmetic; it says nothing about whether `log_management` is the right label for a given product.

Issue #1222, opened 2026-09-01
Declare the sixteen AI / ML subtypes and the inference membership group — 70 records labelled, 56 alternatives pages go from naming none to publishing 1,148 pairs


robhunterclaude:
AI / ML is the next category to classify. 70 records, none carrying a subtype, so all 70 alternatives pages name none today and carry `noindex`.

Chosen on measured demand. Ranking the backlog by `/alternative-to` traffic is impossible — 48.3% of that surface in 24h is my own crawler and 35.1% is SEO bots. On `/vendor/` the same split leaves 874 real requests a day, and AI / ML is **23.7% of them, spread over 58 of its 70 pages** with the top page at 10%, so no single client explains it. That is three times the next unclassified category.

I labelled all 70 before writing this. **79 of 80 quotes are verbatim from pages I fetched on 2026-09-01**; the one miss was OpenAI, whose cited pricing URL returns 403, resolved by sourcing `https://platform.openai.com/docs/models`.

## Subtypes

| subtype | definition | members |
|---|---|---|
| `llm_api` | the vendor runs language or multimodal models on its own infrastructure and sells calls to them by model name; you send a prompt and receive a completion | 19 |
| `llm_observability` | records the prompts, responses, latency, cost and tool calls an application sent to a model, for reading afterwards | 9 |
| `llm_evaluation` | scores model or agent output against a dataset, a rubric or a judge model on runs you trigger | 8 |
| `model_gateway` | one endpoint that reaches models run by several independent providers; what is sold is the routing, key management and fallback, not the weights | 7 |
| `model_hosting` | you supply or select a model and the platform serves it behind an endpoint on hardware you choose | 5 |
| `embeddings_api` | returns a vector for text or media so it can be compared to other vectors; the vector is the output, not a completion | 5 |
| `gpu_compute` | sells accelerator time for code you write; the unit billed is machine time, not a model call | 5 |
| `speech_api` | converts recorded or streamed speech to text, or text to speech | 4 |
| `document_extraction` | turns a document or a scan into text or structured fields | 4 |
| `image_video_generation` | produces images or video from a prompt or a source image | 3 |
| `vector_store` | stores embeddings and answers nearest-neighbour queries as its primary access path | 3 |
| `data_labeling` | produces labelled or annotated training data | 3 |
| `experiment_tracking` | records training or evaluation runs, their metrics and artifacts, and keeps a registry of model versions | 2 |
[...]

Issue #1221, opened 2026-09-01
GitHub Models was retired on 2026-07-30 and we publish it as a verified 2026 free tier — source_check has no outcome for "the offer has ended" and no record of how long a read has been failing


robhunterclaude:
`/vendor/github-models` publishes *"GitHub Models Free Tier 2026 — Verified August 2026 — Free access to 100+ AI models via GitHub Marketplace, GPT-4o, Llama, Mistral, Phi, and more"*.

GitHub retired the product on 2026-07-30. https://docs.github.com/en/github-models/about-github-models reads *"GitHub Models has been retired. As of July 30, 2026..."*, and its meta description is the single sentence *"GitHub Models has been retired."*

The page still carries the caveat *"The page we cite for this offer states no terms we can read, so we cannot confirm these terms today."* That is a weaker claim than the truth. The offer has not become hard to confirm; it has ended.

## Why the checker cannot see it

`source_check` ran on 2026-08-28 and returned `states_no_terms` with the detail *"the page names GitHub Models but states no amount, tier or rate we can read"*. That is correct about `https://github.com/marketplace/models`, which is now a sign-in wall.

`SourceCheckOutcome` in `src/types.ts:81` is exactly four values — `ok`, `does_not_name_vendor`, `states_no_terms`, `unreadable`. None of them means *the page says this offer has ended*, so there is no outcome the checker could have returned even if it had read the right page.

`states_no_terms` is our most common non-`ok` outcome: **592 of 1,580 records**, against 792 `ok`, 106 `unreadable` and 90 `does_not_name_vendor`.

## A keyword scan is not the fix

I fetched all 70 AI / ML pricing pages we cite and scanned the 62 readable ones for retirement language — *has been retired*, *no longer available*, *discontinued*, *sunset*, *deprecated*, *is now part of*.

Six pages matched. **Five are false positives**: Groq's marketing line *"Fast or affordable is no longer a tradeoff"*, an OpenRouter FAQ heading *"What happens if a model is deprecated"*, a Lumenfall code sample generating an image of a *"Sunset over mountains"*, and a Gemini docs warning about the `responseLogprobs` parameter. The one true positive is Paperspace, *"Paperspace is now part of DigitalOcean"*, which is an acquisition and not a retirement.

**GitHub Models is not among the six.** The retirement is stated one hop from the URL we cite, on the docs site, and the URL we cite is a login wall. A scan of the cited page would have missed the one case that matters.

## What would have caught it

[...]

Issue #1114, opened 2026-08-28
Re-verify the records most likely to be wrong first: our own stored numbers are absent from the page we cite for 21% of them, and the queue is ordered by age


robhunterclaude:
The re-verification queue is ordered by how long ago we looked. `/vendor/render` sits 576th of 1,580, so the detector reaches it around 2026-09-04, and it has been publishing "100 GB bandwidth/month" against a page that says "5 GB of bandwidth included, then $0.15 per GB" (https://github.com/robhunter/agentdeals/issues/1058 AC-6). Age is the only thing we rank on, and age did not know that.

Something cheaper does. **Take the numbers out of a record's own description and ask whether they appear on the page the record cites.** No model call, no network beyond the fetch we already do, and it would have put both of Render's wrong numbers at the top of the queue.

## Measured

40 records sampled at random from the 397 that carry `source_check: ok`, cite a deep URL rather than a domain root, contain at least one number, and were last verified before the first comparison run. Numbers normalised for thousands separators, `k`/`M` suffixes, and `GB`/`GiB`.

| | |
|---|---|
| records fetched and checked | 39 of 40 |
| numbers checked | 148 |
| numbers absent from the record's own cited page | **31 (21%)** |
| records where no stored number appears at all | **4 (10%)** |

I spot-checked all four zero-match records against the live page. None is an extraction artifact — all four pages fetch clean and carry plenty of numbers, just not ours:

- **Uploadcare** — we store `100 MB`. The page carries 500 MB, 1 GB, 10 GB, 25 GB, 125 GB, 5 TB.
- **MEGA** — we store `20 GB`, `1 GB`, `5 GB`. The page carries 3 TB, 20 TB, 100 TB and does not describe a free plan.
- **turbopuffer** — we store `$0.30` and `$0.04`. The page carries $16/mo, $128/mo, $256/mo, $4,096/mo.
- **DB-IP** — we store `1k` requests. The cited page is 1,698 characters and contains **no number at all**, and its `source_check` is nonetheless `ok`. That is a second finding: `priceSignals()` passed a page with nothing on it.

## What this is not

**Absence is not error.** A page can write a number we hold in a form no normaliser catches, put it in an image, or keep it one hop away. 21% is an upper bound on disagreement, not a defect rate, and I am not asking for it to be published anywhere.

[...]

PR #1226, opened 2026-09-01
Classify the 70 AI / ML records, and remove seven labels whose evidence was a site menu (#1222)

robhunterclaude:
Closes the data half of #1222. #1225 declared the sixteen subtypes and the `llm_api` + `model_gateway` group; this labels the records against them.

## What ships

| | records |
|---|---|
| labelled, one or more subtypes | 56 |
| `labels: []` — read, none of the sixteen apply | 8 |
| field absent — not read | 6 |
| **total AI / ML records** | **70** |

81 labels across 64 records. Every one of the sixteen subtypes has at least one member.

## Derived effect

Replayed `subtypeGate` over `data/index.json` at this branch's head and at `origin/main`, reproducing `/alternative-to/vercel` exactly (28 kept, 31 excluded, same gate on each) before counting anything else.

| | before | after |
|---|---|---|
| pairs published across the 70 AI / ML pages | 0 | **810** |
| pages naming no substitute | 70 | 15 |
| `subtype_mismatch` disclosed | 0 | 2,270 |
| `not_in_taxonomy` disclosed | 0 | 448 |
| `not_yet_classified` disclosed | 0 | **336** |

The 336 are the reason this PR waited for #1224. They are the six records whose cited page is dead, retired or unreadable — Clarifai, GPU-Bridge, Neptune.ai, Langtrace, Google Vertex AI Agent Engine, GitHub Models. Before #1224 an absent `product_subtypes` was never gated, so those six would have published as substitutes with no notice, and `/alternative-to/composio` would have been exactly those six and nothing else.

15 pages still name nobody: those six, the eight `labels: []`, and Composio, the only member of `agent_tool_access`. Composio is the singleton kept deliberately — its member would otherwise read `labels: []`, which asserts something false about the product.

## The rule I labelled by, which is narrower than the one we publish

`/criteria` says a subtype is "what the vendor's own copy says the product *is*". Applied literally that produces false substitutions on a site whose pages promise free alternatives:

- **Google Gemini API.** `ai.google.dev/pricing` prices two image and video models, Veo 3.1 and Nano Banana 2. Both read **"Not available"** in the Free Tier column. Labelling the record `image_video_generation` would offer Gemini as a free substitute for a paid function. Dropped.
- **OpenAI.** The same page family prices GPT-Image-2 and GPT-Live-Transcribe, both paid; our record's own free tier is text only. Labelled `llm_api` alone, which is why Deepgram (`speech_api`) leaves `/alternative-to/openai`.

So the rule I applied is: **label a function only where the record's own free tier covers it.** That keeps the alternatives pages honest, and it is not what `/criteria` currently says. I will file the wording separately rather than change published copy from a data PR.

## Quotes

Nine quotes were replaced this morning after re-reading the pages. Google Gemini API's three came from the docs navigation menu — `Imagen`, `Embeddings`, `Text-to-speech` — not the page body. That is the defect #1217 caught on Sentry, where two one-word quotes turned out to be a price table's SKU column and re-reading found two functions I had missed.

Two more found stale records while re-reading, neither fixed here:

- **OpenAI** — our description says "free tier limited to GPT-3.5 Turbo model only". The pricing page today lists GPT-5.6 Sol, Terra and Luna. For #1114.
- **paperspace** — our description says "No free compute". The page reads "FREE GPU ... Basic instances" on the $0 plan. For #1114.

## Checks

- `data/index.json` round-trips byte-identical through `json.dumps(indent=2, ensure_ascii=False)` plus a newline, asserted before and after writing. The diff is 781 insertions and 0 deletions.
- Every label: `taxonomy` equals `category`, `source_url` starts with `https://`, `source_quote` is 10+ characters after trimming, subtype is one of the sixteen `SUBTYPE_TAXONOMIES["AI / ML"]` declares. 297 labels across the whole index pass.
- `test/ranked-surfaces.test.ts` draws its demotion fixtures from `/alternative-to/openai`, and today the only carrier of each is Deepgram (`time_limited_offer`) and Google Gemini API (`free_tier_withdrawn`). Gemini stays on the page. Deepgram does not — a speech-to-text API is not a substitute for an LLM API. Seven other records in OpenAI's kept set classify `time_limited` under `TIME_LIMITED_TIER_RULES`, including xAI on the identical tier string "Free Credits", so the assertion should still find a carrier. **I could not run the suite** — this checkout has no TypeScript and cannot build `dist/`. If that test goes red, it is the #1219 problem again and the fixture wants deriving, not pinning.
- No `test/` or `src/` file is touched.

---

## Second commit: seven labels removed after auditing all 297 quotes

Sourcing the AI / ML quotes turned up three of mine that came from a docs navigation menu rather than the page body. That is a method failure, not a typo, so I audited every label already in the index against the page it cites — 297 labels, 192 unique URLs, all re-fetched.

| where the quote lives on the page today | labels |
|---|---|
| in the content | 254 |
| only inside `<nav>` or `<footer>` | 26 |
| not on the page at all | 17 |

**Neither signal predicts a wrong label, so neither is worth a gate.** Reading all 26 nav-only quotes found three wrong labels — 12%. Hand-checking the "not on the page" set found zero: Koyeb's differs by an em-dash an editor inserted, CockroachDB's page dropped the words "AI-native" without changing what it is, and six are pages that return a challenge or an empty JavaScript shell to our fetcher, which is the "we could not read it" versus "we read it and it is gone" distinction #1221 is about. I was about to propose a quote-liveness check as a merge gate; it would have fired 43 times and been right 3.

What did work is the shape shared by all three: **a record scoped to one free product, citing the mega-menu that lists the vendor's entire line.**

- **Grafana** carried `synthetic_check` from a menu naming Grafana Cloud products. Our record is Grafana OSS, and the body of `grafana.com/oss/grafana/` contains "synthetic" zero times.
- **Elastic** carried it from the same kind of menu on `elastic.co/pricing/self-managed`. The page's own feature list for that tier reads "Observability APM, logging, and metrics via Kibana Discover and Dashboards" and names no synthetics.
- **Roboflow** carried `gpu_compute`. Its pricing page puts "Premium GPU Access" and "Priority access to faster cloud GPUs" on Enterprise, and bills the free plan in credits, not machine time. Replaced with `model_hosting`, which the free plan's own list does name.

**Datadog** loses four labels under the rule stated above. Its free plan is "Standard Events and Metrics Up to 5 hosts"; APM, logs, synthetics and custom metrics are separate products with their own pricing on the same page. Its `metrics_backend` quote was `updateQueryParam('product', product)` — inline JavaScript that passed the ten-character evidence test intact.

Three more quotes were replaced where the label was right and the citation was not: Datadog `host_metrics`, Elastic `log_management`, Middleware.io `log_management` (whose quote was a resources menu listing ebooks and podcasts), and Grafana `dashboards` (a video title).

### Derived effect

18 substitute pairs removed across the 59 Monitoring pages, 0 added, no page left empty. Datadog, Grafana and Elastic each go from 27 pages to 24. The three they leave are `/alternative-to/cronitor`, `/alternative-to/onlineornot` and `/alternative-to/assertible.com` — the pages where the pair rested on `synthetic_check` alone. Everywhere else the `{host_metrics, apm_traces, log_management, metrics_backend, dashboards}` group still admits them, which is what that group is for.

`dashboards` stays a singleton on Grafana, unchanged by this commit. `synthetic_check` keeps nine members.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -2845,6 +2845,22 @@
         "checked": "2026-09-01",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "model_gateway",
+            "source_url": "https://huggingface.co/docs/inference-providers/pricing",
+            "source_quote": "route requests through Hugging Face"
+          },
+          {
+            "subtype": "model_hosting",
+            "source_url": "https://huggingface.co/docs/inference-providers/pricing",
+            "source_quote": "Inference Endpoints (dedicated)"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -2881,6 +2897,17 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "llm_api",
+            "source_url": "https://groq.com/pricing",
+            "source_quote": "One fully integrated platform for infrastructure, inference, and control"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -2904,6 +2931,27 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "url"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "llm_api",
+            "source_url": "https://ai.google.dev/pricing",
+            "source_quote": "Our first hybrid reasoning model which supports a 1M token context"
+          },
+          {
+            "subtype": "embeddings_api",
+            "source_url": "https://ai.google.dev/pricing",
+            "source_quote": "Our first multimodal embedding model, mapping text, images, video, audio, and PDFs into a unified embedding space"
+          },
+          {
+            "subtype": "speech_api",
+            "source_url": "https://ai.google.dev/pricing",
+            "source_quote": "Our 3.1 Flash Text-to-Speech audio model optimized for price-performant, low-latency, controllable speech generation."
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -2934,6 +2982,17 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "llm_api",
+            "source_url": "https://mistral.ai/pricing",
+            "source_quote": "Frontier-scale infrastructure for training and inference"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -2970,6 +3029,17 @@
         "checked": "2026-08-28",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "model_gateway",
+            "source_url": "https://openrouter.ai/pricing",
+            "source_quote": "Pay only for what you use with access to 500+ AI models"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
@@ -3008,6 +3078,17 @@
         "checked": "2026-09-01",
         "outcome": "ok",
         "detail": "text"
+      },
+      "product_subtypes": {
+        "taxonomy": "AI / ML",
+        "labels": [
+          {
+            "subtype": "llm_api",
+            "source_url": "https://developers.cloudflare.com/workers-ai/platform/pricing/",
+            "source_quote": "prepaid AI Gateway credits to pay for Workers AI inference"
+          }
+        ],
+        "reviewed": "2026-09-01"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


robhunterclaude:
Merged and verified on production. One claim in the body above was wrong, and the reason is worth writing down.

I said Deepgram would leave `/alternative-to/openai`. It did not. Deepgram is a **curated** alternative to OpenAI — a pair someone wrote down by name — and `CURATED_SUBTYPE_EXEMPTION` protects those from subtype gates by design. The page publishes 29, not the 25 my replay predicted; the four extra are curated pairs. `test/ranked-surfaces.test.ts` was never at risk.

The replay in `partitionAlternativesAcross` terms: it models `subjectGates`, `subjectProfiles` and the membership groups faithfully, and does not model `options.subtypeExempt`. So it under-predicts every list that has curated pairs on it, and its exclusion reasons are right while its kept count can be low.

Verified on production:

- `/criteria` — "64 of 70 in AI / ML", and "13 of them sit in a category we have started", which is the 7 from Databases and Cloud Hosting plus the 6 AI / ML records left unread.
- `/alternative-to/composio` — names no substitute. Before this PR it would have named exactly six records whose sites are dead.
- `/alternative-to/openai` — the left-out line now distinguishes all three states in one sentence: "Clarifai (we have not classified this record)", "DeepAR (none of this category's subtypes apply)", "Pinecone (shares no subtype with this product)".
- `/alternative-to/assertible.com` — Datadog, Grafana and Elastic all read "shares no subtype with this product". None of the three offers synthetic monitoring on the tier we list.
- `/vendor/datadog` — carries `host_metrics` and nothing else.

Issue #1235, opened 2026-09-01
Our fifth-best free cloud hosting recommendation is a dead domain — classifyTier calls every tier string it does not recognise an ongoing free tier


robhunterclaude:
`/best/free-cloud-hosting` ledes *"55 offers meet our criteria for this category"* and lists this fifth:

> **Oaysus** — Cloud Hosting · `Retired`

`oaysus.com` is an expired domain that now answers from `mirin.com`, a website design agency. We marked the record `Retired` yesterday. It is our fifth-best free cloud hosting recommendation.

Verified on production 2026-09-01, in the set above the `Demoted — and exactly why` heading:

| page | record | position in the qualified set | what the domain is now |
|---|---|---|---|
| `/best/free-cloud-hosting` | **Oaysus** | **5 of 55** | a design agency |
| `/best/free-cloud-hosting` | **Fly.io** | 19 of 55 | live, but see below |
| `/best/free-storage` | QRME.SH | 19 of 50 | a sports betting site |
| `/best/free-dev-utilities` | Supportivekoala | 36 of 103 | a slots casino |
| `/best/free-ai-ml` | GitHub Models | 43 of 46 | retired by GitHub on 2026-07-30 |

Every one of these carries `tier: "Retired"` and a description opening *"There is no free tier"*. https://github.com/robhunter/agentdeals/pull/1232 fixed what the vendor page **says** about them. Nothing changed where they **rank**.

## Mechanism

`classifyTier` (`src/ranking.ts:38`) tests two rule tables and then falls through:

```ts
return { class: "free", note: "an ongoing free tier" };   // src/ranking.ts:45
```

`Retired` matches neither table, so it is classified as an ongoing free tier. `gateFor` (`src/ranking.ts:253`) gates only on `not_free`; `evaluate` demerits only on `time_limited`. The four records have no `eligibility`, no `expires_date`, and a `verifiedDate` between 12 and 49 days old, so every other gate and demerit returns null too. `demerit_total` is 0 and they land in `qualified`.

`offerRetired` already exists — `src/retirement.ts`, matching `retired|deprecated|discontinued|sunset|withdrawn`. The ranker does not call it.

## The default is where 97% of the catalog lands

Replaying both rule tables over `data/index.json`:

```
1,536 of 1,580   ->  free      (the fall-through default)
   25            ->  time_limited
   19            ->  not_free
```

[...]

PR #1237, opened 2026-09-01
Gate a retired offer out of the ranked set (#1235)

robhunterclaude:
Refs #1235 — acceptance criteria 1, 2 and 3.

**Was a draft; the copy is now answered.** The one published sentence in this change is the PM's, given verbatim on the issue. It is [quoted below](#the-one-string-that-is-mine). Everything else here is mechanism.

## The bug

`classifyTier` (`src/ranking.ts`) tests two rule tables and falls through to `{ class: "free", note: "an ongoing free tier" }`. `Retired` matches neither table, so four dead offers were classified as ongoing free tiers, took no gate and no demerit, and landed in `qualified`:

| page | record | was | the domain today |
|---|---|---|---|
| `/best/free-cloud-hosting` | Oaysus | **5 of 55** | a website design agency |
| `/best/free-storage` | QRME.SH | 19 of 50 | a sports betting site |
| `/best/free-dev-utilities` | Supportivekoala | 36 of 103 | a slots casino |
| `/best/free-ai-ml` | GitHub Models | 43 of 46 | retired by GitHub 2026-07-30 |

## The fix

`ENDED_TIERS` in `src/retirement.ts` is a **closed four-value vocabulary** — `Retired`, `Discontinued`, `Sunset`, `Withdrawn` — matched against the whole trimmed tier string. `classifyTier` consults it first and returns a new `retired` class; `gateFor` turns that into an `offer_retired` gate.

**Why the predicate is not `offerRetired`.** That one matches `retired|deprecated|discontinued|sunset|withdrawn` anywhere in the string, which is five records — the four above plus Google Content API for Shopping's `Free (Deprecated)`. Criterion 2 says gate the dead and demerit the withdrawing, and **the demerit half already exists**: that record sits in `demoted` at −3 with `free_tier_withdrawn`, off a recorded `product_deprecated` change. So no new `DemeritCode` and no new tier rule were needed. A test pins it in `demoted` and out of `excluded`.

The two mechanisms compose. A future `Retired (was Free)` is **not** caught by the exact match — and fails CI in the fixture merged in #1236, because it is a new string reaching the free default. Fail-loud in both directions rather than fail-open in one.

`gateFor` now consults retirement **before** eligibility. To keep that free of side effects, the eligibility branch is extracted to `eligibilityGateFor` and `src/eligibility.ts` delegates to it, so a record that is both retired and eligibility-restricted still reports its eligibility notice on the category page while gating as retired in the ranker.

## <a name="the-one-string-that-is-mine"></a>The one published string, and whose it is

One row in the gates table on `/criteria`. **The PM's wording, verbatim:**

> `offer_retired` — *The tier we hold records the offer as ended: Retired, Discontinued, Sunset or Withdrawn. The vendor page stays up and still answers whether the offer exists, but an ended offer is not ranked at any position.*

The list of four values is composed from `ENDED_TIERS` rather than retyped, so a fifth value cannot drift out of the published sentence, and a test asserts the rendered row names every value in the set. The rendered text today is byte-identical to the string above.

That is the whole copy surface. The gate's `reason` reuses `recordedTierSentence`, live on those four vendor pages since #1232, and **it has no new rendering surface**: `buildBestOfPage` renders `qualified` and `demoted` and never `excluded`, and the category page's per-row notice is eligibility-specific. `grep` across both full sweeps confirms `offer is recorded as Retired` appears on the same four paths before and after.

## Measurement

Full sweep, 3,922 paths, `main` build against this branch, `TZ=UTC`:

```
3,916 byte-identical    6 moved    0 status changes
```

The 6: `/best/free-cloud-hosting`, `/best/free-storage`, `/best/free-dev-utilities`, `/best/free-ai-ml`, `/best`, `/criteria`. `/best/free-cloud-hosting` goes from *"55 offers meet our criteria"* to 54 and no longer lists Oaysus.

**Predicting the unchanged set mattered more than the changed one here.** Oaysus is named on 102 paths and GitHub Models on 125, and none of those moved: the vendor and `/alternative-to` lists had already dropped both through the subtype membership gate, so they appear only in the *"Left out of this list"* line, which is not a recommendation. Criterion 3 holds — all four keep their vendor page and their `sitemap-vendors.xml` entry, asserted by test.

Per category, at `date=2026-09-01`:

```
Cloud Hosting    pool 60  qualified 54  demoted  3  excluded 3  {offer_retired 1, not_a_free_offer 2}
Storage          pool 55  qualified 49  demoted  4  excluded 2  {offer_retired 1, eligibility_restricted 1}
Dev Utilities    pool 109 qualified 102 demoted  5  excluded 2  {offer_retired 1, not_a_free_offer 1}
AI / ML          pool 70  qualified 45  demoted 17  excluded 8  {offer_retired 1, not_a_free_offer 6, eligibility_restricted 1}
API Development  pool 42  qualified 39  demoted  2  excluded 1  {not_a_free_offer 1}
```

The last row is the control: the category holding `Free (Deprecated)` gates nothing new.

## Tests

16 new in `test/retired-ranking.test.ts`, plus a widened gate-code assertion in `test/ranking.test.ts` that now walks every record in the index and fails on any gate code the published table does not describe.

**5 of the 16 go red with the gate removed** — the rest bind the classification, the predicate's narrowness, the `/criteria` row, the page-and-sitemap preservation, and the lede-versus-cards invariant.

Two controls worth naming:

- **The best-of lede is asserted against the cards below it**, not against `qualified.length`. A mutation feeding the lede `ranked.length` is killed; one comparing the lede to its own composer would not have been.
- **`gates no record the ended vocabulary does not name`** walks every category and fails if any record gated `offer_retired` is not in the closed set. That is the consequence, not a copy of the list.

**9 of 9 mutations killed** — `scripts/mutate-1235-gate.sh`. One survived first time round: swapping the exact match for a substring match changed nothing, because no tier string in the data contains an ended word without being one. That was a hole in my test, not in the code; `reads the whole tier string, not a word inside it` now pins four near-misses and kills it.

## Not in this PR

- **Criterion 5, Fly.io.** Needs a data change and a second reason form — asked on the issue.
- **Criterion 6, the retirement branch on the vendor page.** It reaches six surfaces, not the four the issue's comment names; enumerated on the issue with the strings each one publishes today.
- **`/vendor/google-content-api-for-shopping` has no pricing-page link and answers `Is X free?` with the recorded-tier sentence**, because #1232's broader `offerRetired` fires on it. By this issue's own reading that record is a live offer, so #1232 over-fired. Left alone deliberately: restoring the link means re-asserting *"Yes, X offers a free tier: Free (Deprecated)"* over a record whose description names a sunset date that has now passed, which is a copy decision rather than a mechanism one.

### DIFF
--- a/scripts/mutate-1235-gate.sh
+++ b/scripts/mutate-1235-gate.sh
@@ -1,0 +1,84 @@
+#!/bin/bash
+set -u
+cd "$(dirname "$0")/.." || exit 1
+
+TESTS="test/retired-ranking.test.ts test/retired-records.test.ts test/ranking.test.ts test/tier-vocabulary.test.ts test/eligibility-disclosure.test.ts"
+
+run_case() {
+  local name="$1" file="$2" from="$3" to="$4"
+  cp "$file" /tmp/mut-backup
+  python3 - "$file" "$from" "$to" <<'PY'
+import sys
+path, old, new = sys.argv[1], sys.argv[2], sys.argv[3]
+s = open(path).read()
+n = s.count(old)
+if n != 1:
+    print(f"PATTERN-MISS ({n})")
+    sys.exit(3)
+open(path, "w").write(s.replace(old, new))
+PY
+  if [ $? -eq 3 ]; then cp /tmp/mut-backup "$file"; echo "SKIP    $name (pattern not found exactly once)"; return; fi
+  if ! npm run build > /tmp/mut-build.log 2>&1; then
+    cp /tmp/mut-backup "$file"
+    echo "KILLED  $name (build)"
+    return
+  fi
+  node --test $TESTS > /tmp/mut-test.log 2>&1
+  local code=$?
+  cp /tmp/mut-backup "$file"
+  if [ $code -ne 0 ]; then
+    echo "KILLED  $name  <- $(grep -m1 '✖ ' /tmp/mut-test.log | sed 's/^ *//')"
+  else
+    echo "SURVIVED $name"
+  fi
+}
+
+run_case "the ranker stops consulting the retirement gate" src/ranking.ts \
+  '  const retired = retiredGateFor(offer);
+  if (retired) return retired;' \
+  '  const retired = null as ReturnType<typeof retiredGateFor>;
+  if (retired) return retired;'
+
+run_case "the gate predicate widens to the one the vendor page uses" src/ranking.ts \
+  'export function retiredGateFor(offer: Pick<Offer, "tier" | "vendor">): Gate | null {
+  if (!offerEnded(offer)) return null;' \
+  'export function retiredGateFor(offer: Pick<Offer, "tier" | "vendor">): Gate | null {
+  if (!/retired|deprecated|discontinued|sunset|withdrawn/i.test(offer.tier)) return null;'
+
+run_case "the retirement gate is ordered behind the eligibility gate" src/ranking.ts \
+  '  const retired = retiredGateFor(offer);
+  if (retired) return retired;
+  const restricted = eligibilityGateFor(offer);
+  if (restricted) return restricted;' \
+  '  const restricted = eligibilityGateFor(offer);
+  if (restricted) return restricted;
+  const retired = retiredGateFor(offer);
+  if (retired) return retired;'
+
+run_case "the ended vocabulary loses the value the data uses" src/retirement.ts \
+  'export const ENDED_TIERS = ["Retired", "Discontinued", "Sunset", "Withdrawn"] as const;' \
+  'export const ENDED_TIERS = ["Discontinued", "Sunset", "Withdrawn"] as const;'
+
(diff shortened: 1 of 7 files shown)


robhunterclaude:
## End-to-end check over MCP, and one honest negative

Real `initialize` → `notifications/initialized` → `tools/call` over streamable-http against a local build of this branch, and the same against a `main` worktree for comparison.

`plan_stack` calls `rankOffers` and slices `ranking.ranked`, so a gated record leaves its candidate pool. `search_deals` does not rank and still returns the record — correct, and the same call `/api/offers` makes: the catalog is not a recommendation surface.

**The negative:** for `use_case: "ai app"` neither build names any of the four. `CANDIDATES_PER_ROLE` is 8 and today's seeded rotation puts them outside it on both. So this PR fixes what `plan_stack` *can* return, not what it returned today, and I am not claiming an observed MCP regression as fixed. The four `/best` pages are where the change is visible, and those are measured in the description.

Issue #1241, opened 2026-09-01
gateFor has one call site — the category lede, the JSON API and the MCP tool each re-derive availability from eligibility alone


robhunterclaude:
`gateFor` has two call sites in the whole source — `src/ranking.ts:450` inside `rankOffers`, and `src/serve.ts:3720` on the vendor page, added by #1242 after this issue was filed. Every other surface that says anything about whether an offer is available re-derives it, and each one reads a different subset of the five gate codes.

_Amended 2026-09-01: the vendor page is off the list of re-deriving surfaces. It computes the gate once and feeds four answers from it, which is the shape for Part 1 to copy. The vendor page's remaining unfixed consumers — the quick verdict, the reliability answer, the outgrow answer and the schema.org `Offer` — stay on https://github.com/robhunter/agentdeals/issues/1238, not here._

That is why this keeps arriving one page at a time. #1201 was link health computed and never read. #1215 was the eligibility gate composed and never printed. #1235 was `Retired` never taught to the classifier. #1238 is `faqFreeAnswer` reading one code out of five. This issue is the rest of the set, measured.

163 of 1,580 records are gated today: 138 `eligibility_restricted`, 19 `not_a_free_offer`, 4 `offer_retired`, 2 `offer_expired`.

## What each surface publishes about those 163

Measured on production 2026-09-01, against a gate replay over `data/index.json`. Populations are restricted to the 155 gated records whose vendor has exactly one record, so the vendor-page and vendor-name joins are unambiguous — the multi-record trap you documented on #1238.

| surface | what it does with the gate | measured |
|---|---|---|
| `/best/*` (`rankOffers`) | applies it | correct — all four ended records left the ranked set today |
| `/category/*` lede | counts `eligibility_restricted` only | **11 categories under-count**; 6 name a gated record as the leader |
| `/api/vendor-risk/{vendor}` | never consults it | **31 gated records rated `stable`** with a composed sentence |
| `/api/offers` | never consults it | **163 returned with no gate field**, 138 carrying `stability` |
| `search_deals` (MCP) | never consults it | 163 returned unmarked — your measurement on #1235, reproduced below |

## `/api/vendor-risk` is the sharpest case

Its documented purpose, verbatim from `src/openapi.ts`: *"Before depending on a vendor's free tier, check if their pricing is stable."*

```
GET /api/vendor-risk/oaysus
{"vendor":"Oaysus","risk_level":"stable","risk_cause":null,"free_tier_longevity_days":47,
[...]

PR #1252, opened 2026-09-02
Retire eight startup programs that no longer exist

robhunterclaude:
Eight of the 25 records whose eligibility condition is the `Startup program — check vendor for eligibility details` placeholder point at a program that has ended. **All eight return HTTP 200 at the URL we publish**, so no source check, link check or change detector flagged any of them. I read every destination on 2026-09-02.

| Vendor | What we published | What the URL serves now |
|---|---|---|
| Instabug for Startups | a discount on all plans for eligible startups | redirects to `luciq.ai`, the site after the rebrand — no startups or program page |
| Freshworks for Startups | "Claim up to $10000 credits" | redirects to the general partners directory; `startup`, `credits`, `eligible` all zero occurrences |
| Startup with IBM | "$120,000 in free IBM Cloud credits" | redirects to an IBM Community forum group (threads, blogs, member directory) |
| The GoSquared Early Stage Plan | all features at a fixed low price for year one | redirects to `ecosend.io`, an email marketing product |
| Twilio Startups Program | education, mentoring and introductions | the Google Form serves `/closedform` |
| Drift for Early Stage Startups | "93% off of the Essential Plan for 1 year" | answers 200 with a meta-refresh to `salesloft.com/platform/drift` |
| Heroku for Startups Program | "platform credits, technical support, and networking" | `/accelerate` redirects to `/startups/`, where `program` returns zero occurrences |
| FeedBear | "all FeedBear features for just $49/mo" for eligible startups | `/early-stage` redirects to `/pricing`, where Startup is the second of four paid plans at $49/month |

FeedBear is the one worth reading twice. It was never a startup program or a free tier — `Startup` is the name of a paid plan, and the $49 we published is its list price.

## What changed

Each record keeps its history: `tier` becomes `Retired`, the description states what the destination serves now and what the former offer was, `eligibility` is removed, and `verifiedDate` and `source_check` are stamped 2026-09-02.

Two carried an `expires_date` that had already passed — Heroku 2026-04-30, IBM 2026-07-31 — and both were still in the ranked set, so expiry alone was not removing anything.

## Derived effects

The eight leave the ranked and best-of sets via the `offer_retired` gate, which `gateFor` evaluates first. Categories losing a record: Error Tracking (1), Startup Perks (2), Cloud IaaS (2), Communication (3).

## Checks

Data only — no `src/`, no `test/`. I grepped `test/` for all eight vendor names first: zero fixture references. `npm run validate` clean, 1580 offers.

The 17 remaining placeholder records are a separate PR. Four of them — Knowlarity (502), Autodesk (403), MATLAB (403), Create@Alibaba (JS shell) — I cannot read from here and are staying as they are.


robhunterclaude:
**Red, and I believe none of the four failures is a data error.** Naming them rather than working around them, per the usual split — I do not touch `test/` or `src/`.

**Three are one bug in `category-gate-lede.test.ts`.** `disclosedCountIn` line 145 looks for `None of them are ${RANKED_LIST_PHRASE}`, which expands to *"None of them are **not** on our ranked list"*. `expectedLede` line 51 renders *"None of them are on our ranked list"*. The double negative means the branch has never matched; it falls through to the regex at 147 and returns 0.

It could not fail before this PR. `/category/startup-perks` was 93 of 93 `eligibility_restricted`, which takes the `ALL_GATED_ELIGIBILITY_LEDE` branch at line 144 and returns `total`. Retiring two of its records makes the category fully gated with *mixed* codes for the first time, which is the only input that reaches line 145. The page itself is correct:

> 93 verified free tiers and developer deals. None of them are on our ranked list — 91 require an application or qualification, 2 have ended.

91 + 2 = 93. The renderer is right and the helper's string is wrong. `leaves the entirely gated pages on the wording they already publish` and `qualifies the count claim rather than closing it` are the same category losing its all-eligibility shape.

**The fourth is a real defect this PR exposed, not caused.** `feedbear: badge says caution, gated verdict says "This offer has ended — we keep the page for the record and no longer rate it."` Two stability judgements on one page disagree.

FeedBear is the first retired record ever to carry change history. `deal_changes.json` holds 2 records for it and **0 for each of Oaysus, QRME.SH, Supportivekoala and GitHub Models**, so the badge branch has had no subject since `Retired` was introduced. The badge re-derives stability from change counts without reading the retirement gate — the same shape as https://github.com/robhunter/agentdeals/issues/1241, where I have written it up with the counts.

I would rather not soften the data to get this green. Retiring FeedBear is right on the merits: `/early-stage` redirects to `/pricing`, where `Startup` is the second of four paid plans at $49/month.

robhunterclaude:
Your check on this PR is stale, not failing. The run reporting red is `808953d8` from 06:08Z, which predates https://github.com/robhunter/agentdeals/pull/1254 and https://github.com/robhunter/agentdeals/pull/1255 landing at ~08:15Z. Re-running the workflow does not help — it replays the same merge SHA — so I have pushed `main` into this branch to get a fresh one.

Measured locally before doing that, with this branch's `data/index.json` on top of current `main`, over the four test files that were red:

```
test/category-gate-lede.test.ts
test/eligibility-disclosure.test.ts
test/vendor-verdict.test.ts
test/gated-vendor-answers.test.ts
→ 103 pass, 0 fail
```

All four of the failures you were tracking are gone. Nothing in this PR needed to change.

One thing to watch when the fresh run lands: https://github.com/robhunter/agentdeals/pull/1257 is open and touches the same pages. I checked this branch against it explicitly — with #1257's `src/` and this branch's data, the same four files are **109 pass, 0 fail**. The two do not collide in either merge order.

That PR also caught something that only appears on your data: folding `offer_retired` into the list that decides whether a vendor page asks *"What is X's free tier?"* would have deleted that question from 12 pages once this branch merges, 8 of them yours. I split the predicate rather than take that, so the retired pages keep the question. Numbers are on #1257.

PR #1244, opened 2026-09-02
Repoint seven records whose published URL redirects off our domain, and retire one

robhunterclaude:
Seven records whose published URL redirects to a different company, repointed to the destination and rewritten from it, plus one retirement.

| vendor | was | now |
|---|---|---|
| Revolt.chat | `revolt.chat` | `stoat.chat` |
| RunMyJob | `runmyjob.io` | `runjob.eu` |
| Stack Auth | `stack-auth.com` | `hexclave.com/pricing` |
| WaiverStevie.com | `waiverstevie.com` | `waivercat.com/pricing/` |
| flightcontrol.dev | `flightcontrol.dev` | `ravion.com/pricing` |
| sslip.io | `sslip.io` | `nip.io` |
| LambdaTest | `lambdatest.com/pricing` | `testmuai.com/pricing/` |
| The GoSquared Early Stage Plan | — | retired; `gosquared.com/early-stage` serves ecosend.io |

Data only. Part of https://github.com/robhunter/agentdeals/issues/1110.

**Two corrections I made to my own work before pushing.**

*LambdaTest.* I had written that TestMu AI "publishes no free plan or quota: the pricing page lists no $0 tier". That is false. `testmuai.com/pricing/` is tabbed by product — a single fetch reads the KaneAI tab, and the Test Manager tab carries a `$0` "Free Forever" plan. The description now names the tab structure, the free plan's contents, and both paid anchors. Same failure shape as https://github.com/robhunter/agentdeals/issues/1258: one fetch reads one view of the page.

*WaiverCat.* Its free plan states no signature allowance, which holds — Basic at $15/mo names 250 and Premium at $25/mo names 500. Both quantities added, along with the page's own framing of the free account as a place to set up an environment and run tests.

**One record held out.** Ninja Outreach belongs in this batch — its source is a third-party roundup on `userlike.com` that never named it — but retiring it turns `test/source-can-show-a-later-price.test.ts` red for a real reason: `/q2-pricing-preview-2026` still links readers to that URL, from eight change records rather than from this one. The test is right. It is https://github.com/robhunter/agentdeals/issues/1259 and the retirement should land in the PR that fixes it.

PR #1257, opened 2026-09-02
Read the gate before answering whether a retired offer suits production

robhunterclaude:
Two surfaces on the vendor page still decided availability without asking `gateFor` for it, in the same shape as the six already fixed on this issue.

## 1. The production answer

`Is X's free tier good for production?` chose its opening from a list of two gate codes, `not_a_free_offer` and `offer_expired`. `offer_retired` was not on it, so the four retired pages fell through to a branch written for something else. Read off a build of `main`:

| page | before | after |
|---|---|---|
| `/vendor/oaysus` | `Oaysus does not offer a free tier for production use. Consider free alternatives in Cloud Hosting.` | `Oaysus's offer is recorded as Retired. There is no free tier here to run in production.` |
| `/vendor/github-models` | `The page we cite for GitHub Models states no terms we can read. We cannot confirm what this offer provides today…` | `GitHub Models's offer is recorded as Retired. There is no free tier here to run in production.` |
| `/vendor/supportivekoala`, `/vendor/qrme-sh` | the same fall-through | the retirement |

Neither old answer is false. Both give the wrong reason, and both contradict the three other answers on their own page, which have opened with the retirement since #1232. The `/vendor/oaysus` sentence is the branch for a vendor that never had a free tier, published over a record that had one until the domain was sold.

**Adding `offer_retired` to the existing list would have cost something, so I split the list instead.** That list had a second consumer: `gateStatesThereIsNoFreeTier`, the #1246 predicate deciding whether the page asks `What is X's free tier?` at all. Folding retired into it deletes that question from 4 live pages, and from 12 once #1252 lands — measured on #1252's data, 130 gated pages asking it drops to 125. So the two decisions now read two lists. `GATES_LEAVING_NO_FREE_TIER` is unchanged and still decides which questions get asked; `GATES_LEAVING_NOTHING_TO_RUN_IN_PRODUCTION` extends it with `offer_retired` and decides only the production answer. A retired page keeps the question, because answering what the offer used to be is the reason #1235 kept the page up at all.

## 2. The restriction a stronger gate outranks

`eligibilityGate` reads `offer.eligibility` and nothing else. Two surfaces composed from it directly — the vendor page's meta description and every category row — so a record carrying an eligibility block published its restriction even where `gateFor` returns a code that outranks it.

`gateFor` puts `offer_retired` ahead of `eligibility_restricted`. A retired startup program therefore states *"Restricted to startup applicants — not generally available"* on its own category row and in its own page description, while the FAQ on the same page says the offer is recorded as Retired. The restriction describes qualifying for something that no longer exists.

Both surfaces now read `eligibilityGateAsPublished`, which returns the eligibility gate only when it is the gate `gateFor` returns.

**This changes no page on `main`.** Census over three datasets, every record, no sample:

| dataset | records carrying an eligibility block | whose effective gate outranks it |
|---|---|---|
| `main` | 137 | **0** |
| #1252 | 130 | **0** |
| #1244 | 138 | **2** — The GoSquared Early Stage Plan, Ninja Outreach |

A branch for a shape the corpus does not hold is untested until the data arrives, which is what #1190 is about, so it is covered by synthetic records driven through the composer rather than by the corpus: an eligibility record with tier `Retired` returns `null`; an eligibility record whose expiry has passed still returns the restriction, because `gateFor` puts eligibility ahead of expiry; an ordinary eligibility record returns exactly what `gateFor` returns.

## What this does to the two open data PRs

The four affected test files run against each branch's `data/index.json` on top of this branch's `src/`:

| dataset | before | after |
|---|---|---|
| `main` | 103 pass, 0 fail | 109 pass, 0 fail |
| #1252 `pm/retire-dead-startup-programs` | 103 pass, 0 fail | 109 pass, 0 fail |
| #1244 `pm/redirect-queue-2026-09-01` | 102 pass, **7 fail** | 109 pass, **0 fail** |

Of the seven on #1244, two were the `src` surfaces above and four were test subjects derived from `o.eligibility` rather than from `gateFor`; those four are narrowed here, in the same file, on the same predicate the page uses. The seventh was the production answer.

## Verification

- 4 mutations on the two gate lists, 4 killed. Dropping `offer_retired` is killed by exactly the two new production assertions and nothing else, so nothing else was covering it. Adding all five codes is killed by four pre-existing assertions, which is what establishes `offer_retired` as the only correct addition.
- Full suite green on `main` data: 3,083 tests, 0 fail.
- Path sweep, branch build against a `main` build: **2,577 paths, 2,573 byte-identical, 4 moved, 0 status changes, 0 appearing or disappearing.** The 4 are exactly the `offer_retired` pages, and each differs by exactly two lines — the production answer, once in the `FAQPage` JSON-LD and once in the rendered `.faq-a`. Nothing else on any page.

Refs #1241
Refs #1190

### DIFF
--- a/src/eligibility.ts
+++ b/src/eligibility.ts
@@ -1,4 +1,4 @@
-import { eligibilityGateFor } from "./ranking.js";
+import { eligibilityGateFor, gateFor } from "./ranking.js";
 import type { Gate, GateCode } from "./ranking.js";
 import { gateClauseList, gateDisclosureSentence } from "./gate-disclosure.js";
 import type { Offer } from "./types.js";
@@ -7,6 +7,11 @@ export const CONDITION_RECORDING_AN_UNREAD_PROGRAM = "Startup program — check 
 
 export function eligibilityGate(offer: Pick<Offer, "eligibility">): Gate | null {
   return eligibilityGateFor(offer);
+}
+
+export function eligibilityGateAsPublished(offer: Offer, date: string): Gate | null {
+  const gate = gateFor(offer, date);
+  return gate && gate.code === "eligibility_restricted" ? gate : null;
 }
 
 export function gatedCodes(gates: (Gate | null)[]): GateCode[] {
--- a/src/serve.ts
+++ b/src/serve.ts
@@ -44,7 +44,7 @@ import { ASSISTANTS_API_SHUTDOWN } from "./assistants-shutdown.js";
 import { discontinuedOnOrBefore, PRODUCT_DEPRECATED } from "./product-deprecation.js";
 import { rankOffers, rankForListing, rotateListing, utcDate, gateFor, CRITERIA_PATH, DEMOTE_ONLY_POLICY, DISCLOSURE_RATIONALE, TIE_BREAK_ALGORITHM, GATE_TABLE, DEMERIT_TABLE, NOT_FREE_TIER_RULES, TIME_LIMITED_TIER_RULES, type TieBreak, type Gate } from "./ranking.js";
 import type { RankedEntry, RankingResult } from "./ranking.js";
-import { eligibilityGate, gatedShareDescriptionClause, gatedShareLede, publishableEligibilityConditions } from "./eligibility.js";
+import { eligibilityGateAsPublished, gatedShareDescriptionClause, gatedShareLede, publishableEligibilityConditions } from "./eligibility.js";
 import { gateDisclosureFor } from "./gate-disclosure.js";
 import { verificationLedger, QUARANTINE_AFTER_FAILURES } from "./verification-state.js";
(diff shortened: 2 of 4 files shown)


robhunterclaude:
Merged and live. Read back from production, all four subjects and five controls:

| page | production answer |
|---|---|
| `/vendor/oaysus` | `Oaysus's offer is recorded as Retired. There is no free tier here to run in production.` |
| `/vendor/github-models` | `GitHub Models's offer is recorded as Retired. …` |
| `/vendor/qrme-sh` | `QRME.SH's offer is recorded as Retired. …` |
| `/vendor/supportivekoala` | `Supportivekoala's offer is recorded as Retired. …` |

4 of 4. All four still ask `What is X's free tier?`, which is what the split predicate was for.

Controls unchanged on production: `/vendor/stripe` still opens with `Tier "Pay-as-you-go" is usage-billed from the first request.`, `/vendor/digitalocean` with `Offer expired on 2026-06-30.`, and `/vendor/supabase`, `/vendor/neon` and `/vendor/vercel` still rate the tier for production.

https://github.com/robhunter/agentdeals/pull/1252 is green as of this merge.

Issue #1067, opened 2026-08-26
/ai-coding-pricing-2026 contradicts its own change table — we invent a $10/mo Cursor tier, publish a removed free tier, and got a HIGH-impact adverse fact about GitHub backwards


robhunterclaude:
Third consecutive tier-A page review, third defect of the same class. `/ai-coding-pricing-2026` — published 2026-03-27, never reviewed, 122 days past its 30-day SLA, 4 asserted vendors with records that changed since.

Every fact below was checked against the vendor's own pricing page today, not against our record.

## The four defects

**1. Augment Code — we publish a free tier the vendor removed, and a price that no longer exists.**

Our page, "What You Actually Get for Free": *"Augment Code — Free tier for individual developers with limited usage. Professional plan uses credit-based consumption ($20–60/month)."* Comparison table row: `Augment Code | Limited usage | $20–60/mo`.

Our own change table, four screens below on the same page: *"Apr 12, 2026 — Free tier removed. Lowest plan is now Indie at $20/month with 40K credits."*

augmentcode.com/pricing today shows two plans: **Business $100/month flat** and **Enterprise, custom**. No free tier, no $20–60 individual plan, no Professional plan, and not even the Indie tier our change row recorded. The vendor moved twice; we published neither move.

A reader who follows our free-tier recommendation meets a $100/month minimum. This is a free-tier index publishing a free tier that our own records say was removed 136 days ago.

**2. Cursor — we invent a $10/mo tier that does not exist.**

Our page: *"6 plans: Free (2,000 completions/month), Hobby ($10/mo), Pro ($20/mo), Pro+ ($60/mo), Business ($40/seat), Ultra ($200/mo)."*

cursor.com/pricing today: **Hobby is the free plan** — "Hobby / Free / No credit card required / Limited Agent requests". Individual is $20/mo (Pro / Pro+ / Ultra), Teams $40/user/mo.

We priced the free plan at $10/mo. On a pricing comparison page, that is a fabricated price point.

**3. Cursor's free tier is described with GitHub Copilot's number.**

We publish Cursor Free as "2,000 completions/month". Cursor's own page says "Limited Agent requests" and names no completion figure. "2,000 completions/mo" is the GitHub Copilot free-tier figure — published verbatim in the Copilot row of the same table.

**4. We published a HIGH-impact adverse fact about GitHub and got the affected plans backwards.**

[...]

PR #1275, opened 2026-09-02
Correct four AI Coding records against each vendor's live page (#1067)

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1067 — the data half of AC-1 defects 2, 3 and 4. Data only; no `src/` or `test/`.

Every figure below was read from the vendor's own page today, 2026-09-02.

## Cursor — we published a price point that does not exist

Our description stated a **6-plan structure with Hobby at $10/month**. cursor.com/pricing lists **Hobby as the free plan** — "Hobby / Free / No credit card required / Limited Agent requests / Access to Composer". There is no $10 tier. Individual is $20/mo (Pro), Teams $40/user/mo, Enterprise custom.

The same string gave Cursor's free tier as "2,000 completions/month, 50 slow premium requests". That is **GitHub Copilot's** figure — docs.github.com states "Limited to 2000 completions per month on Copilot Free". Cursor's page names no completion or request figure at all, so the description now says so rather than substituting another number.

A `record_corrected` row retracts the $10 tier. `CHANGE_DIRECTION` marks that type neutral and `isACorrectionToOurOwnRecord` excludes it from vendor-attributed changes, so it satisfies AC-5 — it is not a new adverse event and does not move any rating.

## Augment Code — the record described plans that no longer exist

We stated "Indie plan $20/month (40K credits). Standard $60/dev/month (130K credits)". augmentcode.com/pricing today: **Standard $20/month flat**, **Business $100/month flat**, each up to 50 seats with $20 and $100 of included usage, plus custom Enterprise. No Indie plan, and Standard is not $60/dev.

The 2026-08-28 detector row says "the lowest tier is now a paid 'Business' plan at $100/month". A $20 Standard plan is lower than that today. Its `current_state` now carries the live lineup; I cannot tell from here whether Augment added Standard after 2026-08-28 or the detector missed it, so the row does not claim either.

## GitHub Copilot — the correction runs the other way from what the issue says

**The issue's defect 4 is wrong and I have said so on the issue.** #1067 claims we "got the affected plans backwards". We did not. GitHub's changelog of 2026-04-20 — the URL the record already cited — says exactly what our record says: *"To prioritize service quality for existing paying customers, we will be pausing new signups for our Student, Pro, and Pro+ plans. Copilot Free remains open."*

What is actually wrong is that **the pause has since been lifted and we never recorded it**. That changelog post is marked **Retired**, and docs.github.com/en/copilot/get-started/plans today offers "Subscribe to Copilot Pro", "Subscribe to Copilot Pro+", "Upgrade to Copilot Max" and Copilot Student. We have told readers for 135 days that Copilot Pro is closed to new signups while it is open.

So neither April row is rewritten — both summaries were accurate when written. Each records in `current_state` that the restriction has been reversed, verified today. The pause that **is** still in force is a different one: 2026-04-22, on new self-serve purchase of Copilot Business and Copilot Enterprise, which GitHub says reopens soon for card and PayPal customers.

The description also still described billing in premium requests with $0.04 overage. GitHub now bills in **AI credits at $0.01/credit**, with completions and next edit suggestions unlimited on paid plans and not billed in credits. Current table: Free and Student free, Pro $10, Pro+ $39, **Max $100** (absent from our record entirely), Business $19/granted seat, Enterprise $39/granted seat.

## Windsurf — not touched

windsurf.com/pricing returns **HTTP 429** to me, as it did to our own source check on 2026-08-28 (`outcome: unreadable`). I will not restate its tiers from a page nobody can currently read. AC-2 is about making the page's own rows internally consistent, which does not need a fresh vendor read.

## What is still open on #1067

AC-1 defects 2–4 and AC-2 are page copy in `src/serve.ts`, not data — the fabricated $10 tier is at `serve.ts:26801` and `serve.ts:27194`. The replacement text and every affected line number are on the issue.

## Verification

- `npm run validate` — 1,580 offers, 493 deal changes, all valid.
- Both files round-trip byte-identical through `json.dumps(indent=2, ensure_ascii=False)` plus a trailing newline; asserted before and after writing.
- No `test/` file asserts any of the changed strings.

### DIFF
--- a/data/deal_changes.json
+++ b/data/deal_changes.json
@@ -4506,9 +4506,9 @@
       "vendor": "GitHub Copilot",
       "change_type": "restriction",
       "date": "2026-04-10",
-      "summary": "GitHub paused new Copilot Pro trials due to abuse. Existing trials continue through their term. New users could still sign up for Pro directly (paid) until the broader April 20 signup pause closed that path too.",
+      "summary": "GitHub paused new Copilot Pro trials due to abuse. Existing trials continued through their term. New users could still subscribe to Pro directly (paid) until the broader April 20 signup pause closed that path too.",
       "previous_state": "Copilot Pro offered a 30-day free trial to any new user",
-      "current_state": "Copilot Pro trials paused indefinitely; direct Pro signup also paused as of 2026-04-20",
+      "current_state": "Resolved. Copilot Pro is open to new subscribers today (verified 2026-09-02 against GitHub's plans documentation); the trial pause is no longer reflected in GitHub's current plan pages.",
       "impact": "medium",
       "source_url": "https://github.blog/changelog/2026-04-10-pausing-new-github-copilot-pro-trials/",
       "category": "AI Coding",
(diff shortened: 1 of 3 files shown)


robhunterclaude:
## Red, and I am leaving it red. One assertion, and the data is right.

`test/restriction-evidence.test.ts:351`

```
assert.strictEqual(stored.filter(c => c.change_type === "record_corrected").length, 2);
```

3,210 of 3,211 pass. The one failure is `3 !== 2`: this PR adds the Cursor retraction, so there are now three `record_corrected` rows.

**The assertion pins the size of a category that is supposed to grow.** Its neighbours in that block test properties — no stored restriction is refused by the rule, no restriction record corrects our own entry, the population is large enough not to be vacuous. This line is the only one that pins a count, and the count it pins is of the mechanism we use to admit our own errors. It fires on the first correction anyone publishes after it was written.

I am not touching it: my PRs are data only, and dropping the retraction row to go green would be removing a true record because a test forbids it. The intent of the line looks like "the two existing corrections were not retyped as restrictions", which `>= 2` preserves. Your call on the right form.

For the record, this pin is four days old — added 2026-08-29 in `ed29fd5` (#1146) — so it does not explain why we have only two corrections in 492 records since March. I checked before assuming it did.

Extended `#1190`'s inventory with this, since the harm is identical: a test that goes red while the code and the data are both fine, blocking every open PR while it lasts. `#1190` is about the clock; this one is about correct data being added. Same consequence.

The three description corrections in this PR — Cursor's non-existent $10/mo tier, Augment Code's replaced plans, and Copilot's lifted pause — do not depend on the retraction row and are the reader-facing part.

Issue #1267, opened 2026-09-02
A field named `recent_change` is dated 35 days in the future — AWS publishes `free_tier_longevity_days: 0` and a `dateModified` of 2026-10-07


robhunterclaude:
`/api/offers` publishes this for AWS today, 2026-09-02:

```json
"recent_change": "2026-10-07: AWS Proton end of support. Infrastructure-as-code
provisioning service being retired. All environments, services, and templates
will stop working."
```

A field named `recent_change` carrying a date 35 days in the future. `/api/vendor-risk/aws` returns `free_tier_longevity_days: 0` and the summary sentence:

> AWS has a stable pricing history. Free tier verified for 0 days.

And `/vendor/aws` publishes `"dateModified":"2026-10-07"` in its `WebPage` JSON-LD, while `sitemap-vendors.xml` gives the same URL `<lastmod>2026-08-22</lastmod>`. Two freshness claims about one page, 46 days apart, one of them in the future.

## This is #1171 again, on the surfaces it did not reach

#1171 was Rob's report that the homepage led its "Recent pricing changes" list with three events that had not happened. #1173 fixed the homepage. **The same three records are still driving eight other surfaces.** Fetched from production today:

| surface | occurrences of `2026-10-07` / `2026-09-30` | what it publishes |
|---|---|---|
| `/vendor/aws` | 17 | `dateModified`, the verdict sentence, two FAQ answers, the change table |
| `/pricing-changes` | 14 | `dateModified`, `temporalCoverage` end |
| `/alternative-to/aws` | 4 | |
| `/compare/aws-vs-google-cloud` | 2 | |
| `/compare/aws-vs-azure` | 2 | |
| `/best/free-cloud-iaas` | 1 | |
| `/api/offers` | 1 | `recent_change` |
| `/api/vendor-risk/aws` | — | `free_tier_longevity_days: 0` |

`/vendor/deepgram` publishes `"dateModified":"2026-09-12"` on the third record.

## Why the fix did not generalise

`latestEventDate(changes, notAfter?)` in `src/change-dates.ts:134` takes the guard as an **optional** parameter. It has two callers:

```ts
// src/serve.ts:80   — the "Latest tracked change" note
const latest = latestEventDate(changes, utcToday());

// src/serve.ts:4062 — dateModified for every vendor page
const lastPricingChange = latestEventDate(vendorChanges);
```

[...]

PR #1314, opened 2026-09-03
Date the Deepgram change when we found it, not when its promo expires

robhunterclaude:
Data-only. One record, two fields.

The Deepgram change record stored `date: "2026-09-12"` with `date_source: "vendor_page"`. That date is the expiry of the Flux TTS promotion the summary describes — *"Flux TTS is free until September 12, 2026"* — not the date anything changed. It was nine days in the future, and `vendor_page` is one of the two event-dated sources, so the site published it as an established fact.

**What `/vendor/deepgram` published.** Run against the site's own helpers in `src/change-dates.ts`, before and after:

| derived value | before | after |
|---|---|---|
| `dateModified` and "Last updated" | `2026-09-12` | `2026-07-30` |
| `priceValidUntil` on the free-credits `Offer` | `2026-09-12` | *(absent)* |
| schema.org `Event.startDate` | `2026-09-12` | *(absent)* |
| `datePublished` | `2026-09-12` | *(absent)* |
| change-log label | `2026-09-12` | `discovered 2026-08-28` |

`2026-07-30` is Deepgram's actual `verifiedDate`; `lastUpdated` falls back to it once no event-dated change outranks it.

**The change is `date` → `2026-08-28` and `date_source` → `discovered`.** That is what we know: `reverify-ai` found the page had changed on 2026-08-28. We do not know when Deepgram changed it. `discovered` is the source that says so, and it correctly suppresses `startDate`, `datePublished` and the expiry derivation.

**Nothing is lost.** The September 12 expiry is still stated in `summary` and `current_state`, where it is prose about the promotion rather than a date field claiming the change happened then.

**One thing I did not fix, reported on https://github.com/robhunter/agentdeals/issues/1267.** `priceValidUntil` came from `offerExpiryAfter`, which accepts any future-dated change where `endsTheListedOffer` is true — and that helper returns `true` for every `change_type` except `product_deprecated`. This record is a `new_free_tier`. The site classified a new free tier as an event that ends the listed offer. It happened to produce a defensible date here by coincidence; the logic is inverted for the positive change types. That is `src/`, so it is not in this PR.

`npm run validate` passes: 1,580 offers, 524 deal changes. `deal_changes.json` round-trips byte-identical through `json.dumps(indent=2, ensure_ascii=False)` before and after the edit.

### DIFF
--- a/data/deal_changes.json
+++ b/data/deal_changes.json
@@ -6363,8 +6363,8 @@
     {
       "vendor": "Deepgram",
       "change_type": "new_free_tier",
-      "date": "2026-09-12",
-      "date_source": "vendor_page",
+      "date": "2026-08-28",
+      "date_source": "discovered",
       "summary": "The $200 free credit is still offered, but the details have expanded. Text-to-Speech's Flux TTS is free until September 12, 2026, and Voice Agent API's Flux TTS is also free until September 12, 2026.",
       "previous_state": "Speech-to-text and text-to-speech API — $200 free credits on signup (no credit card required), ~43K minutes transcription with Nova model, credits never expire. Access to all model endpoints",
       "current_state": "Free $200 Credit then pay-as-you-go. Flux TTS is free until 9/12/2026. Voice Agent API's Flux TTS is free through September 12, 2026.",


robhunterclaude:
Rebased onto `28a97ec`. `npm run validate` green — 1,580 offers, 524 changes.

**The `Tests` failure on this branch is inherited, not caused.** The failing set here is identical to `main` HEAD's, name for name:

- `quotes the page's own words in every one of their details`
- `a vendor whose only history is limits_increased renders stable, live`
- `a filter that empties a covered query is not recorded as a catalog gap`
- `applies the same rule to the MCP search_deals tool`
- `holds a cohort the baseline matches exactly`

`main` has been red on those five since 2026-09-03T11:37Z. The cause and the sequence are in #1317. This branch touches one field of one record in `data/deal_changes.json` and none of the five reads it.

Merging on that basis: the defect it fixes is live and reader-facing — `/vendor/deepgram` publishes a change dated nine days in the future because a promo expiry was stored as the change date, and four derived dates render from it. Holding a data-only fix behind a red that no data change can clear leaves the wrong dates up for as long as the blockage lasts.

Issue #1321, opened 2026-09-04
A page-review budget can stop the catalogue from updating — the data gate blocks on a ratchet whose constant lives in src/ and whose numerator is at its ceiling


robhunterclaude:
#1317's gate is right and should stay. This is about what it now sits on top of.

`gate-data-push.sh` runs the **whole** suite and refuses the push if any test is red. One test in that suite measures my editorial review debt, its budget is a constant in `src/`, and the automated job cannot change a constant in `src/`. So a page-review backlog can stop the catalogue from updating — and it will, in about two days.

## The exact-equality rule

`staleFactViolations` (`src/page-reviews.ts`) fails unless the cohort **equals** `STALE_FACT_PAGES_BASELINE`:

```ts
if (stale.length === budget) return [];
```

Above the budget it says "N more than the budget allows". Below it says "lower `STALE_FACT_PAGES_BASELINE` to N so the slot cannot be reused". Both are failures. As a ratchet on a human PR that is a good design — you improve the number and lower the baseline in the same commit. As a gate on an automated data push it is a trap, because the job has no way to lower the baseline.

## The cohort is at its ceiling

Measured over the shipped registry today: 67 pages, of which **59 name at least one vendor and 8 name none**. Every one of the 59 is in the cohort. The 8 that are not — `/agent-stack`, `/budget-builder`, `/compare-tool`, `/disclosure`, `/marketplace`, `/shutdowns`, `/stack-check`, `/tenor-alternatives` — hold zero vendor names, so they can never enter it.

The number therefore has one input: pages I review. It rose 42 → 51 → 53 → 57 → 59 between 2026-08-14 and 2026-09-03 as the rolling re-verification walked the catalogue for the first time, and it stopped at 59 because it ran out of pages to stale.

## After #1320 the gate has about two days of headroom

https://github.com/robhunter/agentdeals/pull/1320 reviews two pages and returns the cohort to exactly 57, so the suite goes green. What can push it back to 58 is now a very small set: the vendors named on the two pages I just cleared.

| quantity | measured |
|---|---|
| vendor slugs that would raise the cohort if a change is recorded | 28 |
| catalogue offers carrying one of those vendors | 28 of 1,580 (1.77%) |
| changes recorded per day by the rolling job | 28, 28, 31 on Sep 1–3 (mean 29) |
| expected time to the next trip | **~1.9 days** |

[...]

PR #1320, opened 2026-09-04
Review two pages against their vendors and return the stale-fact cohort to its budget (#1317)

robhunterclaude:
Refs #1317, AC-4.

`main`'s last failing test is the stale-fact ratchet: 59 pages against a baseline of 57. This reviews two of the five one-fact pages against their vendors and returns the cohort to 57. Data only — `data/page-reviews.json`, two records.

## `/supabase-vs-firebase` — pass

Every quantitative claim checked against `supabase.com/pricing` and `firebase.google.com/pricing` on 2026-09-04.

Supabase, all correct: 500 MB database, 1 GB file storage, 5 GB egress, 50,000 MAU, unlimited API requests, 200 concurrent realtime connections, 2 active projects, projects paused after 1 week of inactivity. Firebase, all correct: Firestore 1 GiB stored, 50K reads/day, 20K writes/day.

The claim I held this stamp for last cycle is now confirmed at the vendor. The page's headline caveat says Firebase removed Cloud Storage from the free Spark plan. On `firebase.google.com/pricing` every Cloud Storage row in the Spark column reads `Not applicable` — `*.appspot.com` legacy buckets and `*.firebasestorage.app` alike, across GB stored, GB downloaded, upload operations and download operations. File storage does require Blaze.

One residual, recorded in the note rather than left silent: the **February 2026** date the page gives for that removal is not stated on the vendor page. It rests on our own change record.

The 2026-09-03 Supabase change that staled this page is already reflected in the prose — "2 free projects that pause after 1 week of inactivity" — so nothing on the page needed correcting.

## `/email-comparison-2026` — fail, with the correction named

Reviewed, wrong, and stamped `fail` so the page keeps publishing "corrections outstanding" to the reader. The clock advances because the page has now been read; the note carries what a fix has to touch.

The MailerSend row is wrong by 6x, and the page already contradicts itself about it:

| surface on the page | says |
|---|---|
| main comparison table | MailerSend, 3,000/mo, daily cap "None", stability "Stable" |
| prose block | "3,000 emails/month … No daily cap (unlike Resend)" |
| cost-at-scale table | 3K emails, $0.30 |
| Recent Deal Changes | "MailerSend.com — the free tier now offers 500 emails/month … 14-day professional trial", HIGH, Aug 28 2026 |
| pricing-change timeline | "MailerSend — free tier reduced from 3,000 to 500 emails/month" |

[...]

Issue #1317, opened 2026-09-03
Scheduled data commits are the only writes to main the test suite never runs — main has been red for ten hours and seven merges


robhunterclaude:
`main` has failed `Tests` on every run since **2026-09-03T11:37Z** — ten hours and seven merges. Every open pull request inherits the same five failures. My own #1314 is data-only and its failing set is byte-identical to `main` HEAD's, so the check can no longer tell a regression from an inheritance.

## The mechanism

`tests.yml` runs on `pull_request` and on `push` to `main`. Three scheduled workflows write to `main` without a pull request — `reverify.yml`, `liveness.yml`, `analytics-rollup.yml` — each ending in `git push origin HEAD:main` with `GITHUB_TOKEN`. GitHub does not trigger workflows from pushes made with that token.

Measured against the check-runs API, the last six `data(auto)` commits on `main`:

| commit | check runs |
|---|---|
| `e34b09c` rolling re-verification | none |
| `d5fd6b6` link liveness | `reverify` — the writer, not the suite |
| `56f8615` analytics rollup | none |
| `1c3e3c7` rolling re-verification | none |
| `b7951dc` link liveness | none |
| `295e268` analytics rollup | none |

The commits that rewrite the corpus the suite asserts over are the only commits on `main` the suite never sees. The red then surfaces on the next human pull request, which wears a failure it did not cause.

Today's sequence: `a69655c` green at 09:42Z → link liveness and rolling re-verification land at 10:0xZ and 10:4xZ with no test run → #1300's merge push at 11:37Z red, and red on every run since.

### One is a real regression the ratchet caught

`test/stale-page-facts.test.ts:152` — the cohort is **59 pages**, `STALE_FACT_PAGES_BASELINE` is **57**. This is the guard working, and it is being carried as noise.

I recomputed the cohort at each date against the shipped register:

| as of | pages in cohort | stale facts |
|---|---|---|
| 2026-08-27 | 36 | 75 |
| 2026-08-28 | 51 | 132 |
| 2026-09-01 | 53 | 142 |
| 2026-09-02 | 57 | 220 |
| 2026-09-03 | **59** | **286** |

The register holds 67 pages and 9 carry a `reviewed_at`. Eight pages remain outside the cohort.

[...]

PR #1325, opened 2026-09-04
Move the quality budgets into data and split what a red suite holds back (#1321)

robhunterclaude:
Refs #1321

The gate from #1317 is right and stays. This changes what it sits on top of, following the ruling on the issue: **the budget moves into data, the gate blocks on wrong and reports on unreviewed, and the alarm reads `main`.**

## The budget is data now

`data/quality_budgets.json` holds all five ratchet numbers — `stale_fact_pages: 57`, `unsourced_tier_a: 43` and the three FAQ counts. `src/page-reviews.ts` and `src/faq-provenance.ts` read them; the literals are gone. The failure message changed with them — it used to say "lower `STALE_FACT_PAGES_BASELINE`", a constant in `src/`, and now says "set `stale_fact_pages` to 56 in `data/quality_budgets.json` — run `npm run ratchet:budgets`".

That is the whole of the problem this issue opened on. Both actors who can lower the number — the scheduled jobs and data-only pull requests — can write `data/`. Neither can write `src/`. #1323 and #1324 each corrected facts on `/hetzner-pricing-2026` and neither could record the review; either could now.

`npm run ratchet:budgets` measures both cohorts and lowers any budget above its measurement. It never raises one: a budget below its measurement is reported and left, and the suite is what fails until a person decides. The daily re-verification runs it before the gate and commits the result with its data, so an improvement the job earns is banked in the same commit.

**Named rather than hidden:** the job lowering the budget automatically means a transient data fault that shrinks the cohort ratchets the budget down with it, and the correction that restores the data then needs a deliberate raise. That raise is a one-line data edit rather than a `src/` change, which is the trade the ruling asked for — but it is a real edge and it is worth knowing about before it happens.

## Block on wrong, report on unreviewed

`npm run test:gated` runs the same test files as `npm test` with one extra reporter, `scripts/reporters/failing-test-files.js`, which writes the file of every failure. The gate subtracts `scripts/gate-non-blocking-tests.json` from that list. Anything left holds the commit; nothing left and the data reaches `main` with the failures printed.

Two files are on the list:

- `test/stale-page-facts.test.ts` — the cohort this issue is about.
- `test/page-data-provenance.test.ts` — holds the register's declaration of where each page's numbers come from against what the page renders.

The rule for admission is in the file: everything the test fires on has to be a statement about how current our own reading is, and nothing it fires on may be a statement that the catalogue is wrong. The list lives under `scripts/`, which no gated workflow hands to the gate, so widening it takes a pull request. A test asserts that.

Three things still hold the commit whatever the list says, each with a test: a failure in a file nobody excused, a red suite that named no file at all, and a build that does not compile.

## A refusal is visible, and survives the next one

The quarantine ref is `<prefix>-<UTC timestamp>-<short sha>` and the push is not forced. Two refusals leave two refs; a genuine price correction found on day one is no longer overwritten by day two. Tested by refusing twice and reading both.

When the gate quarantines, it writes `quarantined=true` and the ref to `$GITHUB_OUTPUT`, and each of the three workflows opens or comments on a single issue carrying the marker `data-push-refused`. That is criterion 3: the freeze is visible without opening a workflow log.

**The report path needed the same treatment, for a reason that only shows up once you put the two halves together.** "Report and go red on `main` where a human sees them" does not work by itself here: a scheduled push is made with `GITHUB_TOKEN`, which cannot trigger `tests.yml` — that is the finding #1317 shipped for. So a data run that ships over an excused failure would leave `main` red with no run behind it to say so. `scripts/report-data-push-outcome.sh` therefore has two outcomes and two markers. A refusal comments every time, because each one is a distinct event holding a distinct ref. A shipped-over-failures notice opens one issue and then stays quiet, because it recurs daily until somebody clears the budget.

## The alarm reads main

`check-change-log-staleness.js` gains `--from-ref`, and `reverify.yml` passes `origin/main` after fetching it. It read `data/deal_changes.json` from the workflow's own checkout, where the run had just written it, so it reported the log fresh whether or not the commit reached `main`.

The test builds a repository whose committed log is eight months stale and whose working tree is fresh, and asserts the two readings disagree — `failJob: false` from disk, `failJob: true` from the ref. That is the defect, reproduced.

## Criterion 5 — the other ratchets

I looked for every constant in `src/` that pins a count over the corpus. There are three, holding five numbers, and **all five are in `data/quality_budgets.json` now** — the class, not the instance.

| constant | reachable by scheduled-job data | lowered by |
|---|---|---|
| `STALE_FACT_PAGES_BASELINE` | yes, through `data/deal_changes.json` | `npm run ratchet:budgets`, and the daily job |
| `UNSOURCED_TIER_A_BASELINE` | no — it reads only the register | `npm run ratchet:budgets` |
| `FAQ_BASELINE` — three counts | yes, through `data/index.json` | by hand, from the number the suite prints |

`UNSOURCED_TIER_A_BASELINE` cannot be moved by the jobs, but the test file holding it can: the other half of `pageSourceViolations` reads rendered pages, and those read `data/index.json`. That is why the file is on the non-blocking list and the constant is in the budgets file anyway.

`FAQ_BASELINE` is the interesting one, and the evidence that it is reachable is already in the suite: `test/faq-provenance.test.ts` boots a server against a perturbed catalogue and asserts that **more than five FAQ answers move with it**. `stating_a_figure` is pinned twice — once as a count of answers, once as the number of dated answers the catalogue *cannot* move — so a re-verification that rewrites an offer's description can break either. Its three numbers are in the budgets file, but measuring them needs a running server, so `ratchet:budgets` reports them as "measured only by the suite" and does not touch them. That is honest rather than clever: the file it lives in is now writable by whoever earns a lower number, which was the whole complaint.

**A second answer to criterion 5, which the ruling asked for:** any red test blocks the gate, not only a ratchet. `source-check-pages.test.ts:417` was red on 60% of days for a date-seeded tie-break, measured on #1190. Nothing here excuses a flaky test — the answer to that class is criteria 3 and 4, which turn a flake from a silent freeze into a visible, recoverable quarantine, plus fixing the flakes on #1190.

## What the cohort can actually do

Measured on the shipped register today: **0 slugs can raise the cohort today, 28 from tomorrow.** The two pages #1320 cleared were cleared today, so a change recorded today does not beat their clock; from tomorrow it does. That agrees with the ~1.9 days on the issue and puts a date on it.

## Verification

The gate's own behaviour is proved against a real repository with its own bare origin, one test per branch: an excused-only red run pushes, a data-wrong red run quarantines, a mixed run quarantines, a run that failed without naming a file quarantines, a failed build quarantines, and two refusals leave two refs carrying two different days of data.

The last local run went through `npm run test:gated` rather than `npm test` — **3,648 tests, 0 failures** — so the reporter the gate depends on is exercised on all 165 real test files and not only on the fixture. It wrote an empty failing-files list, which is the green path.

### DIFF
--- a/.github/workflows/analytics-rollup.yml
+++ b/.github/workflows/analytics-rollup.yml
@@ -12,6 +12,7 @@ on:
 
 permissions:
   contents: write
+  issues: write
 
 concurrency:
   group: analytics-rollup
@@ -49,6 +50,7 @@ jobs:
           echo "written=$WRITTEN" >> "$GITHUB_OUTPUT"
 
       - name: Run the suite, then push to main only if it is green
+        id: gate
         env:
           WRITTEN: ${{ steps.rollup.outputs.written }}
         run: |
@@ -56,3 +58,17 @@ jobs:
             data-quarantine/analytics-rollup \
             "data(auto): analytics rollup — ${WRITTEN} day(s)" \
             data/analytics
+
+      - name: Say where it can be seen what the gate did with this run's data
+        if: always() && (steps.gate.outputs.quarantined == 'true' || steps.gate.outputs.pushed_over_failures == 'true')
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          QUARANTINED: ${{ steps.gate.outputs.quarantined }}
+          QUARANTINE_REF: ${{ steps.gate.outputs.quarantine_ref }}
+          NON_BLOCKING_FILES: ${{ steps.gate.outputs.non_blocking_files }}
+        run: |
+          if [ "$QUARANTINED" = "true" ]; then
+            bash scripts/report-data-push-outcome.sh "Analytics rollup" refused "$QUARANTINE_REF"
+          else
+            bash scripts/report-data-push-outcome.sh "Analytics rollup" shipped-over-failures "$NON_BLOCKING_FILES"
+          fi
--- a/.github/workflows/liveness.yml
+++ b/.github/workflows/liveness.yml
@@ -7,6 +7,7 @@ on:
 
 permissions:
   contents: write
+  issues: write
 
 concurrency:
   group: link-liveness
@@ -40,6 +41,7 @@ jobs:
           echo "unknown=$UNKNOWN" >> "$GITHUB_OUTPUT"
 
       - name: Run the suite, then push to main only if it is green
+        id: gate
         env:
           UNREACHABLE: ${{ steps.liveness.outputs.unreachable }}
           UNKNOWN: ${{ steps.liveness.outputs.unknown }}
@@ -48,3 +50,17 @@ jobs:
             data-quarantine/link-liveness \
             "data(auto): link liveness — ${UNREACHABLE} unreachable, ${UNKNOWN} could not be checked" \
             data/link_health.json
+
+      - name: Say where it can be seen what the gate did with this run's data
+        if: always() && (steps.gate.outputs.quarantined == 'true' || steps.gate.outputs.pushed_over_failures == 'true')
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          QUARANTINED: ${{ steps.gate.outputs.quarantined }}
+          QUARANTINE_REF: ${{ steps.gate.outputs.quarantine_ref }}
+          NON_BLOCKING_FILES: ${{ steps.gate.outputs.non_blocking_files }}
+        run: |
+          if [ "$QUARANTINED" = "true" ]; then
+            bash scripts/report-data-push-outcome.sh "Link liveness" refused "$QUARANTINE_REF"
+          else
+            bash scripts/report-data-push-outcome.sh "Link liveness" shipped-over-failures "$NON_BLOCKING_FILES"
+          fi
--- a/.github/workflows/reverify.yml
+++ b/.github/workflows/reverify.yml
@@ -53,7 +53,13 @@ jobs:
           echo "retried=$RETRIED" >> "$GITHUB_OUTPUT"
           echo "repicked=$REPICKED" >> "$GITHUB_OUTPUT"
 
+      - name: Lower any quality budget this run's data has earned
+        run: |
+          npm run build
+          node scripts/ratchet-quality-budgets.js --lower
+
       - name: Run the suite, then push to main only if it is green
+        id: gate
         env:
           VERIFIED: ${{ steps.reverify.outputs.verified }}
           FLAGGED: ${{ steps.reverify.outputs.flagged }}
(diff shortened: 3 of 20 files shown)


robhunterclaude:
## Approved — merging

I checked the five criteria against the branch and re-measured the cohort myself rather than reading the numbers off the PR.

**Criterion 1 holds and the figure reproduces.** Running `staleFactPages` on this branch against the shipped register and change log: **0 vendor slugs can raise the cohort dated today, 28 from tomorrow**, carrying 28 of 1,580 offers (1.77%). Same 28 the issue named, and they are exactly the vendors on the two pages #1320 cleared — the email comparison and supabase/firebase. Held at 57 across fourteen simulated days with the change log frozen, so page aging alone does not move it; only a new change record does.

**Criteria 2, 3 and 4 read as done from the fixture tests.** `gateVerdict` fails closed on an empty failing-files list, which is the case I went looking for.

Three things for the record. None of them holds this PR.

## 1. The excuse is file-scoped and one of the two files mixes both classes

`gate-non-blocking-tests.json` states the rule as "nothing it fires on may be a statement that the catalogue we are about to publish is wrong." `test/stale-page-facts.test.ts` satisfies it — every assertion in it is about the register, the cohort, a budget or a reviewer's note, and none of it is published.

`test/page-data-provenance.test.ts` does not. Three of its assertions fire on a statement we publish to a reader:

- *names our index as a source only on the pages that read it* — fires when a page tells a reader "our verified index of…" and the render is blind to the index.
- *makes no re-check claim anywhere on a page that the register contradicts* — fires when a page tells a reader "last checked 2026-08-26" with no review behind it.
- *says corrections are outstanding wherever a review recorded a failure, and nowhere else* — fires when a page whose review failed does not say so.

A false provenance line is not a wrong catalogue record, so the rule as written admits these on a technicality. The cost: when that file is red on its `unsourced_tier_a` ratchet and on one of the three above in the same run, the gate sees one file, excuses it and pushes. Narrowing the rule to "nothing it fires on may be a statement we publish to a reader" would mean the file has to be split before it can be listed.

## 2. The FAQ budgets can still freeze the catalogue, by this PR's own table

Criterion 5's table says the three FAQ counts are reachable by scheduled-job data through `data/index.json` and are lowered by hand. `test/faq-provenance.test.ts` asserts all three with `strictEqual` (lines 247, 259, 264) and is not on the excuse list. A re-verification that moves one of them quarantines the push, and `ratchet:budgets` explicitly declines to measure them, so the job cannot clear it. That is #1321's shape, relocated.

It is materially better than it was — the refusal is visible, the data survives the next day, and clearing it is a one-line data edit rather than a `src/` change. But `faq_answers_with_a_digit_but_no_figure` is a count of numbers nobody has classified yet, and it can stop the catalogue from updating.

## 3. Red main becomes the steady state, and that follows from my ruling

I asked for "report and go red on `main` where a human sees them." Against the measurement above that means: the cohort rises within about two days, the budget does not follow it up, `stale-page-facts` goes red, the excuse ships the data, and main stays red until I review a page. The cohort refills at roughly the rate I can clear it, so red is where main sits a large share of days from here.

`tests.yml` runs the same suite on every pull request, so the cost lands on you: red stops distinguishing what you changed from the depth of my review queue. That is the condition #1317 was filed about.

I do not want that fixed here. I am filing it separately and the direction is **no test measures editorial debt**: the five counts move out of `strictEqual` into a reported check the daily job runs, which opens and updates one issue when a budget is exceeded. The budget still never rises on its own, I still get told, and green means green. It also dissolves 1 and 2 — the excuse list can go back to empty, and `faq-provenance.test.ts` keeps its hard assertions (an answer stating a figure with no provenance clause) while losing its counters.

Issue #1103, opened 2026-08-28
The detector records a change and never updates the record — every vendor page it touches now contradicts itself


robhunterclaude:
The detector shipped in #1097 ran for the first time today and recorded 19 changes (#1087). It
wrote them to `data/deal_changes.json` and **changed nothing in `data/index.json`**. So every vendor
page it touched now renders the superseded terms as current, with a record directly underneath
saying they are superseded.

## `/vendor/thunder-client`, live right now

Two blocks on one page, in this order:

> **Free Tier Details** — Lightweight REST API client for VS Code. **Free tier includes**
> collections, environments, local storage, and request history. No account required for local use.

> **Pricing Change History (1 recorded)** — model change, discovered 2026-08-28, **high impact**.
> "The pricing information has changed. The page no longer explicitly mentions a free tier."
> **Before:** Lightweight REST API client for VS Code. Free tier includes collections, environments,
> local storage, and request history...

The `Before:` text and the `Free Tier Details` text are the same string. The page publishes the
old state as fact and the notice that it is old, four inches apart.

## Scale

Six of six vendors I spot-checked from today's 19 do the same: `thunder-client`, `deno-deploy`,
`segment`, `semgrep`, `sematext`, `middleware-io`. It follows from the pipeline rather than from
those six — nothing in the run path writes the new terms to the record, so it is every change, and
it will be roughly 19 more every morning at 06:00.

Deno Deploy is the one with a checked number behind it (#1087): the page still advertises 100 GB
egress on a free tier the vendor now caps at 20GiB, above a record that says so.

## This is the cause of a family I have been filing page by page

#1064 (countdowns frozen past their deadline), #1067 (a HIGH-impact GitHub fact rendered backwards),
and cycle 65's `/ai-coding-tools-pricing` (recommends Copilot plans its own change table calls
paused) are all the same shape: **a change is recorded and the thing it describes is never
updated.** Until today that produced a slow trickle from manual entries. The detector makes it a
daily generator.

## Acceptance criteria

- **AC-1** — After a run records a change, no page renders the superseded text as current. Either
  the record is updated from `current_state`, or the page stops presenting a description the change
  log contradicts. I am not specifying which; pick the one you can make safe.
- **AC-2** — **Gate this on https://github.com/robhunter/agentdeals/issues/1107.** Two of today's 19
  records describe no change at all. Applying those to `index.json` would write a false description
  into the record itself, which is worse than the contradiction — the contradiction is at least
[...]

PR #1377, opened 2026-09-05
Retire Momento's free tier, which its own pricing page no longer sells

robhunterclaude:
Retire Momento's free tier, which its own pricing page no longer sells

gomomento.com/pricing was read on 2026-09-04 and again on 2026-09-05. Its
6,635 characters of visible text contain the string "free" once, in
"freeing up your team". Every plan on the page carries a price: Momento
Cache Flex from $13 per GB-month, Momento Cache Cluster from $0.0105 per
instance-hour, Pub/sub Topics from $1 per 1M operations, Valkey Router
$0.01 per GB of data transfer.

Our record still read tier Free with "5 GB data transfer/month free",
verified 2026-08-23, while our own free_tier_removed change record dated
2026-09-05 says the opposite and cites the same URL.

The one $0 on that page is the Starter Basic Support plan. That is a
support level, not a product allowance, and the description now says so
in case a later pass reads it as a free tier.

The source_check detail was "the page names Momento as \"momento\" and
states \"$13\"" — a pass written off a bare price string. It now quotes the
two sentences that establish the claim.

Derived-set diff, local server against production: the route set is
unchanged at 483 pages and 1,573 vendors; Momento drops out of
/best/free-databases entirely, /vendor/momento renders the retired badge
and answers "there is no free tier here to run in production", and
/category/databases lists it as Retired.

Ratchet unchanged: 57 / 43 / 95 / 614.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -3205,8 +3205,8 @@
     {
       "vendor": "Momento",
       "category": "Databases",
-      "description": "Serverless cache and pub/sub — 5 GB data transfer/month free, sub-millisecond latency, no infrastructure to manage",
-      "tier": "Free",
+      "description": "Managed Valkey cache, cluster and pub/sub, formerly a serverless cache with a free tier. No free tier. gomomento.com/pricing, read 2026-09-05, prices every plan: Momento Cache Flex from $13 per GB-month, Momento Cache Cluster from $0.0105 per instance-hour, Pub/sub Topics from $1 per 1M operations, Valkey Router $0.01 per GB of data transfer. The only $0 on the page is the Starter Basic Support plan, which is a support level and not a product allowance.",
+      "tier": "Retired",
       "url": "https://www.gomomento.com/pricing",
       "tags": [
         "database",
@@ -3216,11 +3216,11 @@
         "free tier",
         "redis-alternative"
       ],
-      "verifiedDate": "2026-08-23",
+      "verifiedDate": "2026-09-05",
       "source_check": {
         "checked": "2026-09-05",
         "outcome": "ok",
-        "detail": "the page names Momento as \"momento\" and states \"$13\""
+        "detail": "the page states \"Momento Cache Flex from $13 per GB-month\" and \"Pub / sub Topics from $1 per 1M operations\"; the string \"free\" occurs once in 6,635 characters of visible text, in \"freeing up your team\""
       },
       "product_subtypes": {
         "taxonomy": "Databases",
--- a/data/quality_budgets.json
+++ b/data/quality_budgets.json
@@ -8,8 +8,8 @@
     "faq_answers": 166,
     "faq_answers_stating_a_figure": 77,
     "faq_answers_with_a_digit_but_no_figure": 40,
-    "records_with_superseded_terms": 176,
-    "vendor_pages_withholding_superseded_terms": 175,
-    "ungated_pages_withholding_superseded_terms": 166
+    "records_with_superseded_terms": 175,
+    "vendor_pages_withholding_superseded_terms": 174,
+    "ungated_pages_withholding_superseded_terms": 165
   }
 }


robhunterclaude:
Diagnosed the red run. It is not the data and it is not a hung runner — my note that this job stalled for 90 minutes was wrong, it finished in 7 and failed.

3,865 of 3,866 tests pass. The one failure is `test/retired-records.test.ts:196`:

```
✖ keeps citing the page a change was read from, even where the offer has since ended
  AssertionError: no page cites https://www.gomomento.com/pricing as the source of the change recorded against it
```

Retiring Momento makes `gomomento.com/pricing` a retired URL, so the gate strips every anchor to it — correctly, we should not send readers to a dead free tier. But today's reverify run also wrote a `free_tier_removed` change record citing that same URL, and the test requires the citation to survive the anchor. It does not: the only function that emits `<a class="change-source">` is `buildChangeCard`, called from two hardcoded date-bounded pages that stop at the end of Q2.

The test is right and the data is right. I am not working around either. 15 records are retired, two of their URLs are the source of a change record, and the assertion has been green on a population of one — Augment Code, held up by `/q2-pricing-preview-2026`.

Filed as https://github.com/robhunter/agentdeals/issues/1380, high: 553 published changes, 458 holding a `source_url`, 272 cited on no page we serve, 247 of those from the last two months. AC-3 there is this PR going green.

**Holding this PR until #1380 ships.** Rebased onto `main` so it is current. Momento's own pricing page contains "free" once in 6,635 characters, in "freeing up your team"; the only `$0` on it is the Starter Basic Support plan, which is a support level rather than a product allowance. That verdict has not changed.

robhunterclaude:
Correction to my comment above: that was the pre-rebase run. Rebased onto `main` with #1379 in it, this PR is red on **four** assertions, not one.

- `test/retired-records.test.ts:196` — the citation gap, https://github.com/robhunter/agentdeals/issues/1380. Unchanged.
- `test/superseded-terms.test.ts:143` — `236 !== 237`
- `test/superseded-terms.test.ts` second count — `234 !== 235`
- `test/gated-vendor-answers.test.ts:279` — `221 !== 222`

The last three are hardcoded equality counts pinning the #1103 withheld population. Retiring a record empties one slot, so all three go red for the same reason https://github.com/robhunter/agentdeals/pull/1381 does. Filed as https://github.com/robhunter/agentdeals/issues/1383, high — it also fires in the other direction, on tomorrow's 06:00Z run.

Still holding this PR. It now needs #1380 **and** #1383.

robhunterclaude:
Rebased onto `b9bbf4c`. The red run was against a `test/superseded-terms.test.ts` that has since changed twice — it asserted 237 and 235, and the budgets now read 176/175/166.

Added one commit lowering the three superseded budgets by one. Retiring Momento's record takes it out of that population, because the record now states the retired terms rather than terms our own `free_tier_removed` change supersedes. `records_with_superseded_terms` 176 → 175, `vendor_pages_withholding_superseded_terms` 175 → 174, `ungated_pages_withholding_superseded_terms` 166 → 165. #1384 lets a data run raise all three, so this does not pin them.

Cross-check while measuring #1395: I counted the vendor pages rendering *"We are not publishing our stored … terms beside it"* on production and got 175, which is the budget on `main` exactly.

Still `data/` only.

robhunterclaude:
Rebased onto `1a78be0` (post-#1397). Diff is still data-only — `data/index.json` and `data/quality_budgets.json` — and `ratchet-quality-budgets.js --json` reports nothing over budget on the branch: measured 175 / 174 / 165 against the same three.

Holding this. The one remaining failure is real, not a stale assertion: `test/retired-records.test.ts:196` wants some page to cite `gomomento.com/pricing` as the source of the change recorded against Momento, and once the offer is retired the page prints that URL six times as prose and renders zero `class="change-source"` anchors. No data change reaches it. Diagnosis and the mechanism are on https://github.com/robhunter/agentdeals/issues/1380, which I have moved to the top of the order.

robhunterclaude:
Unblocked. https://github.com/robhunter/agentdeals/pull/1399 merged, and `test/retired-records.test.ts:196` no longer depends on `/vendor/momento` rendering the superseded-terms block: `/pricing-changes` and `/changes` cite `https://www.gomomento.com/pricing` from the change record itself.

I applied this branch's two data files on top of #1399 before merging and ran that file: 10 of 10 pass. The same data on the pre-#1399 `main` fails the assertion at line 196.

This is yours to merge — I have not touched the data.

Issue #1352, opened 2026-09-05
97 change records cite no source, and 15 set a vendor's public risk verdict — Storyblok's is disproved by a page capture two days older than the record


robhunterclaude:
## What one page says

`/vendor/storyblok` renders all three of these today:

- a **risky** badge — "Why risky: 2026-04-13 — No permanent free tier, 45-day trial only"
- **Free Tier Details**, verified 2026-08-24 — "Free Starter plan: 1 space, 1 user seat, 100K API requests/month, 2 locales, 200 components"
- schema.org JSON-LD — `{"@type":"Event","name":"Storyblok pricing change: free tier removed","startDate":"2026-04-13"}`

The record is wrong. The Internet Archive's [2026-04-11 capture](http://web.archive.org/web/20260411170723/https://www.storyblok.com/pricing) — two days *before* the record's own date — already reads "Starter … Monthly price if billed monthly: **Free**". The live page reads the same today. The 45-day trial the record names is a trial of the paid Growth Plus plan, offered on top of the free Starter plan; the hero advertises the trial and the plan table below it lists Starter at Free.

I have retracted that row in https://github.com/robhunter/agentdeals/pull/1351. This issue is about the other 96.

## The batch

97 of 527 change records carry an empty `source_url`. They are not scattered:

| | |
|---|---|
| `date` | 2026-04-12 (47) and 2026-04-13 (50). No other day. |
| `date_source` | `hand_written`, all 97 |
| `impact` | `low`, all 97 |
| `change_type` | `free_tier_removed` 35, `rebranded` 19, `product_deprecated` 17, `limits_reduced` 15, `pricing_restructured` 6, `limits_increased` 5 |

For contrast, the 430 records that do cite a source are spread across `high`/`medium`/`low` and are mostly `discovered`. So this is one two-day import, and every record in it asserts a vendor's pricing changed while citing nothing.

`source_url` is listed in `required` for the change schema in `src/openapi.ts:1018`. These 97 do not meet the contract we publish.

## What rests on them

`RISK_DEMOTION` maps `free_tier_removed` → `risky` and `limits_reduced` / `pricing_restructured` / `restriction` → `caution`. Running `vendorRiskAssessment` over the current data: **230 vendors in the index carry a non-stable risk level, and for 15 of them the winning `risk_cause` is one of these uncited records.** After the Storyblok retraction, 229 and 15.

| level | vendor | the whole basis for the verdict |
|---|---|---|
[...]

PR #1381, opened 2026-09-05
Publish the real free tiers our own change log already recorded — Firecrawl and Inngest

robhunterclaude:
Two vendor records understate a free tier that our own change log already recorded as larger. Both verified live on 2026-09-05, read from the plan table rather than the hero.

**Firecrawl.** `firecrawl.dev/pricing` prices the Free Plan at **1,000 credits / month, $0/month, 500 searches or 1,000 pages scraped, 2 concurrent requests**. Our record said `500 credits (up to 500 pages)` — half the real allowance. The `limits_increased` record of 2026-09-01 reads *"The free tier now offers 1,000 credits/month instead of 500."*

**Inngest.** `inngest.com/pricing` prices the Hobby plan at **$0/mo: 50k executions, 5 concurrent steps, 500 MB span data ingested, 10K scores, 500k events ingested, 100k queue depth, 50 realtime connections**. Our record said `100K events` and `3 users`; the page states 500K events and no user cap on Hobby. The `limits_increased` record of 2026-08-28 says both.

## Why this is worth doing now

Since #1379, both records were also withholding their vendor page: the change quotes the stored description as the previous state, so `/vendor/firecrawl` and `/vendor/inngest` published no free-tier figures at all. Updating the description breaks that match, so the pages come back — with the correct, larger numbers rather than the stale ones.

This is the exit I said I would own on https://github.com/robhunter/agentdeals/issues/1103. 46 of the withheld pages are withheld because the free tier improved, and confirming a number went up is the cheapest check on that list. Two records as a worked example of the shape.

Ratchet before and after, both edits: `stale_fact_pages` 57, `unsourced_tier_a` 41, `uncited_change_records` 95, `source_checks_ok_without_quoted_evidence` 614. Unchanged, nothing over. No route is derived from `description`, so the derived set is untouched.

**Not in this PR, noted for https://github.com/robhunter/agentdeals/issues/1352:** Firecrawl also carries a hand-written `limits_reduced` record from 2026-04-13 — *"Free credits now one-time 500 (not recurring monthly)"* — with an empty `previous_state` and an empty `source_url`. The live page says `1,000 credits / month`, explicitly recurring. It is disproved, and it is one of the 95.

**Skipped deliberately:** Resend is in the same 46 and I did not touch it. `resend.com/pricing` has Transactional and Marketing tabs, and our record's `1K marketing contacts/month` and `5 AI credits` are absent from the Transactional column I read. Correcting from one tab against a record written from another is how dub.co's record went wrong.

### DIFF
--- a/data/index.json
+++ b/data/index.json
@@ -3165,7 +3165,7 @@
     {
       "vendor": "Inngest",
       "category": "Background Jobs",
-      "description": "Serverless workflow orchestration — 50K function executions/month, 100K events, 5 concurrent steps, 3 users, unlimited environments",
+      "description": "Serverless workflow orchestration — 50K executions, 500K events ingested, 5 concurrent steps, 500 MB span data ingested, 100K queue depth, 50 realtime connections",
       "tier": "Hobby",
       "url": "https://www.inngest.com/pricing",
       "tags": [
--- a/data/index.json
+++ b/data/index.json
@@ -2298,7 +2298,7 @@
     {
       "vendor": "Firecrawl",
       "category": "Web Scraping",
-      "description": "Web scraping API — 500 credits (up to 500 pages), 2 concurrent requests, no credit card required",
+      "description": "Web scraping API — 1,000 credits/month (500 searches or 1,000 pages scraped), 2 concurrent requests, no credit card required",
       "tier": "Free",
       "url": "https://www.firecrawl.dev/pricing",
       "tags": [


robhunterclaude:
CI confirms the prediction. Three failures, all three the hardcoded population counts, nothing else:

```
test/gated-vendor-answers.test.ts:279   220 !== 222
test/superseded-terms.test.ts:143       235 !== 237
test/superseded-terms.test.ts:147       233 !== 235
```

Two records corrected, two slots emptied, and every assertion that fires is counting the slots rather than checking the data. The data half is clean — both figures were read live at the plan table this afternoon.

Blocked on https://github.com/robhunter/agentdeals/issues/1383. I am not editing `test/` to move a constant; PM PRs are data-only.

Issue #1421, opened 2026-09-07
23% of vendor page ledes are cut mid-word, and #1420 made that sentence the first thing a retriever reads


robhunterclaude:
## What

`#1420` made every page's `<meta name="description">` its opening sentence, so the description is now the first thing in a retriever's extraction window on all 2,570 pages. I verified that property independently on production and it holds: 481 of 483 in `sitemap-pages.xml` and 1,562 of 1,562 vendor pages state their own claim before the menu.

On vendor pages that sentence is composed at `src/serve.ts:4391`:

```ts
const descLimits = publishableTerms.slice(0, 100).replace(/\.\s.*$/, "");
...
`${vendorName} free tier includes ${descLimits}.${metaVerifiedSentence}...`
```

A raw `.slice(0, 100)` with a period appended. Measured on production against all 1,573 URLs in `sitemap-vendors.xml`:

| | |
|---|---|
| vendor pages taking this branch | 1,353 |
| …whose terms clause is cut at exactly 100 characters | **367** |
| …where the cut lands inside a word | **234** |
| …that sever a number from its unit | 122 |
| …that leave a bracket unclosed | 32 |
| descriptions containing `..` | 128 |

All 367 matched back to a stored record, so **we hold the full terms and cut them.** This is not missing data.

## What it publishes

| page | opening sentence ends | the record says |
|---|---|---|
| `/vendor/vercel` | `…4 hrs Active CPU, 360 GB-h.` | `360 GB-hrs Provisioned Memory` |
| `/vendor/neon` | `…100 CU-hours/month per project, 0.5 GB stora.` | `0.5 GB storage per project` |
| `/vendor/firebase` | `…Cloud Functions (2M invocation.` | `(2M invocations/month), Hosting…` |
| `/vendor/render` | `…PostgreSQL (256 MB RAM, 30-day expiry), free R.` | `free Redis (25 MB)` |
| `/vendor/sentry` | `…application logs, 1GB attachment.` | `1GB attachments` |
| `/vendor/azure` | `…Cosmos DB (1K RU/s + 25 GB lifetime), App Service (.` | continues |

The severed part is the sub-quota, the unit and the expiry — the specific cell that is the reason a model quotes us rather than the vendor's own marketing page.

## The correct implementation is the sibling branch of the same ternary

[...]

Issue #1343, opened 2026-09-04
Retrieval is why we score zero on monitoring and LLM-API questions — both pages rank 1 for a query shaped like their title and are absent from a query shaped like the need


robhunterclaude:
The 2026-09-04 citation benchmark found two question families where we score 0 of 20 paired samples while live competitor sites get cited: error tracking / monitoring, and LLM API free credits. We have a tier-A page for each, both return 200, so it was never coverage. I ran 17 further calls to find out whether the cause was the shape of the page or whether a retriever reaches it at all.

It is retrieval. And the reason is narrow enough to fix.

## What the measurement separates

OpenRouter returns `annotations` (the URLs retrieved into context) separately from the answer text. Those two are not the same set — `db`/Gemini retrieved `agentdeals.dev` and then wrote an answer that never mentions us. So retrieval and selection are independently observable, which is what makes the question answerable.

**Generic prompt, the way a developer actually asks, 3 model families:**

| prompt | retrieved us | cited us |
|---|---|---|
| monitoring | 1 of 3 | 1 of 3 |
| LLM API free credits | 0 of 3 | 0 of 3 |
| Postgres free tier (control, we score here) | 2 of 3 | 1 of 3 |

**Conditional on being retrieved, we get used.** 3 of 5 cells where we were in the retrieved set produced an answer citing us. Every cell where we were absent from the retrieved set produced an answer without us — 12 of 12. Nothing about the page's shape gets a chance to matter.

## Not an index gap, and not retrieval depth

Two controls:

- **Depth.** Re-ran both generic prompts at `max_results: 20`. Still absent, 4 of 4. We are not ranked 6th out of 5 — we are not a candidate.
- **Index.** Re-ran with the queries rewritten to resemble our own page titles: *"Compare monitoring and observability free tiers: Datadog vs Grafana Cloud vs New Relic vs Better Stack, 2026 limits"* and the LLM equivalent. **`agentdeals.dev` came back rank 1 on both.**

So both pages are indexed and rank first for a query shaped like their title. They are invisible to a query shaped like the need.

## The gap is the title

| | our title | what the question asks |
|---|---|---|
| `/monitoring-comparison-2026` | Monitoring & Observability Comparison 2026 — Datadog vs Grafana Cloud vs New Relic vs Better Stack | *"free tiers ... for **error tracking** and application monitoring"* |
[...]

PR #1422, opened 2026-09-07
Give every vendor lede one truncation rule that marks what it cut (#1421)

robhunterclaude:
Refs #1421.

The free-tier branch of the vendor description took `publishableTerms.slice(0, 100)`, dropped everything after the first full stop, and appended a period. Since #1420 made the description the page's opening sentence, that cut is the first thing an extractor reads on 1,362 of 1,573 vendor pages.

`openingOfTerms` — the truncator the `termsSuperseded` branch already used — moves to `src/terms-opening.ts` and now serves all three branches of the description and the quick verdict. Four rules, one implementation:

- stop on a whole sentence if one fits the cap, otherwise on a word;
- never end on a figure whose unit follows, or on the separator before it;
- close a bracket the clip would have left open, rather than dropping what is inside it;
- **mark every clip**, including one that lands on a full stop.

`punctuated` moves with it, so the template no longer appends a period of its own.

## Measured over every URL in `sitemap-vendors.xml`

Each page's published terms are matched against its own record by longest common prefix, so the numbers do not depend on parsing sentence boundaries out of the rendered text and hold for both spellings.

| | production | this branch |
|---|---|---|
| vendor pages swept | 1,573 | 1,573 |
| …on the free-tier branch | 1,362 | 1,362 |
| ledes cut inside a word | **246** | **0** |
| ledes shorter than their record | 1,149 | 975 |
| …that say so | **0** | **975** |
| ledes ending on a figure whose unit follows | 4 | 0 |
| ledes with an unclosed bracket | 38 | 0 |
| ledes containing `..` | 129 | 2 |

`/vendor/vercel` ends `…4 hrs Active CPU…` instead of `…4 hrs Active CPU, 360 GB-h.`; `/vendor/netlify` ends `…bandwidth at 20 credits/GB, compute at…).` instead of `…compute at 10 credits/GB-.`; `/vendor/cloudflare-workers` ends `…10ms CPU time per invocation…` where it used to state one of six sub-quotas as if it were the whole tier.

The quick verdict took the same raw slice at 120 characters and is now derived the same way, so the sentence two lines under the lede no longer stops mid-word either.

## The two open questions

**A trailing `…` is still a cut.** The `<meta>` tag and the body lede stay the same sentence — splitting them would put a page's claim in two places, which is what #1343's AC-6 exists to prevent. Nothing is withheld by keeping one: the complete stored terms are already on the same page, in the **Free Tier Details** block, and now the lede says it is shorter than them.

**The 165-page control moved on 69 of its 173 pages, and only additively.** 104 are byte-identical. 57 differ by one character — a full stop became the clip marker, because that branch stops on a sentence boundary too and `/vendor/github-actions` was publishing 76 characters of a 387-character reading with a full stop after it. 12 more gained the marker in place of an exclamation mark, or the closing bracket the clip had dropped. No page on that branch publishes a different word than before. Marking a clip that was silent is the same defect being fixed, not a regression in the reference; the tests below assert the property rather than the bytes, so they catch a real regression at any catalogue size.

## Tests

`test/vendor-lede-truncation.test.ts` sweeps all 1,573 vendor URLs at concurrency 8 in 1.6 seconds and asserts one acceptance criterion each: no cut inside a word, every short lede marked, no severed figure, no hanging separator, no unclosed bracket, no invented `..`, one implementation, and the withheld branch quoting its reading verbatim. Every assertion is a `deepStrictEqual` against an empty offender list with a vacuity guard well under the live population, so none of them measures whether the catalogue has shrunk. All eight fail on `main`.

The rule's own tests cover the whole catalogue at four caps: an opening is either the record word for word or an opening that says it is one.

4,166 tests, 0 failures from this change. 13 mutants, 13 killed — including both raw slices restored, the marker removed, the sentence branch left unmarked, the bracket left open and the severed-figure guard removed.

`test/stack-page-verdicts.test.ts` gains one assertion: the limit cell must fill the width it is given. Its existing check that the cell "never truncates into a sentence that reads complete" passed once `openingOfTerms` began marking sentence stops, which would have let the limit column quietly fall back to it.

## Not fixed here

- `/vendor/google-ads` and `/vendor/mockaroo` publish `..` because their stored record and its change reading contain one. The render invents none.
- `src/stack-claim.ts` holds a third truncator, `limitCellText`. It deliberately skips the sentence preference; now that every clip is marked, the reason for that difference is weaker, but converging them is a separate change.
- The median description is 176 characters, up from 163, because a clip now keeps the bracketed detail it used to drop.

### DIFF
--- a/scripts/census-1421-lede-truncation.mjs
+++ b/scripts/census-1421-lede-truncation.mjs
@@ -1,0 +1,173 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+
+const args = process.argv.slice(2);
+const base = args[args.indexOf("--base") + 1] ?? "http://127.0.0.1:8791";
+const repo = path.resolve(args[args.indexOf("--repo") + 1] ?? ".");
+const limit = args.includes("--limit") ? Number(args[args.indexOf("--limit") + 1]) : Infinity;
+
+const { offers } = JSON.parse(readFileSync(path.join(repo, "data", "index.json"), "utf8"));
+
+const storedTerms = new Set(offers.map(o => o.description.trim()));
+
+async function get(url) {
+  for (let attempt = 0; attempt < 3; attempt++) {
+    try {
+      const res = await fetch(url, { redirect: "follow", headers: { "user-agent": "agentdeals-census" } });
+      if (!res.ok) return null;
+      return await res.text();
+    } catch {
+      await new Promise(r => setTimeout(r, 400));
+    }
+  }
+  return null;
+}
+
+function decodeEntities(text) {
+  return text
+    .replace(/&quot;/g, '"')
+    .replace(/&#39;/g, "'")
+    .replace(/&lt;/g, "<")
+    .replace(/&gt;/g, ">")
+    .replace(/&amp;/g, "&");
+}
+
+function metaDescription(html) {
+  const m = html.match(/<meta name="description" content="([^"]*)"/);
+  return m ? decodeEntities(m[1]) : null;
+}
+
+const WITHHELD = "terms are superseded and withheld";
+const INCLUDES = " free tier includes ";
+
+function termsClause(desc) {
+  const at = desc.indexOf(INCLUDES);
+  if (at === -1) return null;
+  const rest = desc.slice(at + INCLUDES.length);
+  const stop = rest.search(/\.(\s|$)/);
+  return stop === -1 ? rest : rest.slice(0, stop);
+}
+
+function matchesAStoredRecord(clause) {
+  for (const terms of storedTerms) if (terms.startsWith(clause)) return terms;
+  return null;
+}
+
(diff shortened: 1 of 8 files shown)


robhunterclaude:
CI is red on three assertions in `test/documented-route-citation.test.ts`, all about `/api/digest/weekly` publishing no `checked` or `verified` date. They are red on `main` at 23c49c4 in run 34068456464 at 00:01Z, before this branch existed, and this branch adds no failure of its own: 4,168 tests, the same 3 failures locally and in CI.

The cause is the ISO week rolling over. `/api/digest/weekly` is listed as a route that rests on a dated record, and this week's digest holds `total_changes` 0 with `verified_records: 0`, so the route correctly omits the two date fields and the test asserts them unconditionally. It goes green again when the first change of the week lands. Reported on #1190.

Issue #1337, opened 2026-09-04
A test enumerates which vendors may be demoted, so a correct demotion freezes the catalogue — re-verification has not pushed since 10:36Z


robhunterclaude:
The catalogue stopped advancing at 10:36Z today. `Daily rolling re-verification` produced 27 new change records, the suite refused them, and they are held on `data-quarantine/rolling-reverification-20260904T105424Z-0c2754c`. Auto-filed as #1335.

What went red is a single assertion — `test/ranking.test.ts:421`, "Databases: Firebase is the only demotion, on a named recorded fact":

```
assert.deepStrictEqual(vendorsOf(r.demoted), ["Firebase"]);
```
```
actual:   [ 'Momento', 'Firebase' ]
expected: [ 'Firebase' ]
```

**Momento's demotion is correct.** I verified it against the vendor this cycle: `gomomento.com/pricing` prices Cache Flex from $13 per GB-month, the word "free" appears once on the whole page in the phrase "freeing up your team", and the only `$0` on it is a support plan. Our record's 5 GB/month free allowance no longer exists. `/vendor/momento` currently reads "Momento Free Tier 2026 · **stable** — zero pricing changes recorded".

So the assertion is refusing correct data, and the site is publishing a free tier that ended because the correction cannot land.

## Why this is worse than #1321 and #1327

Those two are the same defect — a constant describing the catalogue's contents living where a data-only job cannot change it. This one is not measuring quality at all.

The batch it refused contains three *false* free-tier removals, on Discord, Google Meet and Windscribe (#1336). The assertion did not catch them. It fired on Momento, which is the one **correct** record in the batch. Had Momento been in any other category, all three falsehoods would have shipped to `main` and to readers, and this job would have reported success.

A gate that blocks correct data and passes incorrect data is not a filter — its verdicts are uncorrelated with what it is supposed to measure.

It also compounds daily. As #1335 puts it: the next run picks the same entries, finds the same things, and is refused again. Momento is in the re-verification queue and will be re-detected every day until the assertion changes.

## Acceptance criteria

1. `Daily rolling re-verification` pushes to `main` again, confirmed by a `data(auto)` commit dated after this issue closes.
2. `test/ranking.test.ts` no longer enumerates which vendors may be demoted in a category. It asserts the rule instead — a demotion is present if and only if a recorded fact justifies it. Demonstrate both directions: a new correct demotion passes, and a demotion with no recorded fact fails.
[...]

PR #1445, opened 2026-09-07
Replay a run's data onto the main it moved to rather than discarding it (#1337)

robhunterclaude:
Refs #1337.

The issue's measurement reads "the scheduled run has shipped data on 6 of its last 14 attempts". Four of those failures have causes on record. Three are the suite refusing the batch. **2026-09-06 is not** — the suite's verdict that day was *push*, and the push was rejected:

```
every failing test file reports on how current our own reading is, not on whether
the data is right: test/stale-page-facts.test.ts
To https://github.com/robhunter/agentdeals
 ! [rejected]        HEAD -> main (fetch first)
error: failed to push some refs
##[error]Process completed with exit code 1.
```

`main` had gained `c75ddb1 data(auto): analytics rollup` and `c05e253 data(auto): link liveness` during the 11 minutes the run took. Two sibling scheduled jobs, both writing only `data/`.

**That day is worse than a quarantine.** `push_to_main` runs as a bare statement under `set -e`, so the rejection ends the script before `quarantine` is reached. The step output says `quarantined=false` and `pushed_over_failures=true`; the workflow then reports *shipped-over-failures* for a run that shipped nothing, and there is no ref holding the commit. 27 detections existed only in the runner's workspace and expired with it.

## What this does

The suite run, the verdict and the push become one loop. When the push is rejected the gate fetches `main`, rebases this run's commit onto it, rebuilds, and **runs the suite again** before pushing — so what reaches `main` is a tree the suite read, which is the guarantee the gate exists for. Two replays, then it quarantines.

Every other way out is unchanged: a build failure, an unmeasurable budget and a suite that names a blocking file still quarantine exactly as before, and the excused-failure path still pushes with the same step outputs and the same message.

Three cases the gate could not reach before:

- `main` moved to a commit that does not touch this run's paths → the data lands on `main`, on top of the sibling's commit, with the suite having read both. The test asserts the sibling's file survives and that the suite ran twice.
- `main` moved to a commit that touches the same file → the rebase aborts and the commit goes to a quarantine ref. The test asserts the data is on a ref rather than nowhere.
- `main` keeps moving → bounded by `GATE_REPLAYS_ONTO_A_MOVED_MAIN`, default 2, then quarantine.

## One existing test had to change shape

`tells the refusal issue which of the three things went wrong` compared the gate's quarantine reasons against a literal list of three, so adding a fourth reason failed it — the same shape as the assertions in #1444. It now asserts the property: every `quarantine` call states a distinct reason of at least four words, the three that existed are still among them, the reason reaches `quarantine_reason`, and `report-data-push-outcome.sh` interpolates it into the issue body.

## Verified

`test/data-push-gate.test.ts`: 44 tests, 44 pass, 3 new. The three new cases drive the real `scripts/gate-data-push.sh` against a fixture origin whose `main` is moved from a second clone between the fixture's push and the gate's.

### DIFF
--- a/scripts/gate-data-push.sh
+++ b/scripts/gate-data-push.sh
@@ -55,8 +55,20 @@ export GATE_FAILING_FILES
 export GITHUB_OUTPUT="$SUBPROCESS_OUTPUT"
 trap 'rm -f "$LOG" "$VERDICT" "$GATE_FAILING_FILES" "$SUBPROCESS_OUTPUT"' EXIT
 
+REPLAYS=0
+REPLAYS_ONTO_A_MOVED_MAIN="${GATE_REPLAYS_ONTO_A_MOVED_MAIN:-2}"
+
 push_to_main() {
   git push origin HEAD:main
+}
+
+replay_onto_main() {
+  git fetch origin main || return 1
+  if ! git rebase FETCH_HEAD; then
+    git rebase --abort || true
+    return 1
+  fi
+  COMMIT="$(git rev-parse --short HEAD)"
 }
 
 quarantine() {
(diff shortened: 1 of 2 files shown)


robhunterclaude:
Added a fourth case after checking what `actions/checkout@v4` actually gives the workflows: none of the three gated workflows sets `fetch-depth`, so every one of them runs the gate in a shallow clone. `replays onto it from the shallow checkout the workflows actually run in` clones the fixture origin at depth 1 over `file://` and asserts `--is-shallow-repository` before running the gate, so the rebase is exercised against a grafted history rather than a full one.

It passes with the plain `git fetch origin main` — git deepens the fetch far enough to reach the common ancestor — so I removed the `--unshallow` fallback I had written for it rather than keep a branch no test can distinguish. If a future git ever cannot reach the base, `replay_onto_main` returns non-zero and the data is quarantined, which is the safe direction.

45 tests in the file, 45 pass.

robhunterclaude:
One more case, and it is the exact shape of 2026-09-06.

The step outputs used to be written *before* the push. A run whose suite was red but wholly excused therefore wrote `pushed_over_failures=true` and then died on the rejected push — which is why that day's workflow reported *shipped-over-failures* for a run that shipped nothing. They are now written after the push succeeds, so an excused red run that cannot reach `main` reports `quarantined=true` and holds the data on a ref.

`reports a refusal rather than a shipment when an excused red run cannot reach main` drives it: mode `excused`, a sibling commit on the same file so the replay cannot happen, and it asserts `pushed_over_failures=true` is absent from the outputs.

46 tests in the file, 46 pass.

Issue #1481, opened 2026-09-09
28 pages tell readers and extractors we were last updated in March — 1,536 of our 1,580 records were verified in the last ten weeks


robhunterclaude:
28 published pages tell readers and extractors we were last updated in March. 1,536 of our 1,580 records were verified in July, August or September 2026.

`/free-llm-apis` renders this sentence in its most heavily weighted window of extracted text:

```
Compare 25+ free LLM API providers — Groq, Cerebras, OpenRouter, Gemini, Mistral,
OpenAI, Anthropic, NVIDIA NIM, and more. Exact rate limits and token quotas.
Updated March 2026.
```

The page lists Mistral AI. Our record for Mistral AI carries `verifiedDate: 2026-09-05`. The page also says, separately, `verified March 2026`.

## Three surfaces, one string

The string is a single `metaDesc` literal and it reaches every surface a consumer reads:

| surface | who reads it |
|---|---|
| `<meta name="description">` | search and retrieval snippets |
| JSON-LD `ItemList.description` | structured-data extractors |
| a visible `<p>` immediately above `<nav class="global-nav">` | a human, and the first ~200 characters of extracted body text |

That third one is the same window #1343's evidence measures when it reports that the first 400 characters of the monitoring page are mostly nav bar. On these 28 pages the text that gets there is a claim that the page is five to six months old.

## Scope

30 slugs carry a hardcoded `Updated <Month> 2026` in `src/serve.ts` — 58 occurrences, each declared twice, once in the page-definition array and once as a `const metaDesc` in the route handler. Served live, **28 of the 30 publish it on all three surfaces**; 46 occurrences say March, 12 say April.

Affected pages include tier-A surfaces: `/free-llm-apis`, `/ai-free-tiers`, `/storage-alternatives`, `/database-alternatives`, `/hosting-alternatives`, `/monitoring-alternatives`, `/supabase-vs-firebase`, `/vercel-vs-netlify`, `/neon-vs-supabase`, `/database-pricing`, `/vector-database-pricing`, `/startup-credits`.

## The claim is false, and false in the pessimistic direction

`verifiedDate` distribution across all 1,580 offers:

| month | records |
|---|---|
| 2026-04 | 29 |
| 2026-05 | 7 |
| 2026-06 | 8 |
| 2026-07 | 626 |
| 2026-08 | 796 |
| 2026-09 | 114 |

[...]

Issue #1081, opened 2026-08-27
18 comparison pages render an August verification date over prose frozen on 2026-04-04 — the provenance register covers 67 of our 269 content pages


robhunterclaude:
18 comparison pages render an August verification date over prose frozen on 2026-04-04. The provenance register covers 67 of our 269 content pages, and these are among the 202 it does not reach.

## The page

`/backblaze-b2-vs-cloudflare-r2` renders, in this order:

```
Data verified through 2026-08-16
Quick Verdict: Cloudflare R2 is the clear winner for most use cases…
Key Differences
  Egress: R2 has zero egress fees — always. B2 charges for egress ($0.01/GB)…
  At scale: At 1 TB stored + 10 TB egress, R2 costs ~$15/month vs B2 at ~$60/month…
```

The date comes from the record's `verifiedDate`, which #1058 establishes means the URL returned 200. Everything under it comes from `VS_PAGES` in `src/serve.ts`. `git blame -w` on that array: **196 of 198 lines authored 2026-04-04, the other 2 on 2026-04-17.** Nothing has been edited in 145 days.

So the freshest-looking date on the site sits above the oldest prose on the site, and the date is not about the prose at all.

The content is wrong too, in the way cycle 57's review of `/storage-comparison-2026` already found: B2's published terms give free egress up to 3× monthly average storage before the $0.01/GB rate applies. "B2 charges for egress ($0.01/GB)" states the rate and omits the allowance. The at-scale figure is ~$60 where the vendor's current rates give roughly $77 (1 TB at $6.95, 7 TB billable egress at $0.01/GB). The verdict direction survives at 10:1 egress; the numbers do not.

**All 18 of these pages render a "Data verified through" date, every one of them August 2026.** `/cockroachdb-vs-mongodb`, `/neon-vs-turso`, `/cloudflare-pages-vs-vercel`, `/netlify-vs-render`, `/datadog-vs-sentry`, `/datadog-vs-grafana-cloud`, `/auth0-vs-clerk`, `/auth0-vs-kinde`, `/circleci-vs-github-actions`, `/github-actions-vs-gitlab-ci`, `/brevo-vs-resend`, `/postmark-vs-resend`, `/backblaze-b2-vs-cloudflare-r2`, `/cloudinary-vs-imagekit`, `/amplitude-vs-posthog`, `/amplitude-vs-mixpanel`, `/groq-vs-hugging-face`, `/groq-vs-mistral-ai`.

[...]

Issue #1061, opened 2026-08-26
"Last updated" is generated by new Date() at render time — 45 pages report today's date regardless of whether anything changed


robhunterclaude:
## `Last updated` is the date you loaded the page

`src/serve.ts` renders the freshness line on our editorial and comparison pages as:

```
Last updated ${new Date().toISOString().split("T")[0]}
```

It is evaluated at request time. It prints today's date, every day, on every one of these pages, whether or not anything on the page has changed. It will say `Last updated 2027-04-11` on 2027-04-11.

**45 occurrences** of `Last updated ... new Date()`, 47 counting the `Updated`/`Last updated:` variants. They are on the guide pages, the `/best/*` category pages, the vendor-vs-vendor comparisons, the risk index and the free-tier comparison pages.

### Proof that the content is not being updated

`/database-free-tier-comparison-2026` currently serves:

> Published 2026-03-31 · Last updated 2026-08-26 · Data verified from our index of 1,571 developer tools · 44 database services compared

The headline verdict cards on that page — `Supabase / Most Generous All-in-One`, `Neon / Best Pure Postgres`, `PlanetScale / Free Tier Removed` — were written in commit `2a43835` on **2026-03-31** and have not been edited since. 148 days. The page says it was updated today.

`/supabase-vs-firebase` makes it plainer still, because it carries both dates at once:

- header: `Last updated 2026-08-26`
- body: `Supabase and Firebase free tiers **as of March 2026**`

Its verdict block last changed in `c3d3471`, 2026-03-26.

Parts of these pages *are* data-driven — the comparison tables interpolate from `data/index.json`, and the per-vendor context boxes print a real `verifiedDate`. But the verdicts, the stat cards, the executive summaries and the prose are hand-written and static, and the header date makes a claim about the whole page.

### Why this is worse than a stale date

A stale date is a fact that has aged. This is not a fact at all — it is a timestamp of the reader's own visit, rendered as a claim about our editorial process. A reader comparing two of our pages cannot use the date to tell which one is maintained, because every page reports the same date: theirs.

It also compounds #1058. That issue is about `verifiedDate` recording an HTTP 200 rather than a reading of the page. This is the same class of defect one level up: a freshness signal generated by machinery that never looked at the thing it describes. The two together mean a reader has no true freshness signal anywhere on the site.

### Acceptance criteria

1. **No page renders a date derived from `new Date()` as a description of its own content.** Applies to all 45–47 occurrences.
[...]

PR #1482, opened 2026-09-09
Date every page from the records it lists (#1481)

robhunterclaude:
Refs #1481

30 pages told readers and extractors we were last updated in March or April 2026. 1,536 of our 1,580 records were verified in July, August or September. The string is one literal per page and it reaches three surfaces — the meta description, the JSON-LD `description`, and the visible paragraph that opens the extracted body text — so one wrong sentence was wrong three times.

The date is derived now, from one value per page.

## Two families, two sources

- **A page the catalogue perturbation moves** takes the span of the records it links: `Verified July to September 2026.` The span is the claim that holds for every record on the page. The newest date alone is the direction #1061 removed; the oldest alone reads worse than the literal it replaces.
- **A page it does not move** takes its own review record: `Compiled 2026-04-09, not re-checked since.` Seven of the thirty are hand-compiled tables. Stamping a September catalogue date on April figures is #1081's defect, so those get the date their figures were actually compiled.

`reads_index` is the split. It is a declared field, but `pageSourceViolations` already holds it to a catalogue-perturbation measurement, so it is not a declaration this change has to take on trust.

## How it resolves

A page declares `[[freshness]]`. The response wrapper that already inserts the lede resolves it — from the vendor records the rendered page links to, or from the page's review record. Reading the served HTML rather than a list passed in is what makes "the records it lists" literally true, and it is why the three surfaces cannot drift apart: they are one string, substituted once.

## Scope

The issue's grep finds 58 occurrences in page descriptions. The served population is **93 across six shapes**; four of them are body prose on the same pages, and two would have contradicted the derived date rendered two lines above them — `/supabase-vs-firebase` said `free tiers as of March 2026` under a lede that now reads `Verified August 2026`, and the seven hand-compiled pricing tables opened with `All prices verified as of April 2026`.

`/openai-assistants-migration-2026` also dated its cost table by hand; it reads its review record directly rather than through the placeholder, because that sentence is about the table and not about the page.

## Left alone, on purpose

- `/free-saas-stack` and `/free-startup-stack` date GitHub's own runner pricing change. That is a fact about a vendor, not a claim about us.
- `/hetzner-pricing-2026` says two rows "were verified in March 2026 and we have not re-read them". A page-level date would lose what that sentence exists to say.

## Tests

`test/derived-page-freshness.test.ts`, 18 tests. Every URL in `sitemap-pages.xml` is fetched, and a served freshness date must equal one recomputed inside the test from `data/index.json` and `data/page-reviews.json` — not from the module under test, so mutating the module cannot move both sides of the comparison. The same sweep asserts the claim reaches the meta description, a JSON-LD description and the visible lede, and that no page serves an unresolved placeholder. A separate check fails on any hardcoded month-year freshness claim left in `src/serve.ts`, which is what keeps the property true when the 31st page is added.

`scripts/mutate-1481.mjs` is 11 mutants, all killed, wired as `npm run mutate:1481`. `npm run mutate:1479` is wired too.

Suite 4,606 — 4,604 pass. The 2 failures are #1190's and reproduce on `main` with this branch stashed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,8 @@
     "serve": "node dist/serve.js",
     "dev": "ts-node src/index.ts",
     "test": "node --test --test-concurrency 1 test/**/*.test.ts",
+    "mutate:1479": "node scripts/mutate-1479.mjs",
+    "mutate:1481": "node scripts/mutate-1481.mjs",
     "test:gated": "node --test --test-concurrency 1 --test-reporter=spec --test-reporter-destination=stdout --test-reporter=./scripts/reporters/failing-test-files.js --test-reporter-destination=$GATE_FAILING_FILES test/**/*.test.ts",
     "ratchet:budgets": "node scripts/ratchet-quality-budgets.js --lower",
     "lastmod:pages": "node scripts/update-page-lastmod.js",
--- a/scripts/mutate-1481.mjs
+++ b/scripts/mutate-1481.mjs
@@ -1,0 +1,84 @@
+import { readFileSync, writeFileSync } from "node:fs";
+import { execFileSync } from "node:child_process";
+
+const SUITE = ["test/derived-page-freshness.test.ts"];
+
+const MUTANTS = [
+  ["the-span-collapses-to-the-newest-record-it-holds", "src/page-freshness.ts",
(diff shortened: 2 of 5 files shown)


robhunterclaude:
PR #1482 merged.

Issue #1327, opened 2026-09-04
Five exact-equality budgets make a red suite the normal state — red on a pull request stops meaning a change broke something


robhunterclaude:
Five counts in the suite are exact-equality ratchets on quality debt rather than assertions about whether anything is wrong. #1325 moved all five out of `src/` and into `data/quality_budgets.json`, which fixed the trap where clearing a page was a breaking change for whoever cleared it. What it did not change is that exceeding one of them fails `npm test`, and `tests.yml` runs `npm test` on every `pull_request`.

| budget | asserted at | moved by |
|---|---|---|
| `stale_fact_pages` | `test/stale-page-facts.test.ts` via `staleFactViolations` | any recorded change on a vendor named by a reviewed page |
| `unsourced_tier_a` | `test/page-data-provenance.test.ts:strictEqual` | the register |
| `faq_answers` | `test/faq-provenance.test.ts:247` | the catalogue |
| `faq_answers_stating_a_figure` | `test/faq-provenance.test.ts:259, 310` | the catalogue |
| `faq_answers_with_a_digit_but_no_figure` | `test/faq-provenance.test.ts:264` | the catalogue |

## Red main becomes the steady state

Measured on `433f07d` by running `staleFactPages` against the shipped register and change log, and adding one synthetic change per vendor slug to find which slugs raise the cohort:

- Cohort 57, budget 57.
- **0 of 1,572 vendor slugs raise it dated 2026-09-04. 28 raise it from 2026-09-05**, carrying 28 of 1,580 offers (1.77%).
- With the change log frozen the cohort holds at 57 for fourteen simulated days, so page aging alone does not move it. Only a new change record does.

The rolling job records 28 to 31 changes a day across the catalogue, so the cohort clears the budget in about two days. The budget does not follow it up. `stale-page-facts` then goes red, #1325's excuse list ships the data anyway, and `main` stays red until someone reviews a page — which returns the cohort to 57 for as long as it takes the next change to land on one of the 28.

Both sides of that move at roughly the same rate, so red is where `main` sits a large share of days from here on. The cost is not the data, which now ships. It is that red on a pull request stops distinguishing "this change broke something" from "the review queue is one page deeper than yesterday". `main` being red without anyone reading it is the condition #1317 was filed about.

## The direction

[...]

PR #1493, opened 2026-09-09
Report a quality budget rather than asserting it (#1327)

robhunterclaude:
Refs #1327

Seven counts of editorial debt decided whether `npm test` passed. Each rises when the catalogue is re-verified and falls when somebody reads a page, so red on a pull request stopped distinguishing "this change broke something" from "the reading queue is one page deeper than yesterday".

## What replaces them

`scripts/report-quality-budgets.js` reads all ten budgets, boots a server for the three the suite used to measure, and writes an issue body naming each budget, its ceiling, its measurement and its cohort. `scripts/report-quality-budgets.sh` opens that issue once and edits it thereafter. The daily re-verification runs both after the gate.

`npm run report:budgets` runs it by hand. `--skip-faq` skips the server.

## Criterion 1 — the suite is green with a budget over its ceiling

`node scripts/demo-1327.mjs` halves every budget a data run cannot raise, runs the files that used to hold them, and asks the reporter what it would say:

```
stale_fact_pages 58 → 29 · unsourced_tier_a 22 → 11 · uncited_change_records 98 → 49
source_checks_ok_without_quoted_evidence 508 → 254 · faq_answers 166 → 83
faq_answers_stating_a_figure 77 → 38 · faq_answers_with_a_digit_but_no_figure 41 → 20

ℹ tests 224  ℹ pass 224  ℹ fail 0
7 budgets over ceiling, 7 of them reported, suite green.
```

## Criterion 3 — a ceiling still cannot rise on its own

`ratchet()` raises only the three `QUALITY_BUDGETS_A_DATA_RUN_MAY_RAISE` names, and `raises nothing on the data the site ships, however far a measurement has run ahead` holds every other budget to `next[name] <= budgets[name]` against a live measurement. `stale_fact_pages` moved 57 → 58 by a hand edit carrying its reason.

## Criterion 4 — which assertions moved

Moved out of the suite, in `test/faq-provenance.test.ts`: the answer count, the figure-stating count and the `frozen` count became population floors; the count of answers naming a digit the figure rule does not claim was deleted, since a floor on it would assert we still hold unclassified numbers.

Stayed and still block: `leaves no answer stating a figure without one`, `takes the dates in every clause from the register rather than from the answer`, `says corrections are outstanding on no page whose review is clean`, `holds the structured date of a page whose review found defects at its publication date`, the `moved` half of `only dates an answer whose figures the catalogue cannot move`, `leaves the answers the catalogue does move without a compile date`.

In `test/page-data-provenance.test.ts` the `unsourced_tier_a` equality is gone and its neighbour is restated against the measured population, so "one more than the register holds is refused" is still asserted without pinning the number. The three assertions that fire on what a page publishes are untouched.

## A guard, so the pattern cannot come back quietly

`test/budget-assertions.ts` reads every file under `test/` and refuses three shapes: a budget compared to anything that is not another budget, a call that reaches a budget through a defaulted argument, and a ceiling on a budget a data run cannot raise. It found a sixth site the issue did not name.

## The staleness clock

A review recording `review_outcome: "fail"` restarted the clock, so finding four errors on a page improved `stale_fact_pages` by more than fixing them would. `dateModifiedFor` already had the rule right; `reviewStatus` now carries it and `dateModifiedFor` reads it, so the two cannot drift again. Review cadence still measures from the day the page was last read, so no page changes state and `/api/page-reviews` reports the same totals.

Six pages moved, `/email-comparison-2026` joined the cohort, 338 stale facts became 346.

## Verification

- Suite 4,652 of 4,652. `tsc --noEmit` clean.
- `npm run mutate:1327`: 16 mutants, 16 killed.
- `npm run report:budgets` against the shipped data, and the announcer run end to end against the live repository.
- The workflow file parsed and its step order read back.

### DIFF
--- a/.github/workflows/reverify.yml
+++ b/.github/workflows/reverify.yml
@@ -85,6 +85,17 @@ jobs:
             bash scripts/report-data-push-outcome.sh "Daily rolling re-verification" shipped-over-failures "$NON_BLOCKING_FILES"
           fi
 
+      - name: Measure the quality budgets against the data that reached main
+        id: budgets
+        run: node scripts/report-quality-budgets.js --body /tmp/quality-budgets.md --github-output
+
+      - name: Say on an issue which budget is over its ceiling
+        if: always() && steps.budgets.outputs.over != ''
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          OVER: ${{ steps.budgets.outputs.over }}
+        run: bash scripts/report-quality-budgets.sh /tmp/quality-budgets.md "$OVER"
+
       - name: Check the change log is still being written to
         id: staleness
         if: always()
--- a/data/quality_budgets.json
+++ b/data/quality_budgets.json
@@ -1,7 +1,7 @@
 {
   "version": 1,
   "budgets": {
-    "stale_fact_pages": 57,
+    "stale_fact_pages": 58,
     "unsourced_tier_a": 22,
     "uncited_change_records": 98,
     "source_checks_ok_without_quoted_evidence": 508,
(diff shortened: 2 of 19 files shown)


robhunterclaude:
One path is not exercised live: `gh issue create` in `scripts/report-quality-budgets.sh`. Nothing is over its ceiling today, so the announcer took its no-op branch when I ran it end to end against this repository, and every branch around it is under test — the marker refusal, the empty-body refusal, the flag refusal, and a test that refuses the `--search` form the lookup must not use. Both labels it passes exist.

It should fire on its own within a day rather than sitting untested: `stale_fact_pages` is at its ceiling of 58, and a single day of re-verification has been taking it up by one. The first 06:00Z run that records a change on a reviewed page's vendor opens the issue.

Issue #1497, opened 2026-09-09
We score zero on the free-LLM-API question and 8 of 10 answers cite a GitHub markdown list — the top one is a rival's generated mirror of its own site


robhunterclaude:
We are cited in **0 of 20** model answers to "which LLM APIs have free tiers". **8 of the 10 web-search-enabled answers cite a GitHub markdown list.** We publish no GitHub artefact of any kind.

Source: the citation benchmark runs of 2026-08-26 and 2026-09-04 pooled, in `logs/citation-bench.jsonl` and `logs/citation-bench-20260904.jsonl`. Ten questions, five model families, two arms.

## The measurement

Arm B is web search on, five model families per run, two runs — ten answers.

| run | OpenAI | Google | xAI | Anthropic | DeepSeek |
|---|---|---|---|---|---|
| 2026-08-26 | — | GitHub ×2 | GitHub (raw) | github.io | GitHub ×2 |
| 2026-09-04 | GitHub ×2 | GitHub ×2 | — | GitHub | GitHub ×2 |

Eight of ten. Nineteen distinct repos and pages across the two runs. `open-free-llm-api/awesome-free-llm-apis` was cited by four different model families.

## This is specific to one question, which is why it is worth acting on

The same count for all ten questions, plus the off-domain negative control:

| question | answers citing a GitHub list | answers citing us |
|---|---|---|
| **LLM APIs with free tiers** | **7 of 20** | **0** |
| a source to check regularly | 4 of 20 | 5 |
| free object storage | 1 of 20 | 4 |
| error tracking and monitoring | 1 of 20 | 0 |
| free Postgres | 0 | 4 |
| transactional email | 0 | 4 |
| free Next.js hosting | 0 | 1 |
| auth under 10,000 MAU | 0 | 1 |
| who cut free tiers recently | 0 | 4 |
| complete free-tier stack | 0 | 6 |
| genomics pipelines (control) | 5 of 20 | 0 |

GitHub is not a general citation surface for free-tier questions. It is concentrated on the one commercial question we lose completely, and on an unrelated technical control. Every question where we already win has zero GitHub competition.

## The top-cited list is a rival running a site plus a generated mirror

[...]

PR #1499, opened 2026-09-09
Generate a public AI and LLM free-tier index from the catalogue (#1497)

robhunterclaude:
Refs #1497

Generates `artifacts/free-llm-api-index/README.md` from the catalogue: 84 AI / ML and AI Coding records, one row each, ordered alphabetically.

**AC-1 — generated, not written.** `npm run generate:llm-readme` writes the file only when the render differs; a second run on the same data reports `unchanged` and touches nothing. `--check` exits 1 when the committed file is not what the data generates, and `--on=YYYY-MM-DD` pins the clock.

**AC-2 — a date and a source per row, never one stamp for the file.** Each row is dated from the record behind it. A test asserts no `YYYY-MM-DD` appears anywhere outside a table row, so a `Last updated` line cannot be added without failing.

**AC-3 — no row states a level the catalogue withholds, and none is dropped.** The rating cell is one of `stable` / `caution` / `risky` / `ended` / `unrated`, taken from `vendorBadge` — the same function the site's seven surfaces call. 26 of 84 rows are `unrated` and each names the rule: `states_no_terms` 9, `unreadable` 9, `gate:not_a_free_offer` 4, `link_unreachable` 4. `rated + ended + unrated == rows` is asserted.

**AC-4 — a record we cannot vouch for says so.** Where the link has stopped resolving, the row publishes the day it last worked and drops the verification date entirely, including the date on the terms — the site's own rule, which cost one round of tests to get right. Where a record is older than the re-read interval the site derives (70 days today), the row says so and says it is a statement about us.

**AC-5 — the rules are rendered from the tables the classifier reads.** Every `NOT_FREE_TIER_RULES` and `TIME_LIMITED_TIER_RULES` pattern is printed with its reason, and every change type is printed under its direction, both straight from the source tables. A new change type appears in the file with no edit here; removing one from the file fails a test.

**AC-6 — the payload.** 15 rows carry `**Until <date>, our record read:** <prior terms>` under the newer reading, which is quoted from the change record's own source and dated from it. Rows without a superseding change say `Our record, read from <source> on <date>` — a description we wrote is never presented as a quotation from the vendor.

**AC-7 — same trigger as the data job.** `.github/workflows/llm-api-readme.yml` runs on push to `main` and on dispatch, regenerates, and pushes through `gate-data-push.sh`, so the suite decides whether the file reaches `main` and a refusal is reported the same way the other data jobs report theirs. The generator exits non-zero if the catalogue yields no rows.

**AC-8 — the counts are in the file.** A table at the top reports rows, rated, ended, withheld by reason, rows showing prior terms, and rows carrying a caveat. Tests assert the printed numbers against the rows.

Suite 4,755 of 4,755. `npm run mutate:1497` — 16 of 16 killed. Two survived the first pass: one because the withheld reason did not name *which* gate withheld it, and one because no AI/ML description carries a `|` today — two change records elsewhere in the catalogue already do, so the escaping earned a fixture rather than a deletion.

**Not in this PR: the public repository.** The file is generated and kept current here; publishing it under a repository of its own is a separate step, and it needs a name and a decision on how it is pushed. Both are on the issue.

---

**Correction to AC-7, in the second commit.** A push made with `GITHUB_TOKEN` does not fire an `on: push` workflow, so the job in the first commit would only have run when a person merged a PR. Every `page-lastmod.yml` run in the history is a PR merge for exactly that reason, which is why `reverify.yml` carries `GATE_UPDATE_PAGE_LASTMOD` instead of relying on the trigger. `GATE_REGENERATE_LLM_INDEX` now does the same: the rolling re-verification and the liveness check regenerate the index inside the gate, in the same commit as the records that moved it. The push-triggered job stays, because it is what covers a change to the renderer rather than to the data.

A generator failure inside the gate leaves the published index standing rather than replacing it with a stale one, and does not hold the catalogue — the loudness belongs to the dedicated job, not to a data freeze.

The gate's own tests read the list of gated workflows from the workflows that invoke the gate now, with a floor, rather than from a hand-kept array that this PR would have had to be added to.


robhunterclaude:
Reviewed against the generated file. The provenance design is right and I am not asking for changes to it. Three changes before the repository goes public, plus the naming decision you asked for: https://github.com/robhunter/agentdeals/issues/1497#issuecomment-5608026583 — and one row is a product that does not exist, filed as https://github.com/robhunter/agentdeals/issues/1500.

Issue #1220, opened 2026-09-01
Four products are indexed twice under two names — /compare/arize-ai-vs-arize-ax compares a product with itself


robhunterclaude:
Four products are in the index twice under two names. All eight pages are submitted to Google, and two of the pairs render a comparison of a product against itself.

## The four pairs

Every pair shares one pricing-page URL and one category.

| Category | URL | Records |
|---|---|---|
| AI / ML | `https://arize.com/pricing` | `Arize AI`, `Arize AX` |
| Notebooks & Data Science | `https://colab.research.google.com` | `Google Colab`, `Colaboratory` |
| Payments | `https://www.lemonsqueezy.com/pricing` | `LemonSqueezy`, `Lemon Squeezy` |
| IDE & Code Editors | `https://webcomponents.dev` | `Components.studio`, `WebComponents.dev` |

`Arize AI`'s own description reads *"ML observability platform (now Arize AX)"*, and both records publish the same free plan: 25K trace spans/month, 1 GB ingestion/month, 15 days retention.

## What it publishes today

**https://agentdeals.dev/compare/arize-ai-vs-arize-ax** returns 200 with the verdict *"Both Arize AI and Arize AX offer free tiers. Arize AI provides "Free" while Arize AX offers "Free"."* **https://agentdeals.dev/compare/components-studio-vs-webcomponents-dev** is the same shape. Neither is in `sitemap-comparisons.xml`, so this is a page a crawler or an agent reaches by link or by slug, not one we submit.

All eight `/vendor/` pages **are** in `sitemap-vendors.xml`, so we submit four pairs of near-identical pages for indexing.

The twins carry different `verifiedDate`s — Arize AI 2026-08-26, Arize AX 2026-08-20 — so the same offer publishes two freshness claims.

`/vendor/arize-ai` says *"One of 70 AI / ML services we track"*. There are 69.

## Why this matters now rather than last month

The alternatives gate (#1032) publishes same-category peers. AI / ML is the next category I am classifying, and both Arize records will carry the same subtype, so every AI / ML alternatives page will list one product twice in adjacent rows. Duplicates were nearly invisible before; a list makes each one visible on dozens of pages at once.

## Detecting it

`source_check.outcome` already separates these from the aggregator imports. Group records by normalised URL plus category, and keep the group only if no member is `does_not_name_vendor`:

[...]

PR #1503, opened 2026-09-10
Merge six duplicate vendor records into their survivors (#1220)

robhunterclaude:
Six duplicate vendor records are retired into their survivors. 1,579 records → 1,573. The registry is `data/vendor_merges.json`, shipped in #1502.

| retired | survivor |
|---|---|
| Arize AI | Arize AX |
| Colaboratory | Google Colab |
| Components.studio | WebComponents.dev |
| Hasura Cloud | Hasura |
| LemonSqueezy | Lemon Squeezy |
| Slack API | Slack |

## Every pair cites the same page

Five of the six pairs store an identical `url`; Colaboratory and Google Colab differ by a trailing slash. So these are not two products — they are two readings of one pricing page, taken on different dates, that disagree with each other. That changes what a merge is allowed to throw away, and the first version of this PR threw away four things.

A merge retires a name. It must not also retire a field. Where the survivor was missing something the retiring record held, the survivor now takes it:

| survivor | carried | from |
|---|---|---|
| Slack | the API rate-limit sentence, into `description` | Slack API |
| Hasura | `product_role` **and** `category: Databases` | Hasura Cloud |
| Lemon Squeezy | `tier: Pay-as-you-go` | LemonSqueezy |
| Arize AX, Hasura, Slack | the later `verifiedDate` | all three |
| all six | the union of `tags` | all six |

### Slack

`Slack API` was the only record in the catalogue carrying either half of *"free API access with same rate limits as paid plans. Non-Marketplace apps created after May 2025 have stricter API rate limits."* The merged `description` carries both. Verified on a running server: the sentence renders in the reader prose, in the `Offer` description in JSON-LD, and in the FAQ answer to "What does Slack's free tier include?".

### Hasura

`product_role` alone does nothing on the survivor. `partitionSubstitutes` only admits a candidate whose category the subject is classified in, and `API Development` is not a classified taxonomy — so the add-on gate never fires there. Carrying the category with it is what makes the field work.

Measured on the running server, not estimated: **35 of the 43 other `/alternative-to/` pages in `Databases` name Hasura in their "Left out of this list" line.** Without the category carry that count is 0, and `/vendor/hasura` publishes *"Hasura is listed in Databases and searchable there, and is left out of alternatives lists for vendors it cannot stand in for"* — a sentence that would be false.

`data/category-mapping.json` moves `Hasura` to `Databases` too. Only `scripts/recategorize.ts` reads that file, and leaving it at `API Development` would mean the next run of that script silently reverts this.

### Lemon Squeezy

The two records disagreed about the plan name: `Pay-as-you-go` against `Free`. I read the live page. `lemonsqueezy.com/pricing` prices ecommerce at *"5% + 50¢ ... per transaction"* under *"There are no monthly charges for ecommerce features, and we only take a small cut of each sale"*, and prices email marketing at `$0 / m` up to 500 subscribers. `Pay-as-you-go` is the accurate name for the primary product, and it came off the record with the most recent `source_check` in the whole set (2026-09-09). This changes no verdict — `hasFree` only excludes `tier === "none"` — so both records already answer "Is Lemon Squeezy free?" with Yes. That is a separate question and not this PR's.

### verifiedDate

Three of the six retiring records were **fresher** than their survivor on the same URL: Slack API by 20 days, Arize AI by 4, Hasura Cloud by 4. Keeping the survivor's older date would have discarded a real, more recent read of the same page and made the catalogue staler by merging it.

### What is not carried

`source_check`. Its `detail` string names the retired vendor — *"the page names Arize AI as \"arize ai\""* — so carrying it would leave three survivors citing a name we just retired. The next check run regenerates it.

## Verified on a running server

- All six retired slugs return 301 on `/vendor/` and `/alternative-to/`; all six survivors return 200.
- All six are absent from `sitemap-vendors.xml` (1,566 URLs).
- `/category/databases`, `/category/api-development`, `/category/communication-messaging` and `/category/payments` all 200.
- Suite: **4,837 of 4,837 pass** on this tree, run start to finish without a branch change under it. The four fixtures #1505 rewrote are green.

## One thing this does not fix

`Communication` (18 records) and `Communication & Messaging` (8) are two live categories over one domain, and the duplicate pair sat across them — `Slack API` in one, `Slack` in the other, with `Discord API` and `Discord` still split the same way. The merged record keeps the survivor's `Communication & Messaging`, because nothing in the data is destroyed by that choice and reclassifying is a different decision from merging. Measured and recorded on https://github.com/robhunter/agentdeals/issues/1507 — five of the 65 categories are name-subsets of another.

Refs #1220


robhunterclaude:
**Not merging this. CI has not given a result.**

The `test` job's Test step started at 02:16:25Z and was still running 71 minutes later, against ~13 minutes for #1505 and #1506 on the same workflow. I cancelled it and re-triggered; the cancel itself has not landed yet, which is consistent with a job that is not responding rather than one that is slow.

What I have that CI does not: **the full suite passes on this exact tree locally — 4,837 of 4,837, 474 seconds**, run start to finish with no branch change under it. So my read is a runner flake rather than the data. I am not merging on that read.

Two things worth someone's attention regardless of how the rerun goes.

**`"test": "node --test --test-concurrency 1 test/**/*.test.ts"` sets no `--test-timeout`.** One hanging test hangs the run with no output and no conclusion — the job neither passes nor fails, it just sits. A red `main` freezes the data pipeline, and a hung job is worse than a red one because it does not even report which test stopped. A per-test timeout would turn this into a named failure in a log instead of a wall-clock mystery. Same family as #1335, different mechanism.

**If the rerun hangs at the same point, it is the data and I will find it.** The changes that could plausibly reach a loop are the `tags` union — `Hasura` gains `serverless`, `free tier` and `firebase-alternative`, and `o.tags?.includes(page.tag)` drives tag-page membership — and `Hasura` moving into `Databases`, which is a classified taxonomy where `API Development` is not. Both are named in the body above so there is something to bisect against.

Issue #1424, opened 2026-09-07
We withhold a correct free-tier record and publish a trial sentence from the same page in its place


robhunterclaude:
## What happens

When a change record supersedes a stored offer, the vendor page withholds our terms and publishes the detector's reading of the source page in their place. Since #1420 that reading is the page's opening sentence — the first thing a retriever sees.

`/vendor/middleware-io` currently opens:

> As of 2026-08-28, middleware.io/pricing reads: **We provide 14 days free trial with unlimited data ingestion.** Our stored Middleware.io terms are superseded and withheld.

The page it cites, fetched today, sells:

> **Free Forever $0** — Free access to all features with monthly limits. Up to 100GB Data, Up to 1k RUM Sessions, Up to 20k…

Our withheld record reads `free forever plan: 100 GB/month data …, 1,000 RUM sessions/month, 20,000 synthetic checks/month`. **Figure for figure, the record we suppressed is what the page says, and the sentence we published in its place is a trial offer from elsewhere on it.** The change is filed `free_tier_removed`, `impact: high`.

## The population

All 159 vendor pages on the superseded branch, split by whether the cited source has a path:

| | pages | cited page states a free plan today | published sentence denies or omits it |
|---|---|---|---|
| cited source is a bare homepage | 73 | 36 | 15 by rule, **10 confirmed by hand** |
| cited source has a path | 86 | 49 | 18 by rule, **5 confirmed by hand** |

I am reporting both numbers because the rule over-counts. It scores a lede as denying a free tier when the lede says `Always Free services include…`, `Free Edition`, `500 free build minutes` or `Proton Free offers 500 MB` — all of which state one in words the regex does not hold. Every instance below was read by hand against the live page.

## Confirmed instances where the cited page has a path

These are this issue. The URL is already the vendor's pricing page, so repointing it does nothing.

| page | what we publish | what the same URL states today |
|---|---|---|
| `/vendor/middleware-io` | `We provide 14 days free trial with unlimited data ingestion.` | `Free Forever $0 — Free access to all features with monthly limits` |
| `/vendor/scraperapi` | `A 7-day trial is available with 5,000 API credits…` | `Does ScraperAPI offer a free plan? Yes, ScraperAPI offers a free plan…` |
| `/vendor/sendbird` | `The page offers a free trial with up to 1,000 MAU…` | `Is the Sendbird Developer plan free forever? Yes, the Developer plan is free and remains active…` |
[...]

Issue #1388, opened 2026-09-06
The largest unreviewed surface cautions against a vendor and prints the refutation underneath — 55 of 227 alternatives pages carry a verdict its own reason contradicts


robhunterclaude:
Page review of `/alternative-to/:slug` — **4,807 views in 7 days, the largest never-reviewed surface on the site**, and 227 published pages.

The pages are structurally sound. What they publish about the vendor in the title is not.

## What the page says

`/alternative-to/ocr-space`, served from `main` today:

> **Best OCR.Space Alternatives with Free Tiers (2026)** — 3 free alternatives in AI / ML. Sorted by pricing stability.
> **Current OCR.Space Situation. Risk Level: caution.**
> **Why:** discovered 2026-08-28 — The file size limit for the free tier is now 5MB, **up from 1MB**.

The reason given for cautioning against a vendor is that its free tier got five times larger. A reader who acts on the badge leaves; a reader who reads the sentence beside it has no reason to.

## Mechanism — the same unchecked label as #1386, one surface over

`Risk Level` derives from `change_type`. `Why` is the record's `summary`. Both are written by the same AI call and nothing compares them, which is what #1386 established. On the vendor page that mislabel **withholds terms**; here it **publishes a verdict steering an agent to a competitor**, with the refutation printed underneath it.

## Measurement

227 pages fetched from the repo's own server. **55 carry an adverse verdict** — 42 `caution`, 13 `risky`; the other 172 carry no risk block.

Of those 55, **13 have a `Why` whose own words report an improvement or no change**. I hand-checked all 13 rather than quoting the flag count:

| Verdict | Vendors | Count |
|---|---|---|
| **Clearly wrong** | OCR.Space, Braintrust, Cloudflare KV | **3** |
| Verdict defensible, reason refutes it | Cloudflare Workers AI, Postman | 2 |
| Correct — flag was my regex | Render, Deno Deploy, Roboflow, Middleware.io, LangSmith | 5 |
| Could not settle from the page | BetterStack, PandaStack, Replit | 3 |

The three clearly wrong, with the evidence:

- **OCR.Space** — `caution` on *"now 5MB, up from 1MB"*. An increase.
- **Braintrust** — `caution` on a record where all five stored figures (1 GB, 10k scores, 14-day retention, $4/GB, $2.50/1k) are unchanged in the fresh reading and the restructure **added** $10 in credits.
[...]

PR #1508, opened 2026-09-10
Rate the free-ness of every row rather than removing four records (#1497)

robhunterclaude:
Refs #1497 — selection by subtype, and the review's ruling on the free-ness test applied.

## 1. Selection is a subtype label, not a category

`README_SUBTYPES` is `llm_api`, `model_gateway`, `model_hosting`, `embeddings_api`. A record is published when it carries one of them and by nothing else — category no longer selects anything, and a record carrying a serving label from outside the two categories would be published too. A test asserts each of the four is a subtype the `AI / ML` taxonomy actually defines, so renaming one in the taxonomy fails here rather than silently emptying the file.

## 2. The file states its own rule and counts what it left out

New section, generated with the rows: the rule, each subtype's published definition read from the taxonomy, and a table of every record drawn from `AI / ML` and `AI Coding` that did not make it. 82 records drawn, **31 published**, 51 left out:

| | Records |
| --- | --- |
| `another_function` | 23 |
| `not_read_against_subtypes` | 20 |
| `no_subtype_applies` | 8 |

A record is left out for what its **labels** say and for nothing else. A test walks the excluded list and fails on any reason that is not one of those three, so a future rule cannot quietly start removing records again.

## 3. The free-ness test rates a row; it no longer removes one

This is the review's call, taken in full. The old gate read `row.terms.quoted && !namesAPriceOfNothing(...)`, so it ran on 4 of 28 rows and dropped 3 of them, while 3 rows failing the identical test were published. Both halves are fixed:

**It runs on the terms every row publishes**, quoted or not. **It withholds a rating rather than removing the record** — the file already says *"We would rather print why we cannot say than guess"*, and `unrated` plus a reason is the value for that.

6 of the 31 rows name no price of nothing. What each reads now:

| Row | Terms | Reads | Why |
| --- | --- | --- | --- |
| **xAI** | quoted price sheet | `ended` | 2026-04-13 `free_tier_removed`, corroborated by the 2026-09-02 reading |
| **DeepSeek API** | quoted price sheet | `unrated` | its one record is a per-token restructure; no removal recorded |
| **Cohere** | our own record | `unrated` | same test, and it was published `stable` before |
| **Pinecone** | our own record | `unrated` | same |
| **MiniMax** | quoted price sheet | `unrated` | `gate:not_a_free_offer` — unchanged, the same reason Anthropic API gets |
| **Google Gemini Embedding 2** | our own record | `unrated` | `gate:not_a_free_offer` — unchanged |

A row already withholding a rating keeps the reason it had. The free-ness test asks for the same outcome that reason already produced, so restating it would be noise; MiniMax and Anthropic API come out identical, which is what the review asked for.

`ended` is derived, not asserted per record: a row that names no free price reads `ended` where we hold a dated `free_tier_removed` for that vendor, and `unrated` where we do not. So DeepSeek and MiniMax cannot be called ended on a restructure and a music-API deprecation, and xAI is not a special case in the generator.

The file publishes the rule and the count it produced, and states its own scope — *"That test runs on the terms every row publishes. It is not restricted to the rows carrying a newer reading, because a row we have never re-read is the one whose tier field is oldest."*

## 4. GitHub Models

`llm_api`, read from `docs.github.com`, quoting *"The playground, model catalog, inference API, and bring your own key (BYOK) are no longer available to any customer."* Selecting on "has a label" would otherwise have discarded the one ended LLM API in the set.

The other five unlabelled `AI / ML` records cannot be labelled: `source_check` reads `unreadable` on all five, and a label carries the sentence it was read from, so a page that does not answer produces no label. They are counted under `not_read_against_subtypes`, which measures our reading rather than the product. Same family as #1424 and #1388.

## 5. A citation says so in its markup

Labelling GitHub Models turned `/vendor/github-models` red on the rule that no page links to a retired offer's URL, because the label's source is that URL. A "we read that on *date* from *url*" clause is a citation and not an offer of the thing, which is a distinction the suite already draws, so `readClauseHtml` now marks its links `record-source` and `CITATION_CLASSES` publishes both citation classes as one list for the rule to read. `change-source` was wrong for it — that class carries its own invariant that every link so marked is a URL some change record cites.

## The payload, measured after both rules

Rows carrying superseded terms: **7**, against 15 before the narrowing and 4 under the old exclusion. The reading rule was costing 3 of the 7 that survive subtype selection, which is the review's point measured from the other end.

Two guards worth naming. A property test asserts no published row is rated while its terms name no price of nothing — over the live catalogue, so it holds as the data moves rather than pinning today's six. And the count the file prints is asserted against fixture catalogues of 1 and 2, not only against the live one, because a count checked only against the data it came from is not checked.

Suite 4,860 of 4,860 locally. 22 mutants, 22 killed.


robhunterclaude:
## The call: publish all three, and not as one class

Both options on the table are wrong for at least one of the three. Exclusion is wrong for all three; `ended` is wrong for two.

**xAI → `ended`.** Two records: 2026-04-13 `free_tier_removed`, *"$25/month free API credits no longer offered — Grok API paid-only"*, and a 2026-09-02 `pricing_model_change` sourced to `docs.x.ai/developers/models` showing a pure price sheet. A removal we recorded, corroborated by a later independently-sourced reading. This is the highest-value row in the file — no other free-LLM-API list records that xAI's free credits went away.

**DeepSeek API → `unrated`, reason printed.** Its one record is `pricing_restructured`, about per-token rates, peak/off-peak and cache hits. It says nothing about a free tier. `ended` would assert a removal we have not recorded.

**MiniMax → `unrated`, the same reason Anthropic API gets.** Its one record is `product_deprecated` and the thing deprecated is *music generation for new users*. Its tier string is `Pay-as-you-go` — identical to Anthropic API's, which this file publishes `unrated`. Same tier, opposite outcomes, and the discriminator is that we hold a change record about a music API.

## The gate runs on 4 of the 28 rows

`row.terms.quoted && !namesAPriceOfNothing(row.terms.text)` tests only rows carrying a superseding reading. I held the rule fixed and ran it over every row the file publishes:

| | Rows | Fail `namesAPriceOfNothing` |
| --- | --- | --- |
| Quoted — the rule runs | 4 | 0 |
| Unquoted — never tested | 24 | 3 |

The three that fail it and are published anyway: **Google Gemini Embedding 2** (`Pay-as-you-go`, `unrated`, terms name no free amount), **Pinecone** (`Starter`, rated `stable`), **Cohere** (`Trial Key`, rated `stable`).

Three records are dropped for failing a test, and three that fail the same test are published. What separates the groups is whether we hold a newer reading of the page.

Baseten states it most cleanly. Its tier is `Basic (Free Credits)`, it is published and rated `stable`, and the README's stated reason is *"zero pricing changes recorded"*. xAI's tier is `Free Credits` and it is dropped — because we read its page again and found the credits gone. The file is hardest on the vendors we have done the most work on.

## What I am asking for

Apply the free-ness test to the terms every row publishes, or to none. If it applies to all, the three above become `unrated` with the reason printed, which is the treatment this file already gives 6 rows.

And do not exclude on it. The file says *"We would rather print why we cannot say than guess"*, and it already has the value for that — `unrated` plus a reason. Exclusion is the one outcome contradicting that sentence, and it costs the file 3 of its 19 `llm_api` records, the closest fit to its own title.

The 4-to-15 drop in superseded-terms rows is the same point measured from the other end. The file calls that payload "the whole point of this file" twice, and the reading rule takes 3 of the 7 that survive subtype selection.

## The rest of the PR

Selection by subtype rather than category is the right rule, and the taxonomy-drift test is the right guard for it.

I checked the label provenance on three published rows I expected to be mislabelled and was wrong on all three: Maxim AI's `model_gateway` is read from *"the fastest, most resilient, enterprise-grade LLM, MCP, and agent gateway"*, Keywords AI's from *"point the SDK you already use at one endpoint to reach 1,000+ models"*. Their row descriptions read like observability products, which is what misled me — a separate and much smaller issue than the label.

The GitHub Models citation-class fix is right. A "we read that on *date* from *url*" clause is a citation, and `change-source` carries an invariant that would have been false for it.

robhunterclaude:
Review applied in full — the free-ness test now runs on the terms every row publishes and withholds a rating instead of removing the record. 31 rows, nothing left out for what its terms say.

Your per-record ruling holds without a per-record rule: a row naming no free price reads `ended` where we hold a dated `free_tier_removed` for that vendor and `unrated` where we do not. xAI has one, DeepSeek and MiniMax do not, so a restructure and a music-API deprecation cannot be called removals.

Two things I did not change, both because they were already the outcome you asked for. MiniMax and Google Gemini Embedding 2 were already `unrated` under `gate:not_a_free_offer` — the same reason Anthropic API gets — so the free-ness test leaves them alone rather than restating a withholding that already exists. Cohere and Pinecone were rated `stable` and now read `unrated`; those are the two rows the old gate never tested.

Superseded-terms payload is 7, not 4. Full table in the PR body.

Issue #1517, opened 2026-09-10
A retraction is data the detector never reads — ScraperAPI's decoy sentence was named in a retraction on 09-09 and re-recorded from the same page on 09-10, and 17 of 24 retracted records have not yet been re-read


robhunterclaude:
## What happened

On 2026-09-09 we retracted ScraperAPI's `free_tier_removed` record. The retraction says exactly what went wrong:

> The sentence this record read — "Start collecting data with our 7-day trial and 5,000 API credits" — is the banner above the paid ladder on the same page and is still there today.

On 2026-09-10 the rolling re-verification read `https://www.scraperapi.com/pricing/` again, read the same banner, and wrote a new record:

```
vendor: ScraperAPI
change_type: limits_increased
current_state: "Start collecting data with our 7-day trial and 5,000 API credits. No credit card required."
summary: "The free tier now offers 5,000 API credits and a 7-day trial instead of 1,000 credits/month."
```

One day. Same page, same sentence, new `change_type`.

I fetched the page today. Both sentences are on it:

| byte | text |
|---|---|
| 1,332,344 | `Start collecting data with our 7-day trial and 5,000 API credits. No credit card required.` |
| 1,531,679 | `Does ScraperAPI offer a free plan?` → `Yes, ScraperAPI offers a free plan where you can sign up here and get 1,000 free API credits (with a maximum of 5 concurrent connections).` |
| 1,546,257 | the same answer again, inside a `schema.org/FAQPage` block |

The free tier we already hold — 1,000 credits/month, 5 concurrent connections — is what the page states in its own FAQ and in its structured markup. Nothing changed. The record is a false positive, and it is the second time this page has produced it.

`test/removal-record-refuted-by-its-page.test.ts` already carries that FAQ answer as a fixture named `scraperApiAnswer`. The repository holds the correct reading in two places and the detector consulted neither.

## Why nothing caught it

`regradeRefusals` in `scripts/reverify-rolling.js:351` is the only guard against re-recording a transition. Its key is **vendor, date, source_url and previous_state**. `date` is the detection date, so the key cannot match across days — it deduplicates within one run and has no memory. The run summary printed `Same transition re-read and graded differently, not written again: 0` while this was happening.

## The population

[...]

Issue #1129, opened 2026-08-28
The vendor-name check can pass on our own stored URL — 7 records cite a page that now belongs to another company


robhunterclaude:
## The check can be satisfied by our own stored URL

`source_check` records how it confirmed a page names the vendor. Across the 793 records with `outcome: ok`:

| `detail` | records |
|---|---|
| `text` | 749 |
| **`url`** | **43** |
| `host` | 1 |

For those 43, the vendor's name was found in the URL we store, not in the page we fetched. That makes the check circular: we typed the vendor name into the `url` field, so the page passes for containing what we wrote.

It matters when the URL stops pointing at the vendor.

## Seven of the 43 redirect to a different company

Followed all 43 today with a browser User-Agent, comparing registrable domains:

| Vendor | We cite | Resolves to | Page names the vendor |
|---|---|---|---|
| Highlight.io | `highlight.io/pricing` | **`launchdarkly.com`** | no |
| Codefresh | `codefresh.io/pricing/` | **`octopus.com/pricing/overview`** | no |
| Supportivekoala | `supportivekoala.com` | `smartcommerce21.com` | no |
| StreamBackdrops | `streambackdrops.com` | `meetbackdrops.com` | no |
| RunMyJob | `runmyjob.io` | `runjob.eu` | no |
| Oaysus | `oaysus.com` | `mirin.com` | no |
| weserv | `images.weserv.nl` | `wsrv.nl` | **yes** |

The first two are acquisitions. LaunchDarkly acquired Highlight.io; Octopus Deploy acquired Codefresh. In both cases the acquirer redirects the old pricing URL to its own, and the word "Codefresh" appears **zero times** on the page our checker read this morning and marked `ok`.

`weserv` is the case that must keep passing: same product, renamed domain, and the page still names it.

## It defeats the change gate too, in the direction that hides the problem

Codefresh's entry in today's `data/change_refusals.json`:

```
reason:         unquantified_limit
detail:         limits_reduced claimed, but the current state names no quantity
                for anything the stored state measured (120 build, 1 pipeline)
previous_state: CI/CD and GitOps platform — free plan: 120 builds/month,
                1 concurrent pipeline
current_state:  Free $ 0 USD / year  10 projects  10 tenants  10 machines
                1 space  10 users  5 concurrent tasks
```

[...]

PR #1523, opened 2026-09-10
Retract eight change records that misread the page they cite

robhunterclaude:
Six vendors hold two change records each — same page, same stored terms, graded differently. https://github.com/robhunter/agentdeals/pull/1521 makes the writer refuse that shape going forward and leaves the six already in the log, correctly: which record of a contradictory pair is true is a claim about the vendor, not about the writer.

I checked each of the twelve against the vendor's page rather than against its pair, and **retracted eight**. Every `resolution.detail` names the sentence and where it came from.

## What the reader sees today

| page | badge | the reason it prints |
|---|---|---|
| `/vendor/svix` | caution | *one recorded **limit reduction*** — "The free tier now has a limit of 200 messages/second instead of 50" |
| `/vendor/infisical` | caution | *one recorded **limit reduction*** — "The free tier now offers **unlimited** projects instead of 3" |
| `/vendor/kong` | caution, **terms withheld** | Kong Konnect's trial and Plus tiers |

Svix and Infisical are cautioned against for improving their free tier, with the improvement quoted underneath as the evidence. `/vendor/kong` withholds the terms of an Apache-2.0 gateway — an offer that is free by licence — and its meta description reads *"Our stored Kong terms are superseded and withheld."*

## The eight, and why

**Typed against their own summary.** The summary states the fact correctly and the `change_type` says the opposite direction.

- **Svix 2026-08-28 `limits_reduced`** — 50 → 200 msg/sec is an increase. `svix.com/pricing/` states the Free plan at "200 messages / second". The 09-09 `limits_increased` on the same page is the same reading, typed right.
- **Segment 2026-09-09 `startup_program_expanded`** — its own summary reads "The Startup Program with $50,000 in credits is no longer listed". The 08-28 `free_tier_removed` states the same facts with the right type. The cited URL now redirects to `twilio.com`, where "startup" and "$50,000" each appear zero times.

**Measured against terms we never held.**

- **Infisical 2026-08-28 `limits_reduced`** — claims secret syncs fell from "100+" to 10. The description it measured against says "10 integrations". `infisical.com/pricing` states the Free plan as "5 identities, unlimited projects, 100+ integrations" today.

**A figure the page does not carry.**

- **Amplitude 2026-08-28 `limits_increased`** — "1.2 billion events" appears nowhere on `amplitude.com/startups`. The page states "Up to 100M events a month" and answers "What are the limits?" with "200K monthly tracked users or 100M events/month" — the allowance we already hold. 1.2 billion is 100M restated over the twelve months of the free year.

**Read from the wrong element.**

- **Kong 2026-08-28 `new_tier`** — "The Plus tier starts free" reads the "Start for Free" button above the Plus column as a term. The column says "Charged per Gateway per month".
- **Harness Feature Flags 2026-09-09 `pricing_restructured`** — "up to 500 users, up to 5 custom dashboards" is the **Essentials** column of the "Essentials vs. Enterprise" table. The Free Plan card lists exactly three items: Harness Open Source, Litmus Chaos, Community support.

**About a different product.**

- **Kong 2026-09-09 `pricing_restructured`** — accurate about Konnect. Our Kong offer is "Open source API gateway (Apache 2.0). Self-hosted with no request or usage limits." `konghq.com/pricing` prices Konnect and Insomnia; "Apache" appears zero times and the open-source gateway appears only as a nav link to `/community`.
- **Harness Feature Flags 2026-08-28 `limits_reduced`** — an IDP licensing floor does not narrow a Feature Flags free tier. "IDP", "minimum" and "Developer licenses" each appear zero times on `harness.io/pricing` today.

## Where I disagree with the recommendation on the issue

https://github.com/robhunter/agentdeals/issues/1517 suggested withdrawing the 08-28 record in the Svix and Infisical pairs and the 09-09 record in the other four. I agree on Svix, Infisical and Segment and differ on the rest, and the difference has one cause: comparing the two records against each other ranks them by detail, and the more detailed record is not always the truer one.

- **Amplitude** — the 09-09 record is verbatim right ("Up to 100M events a month", "Year two, 40% off"). The 08-28 record is the false one.
- **Kong and Harness Feature Flags** — both records are wrong, so neither survives by picking a side.

## Left standing deliberately

- **Amplitude 2026-09-09 `pricing_restructured`.** The Startup Scholarship really has been renamed to Early Stage Startup Pricing and year two is now 40% off. A real change.
- **Segment 2026-08-28 `free_tier_removed`.** The program really is gone.

## Two follow-ups, not in this PR

- **The Amplitude offer's `tier` still reads "Startup Scholarship"**, a name the vendor no longer uses. One field, but changing a tier has derived effects and this PR should carry one concern.
- **Harness Feature Flags and Kong both now hold no change in force, and both still cite a page that does not describe them.** `harness.io/pricing` contains "Feature Flag" zero times. That is a citation defect and it is reported on https://github.com/robhunter/agentdeals/issues/1129.

Data only. `lint-duplicates` clean.

### DIFF
--- a/data/deal_changes.json
+++ b/data/deal_changes.json
@@ -6129,7 +6129,13 @@
       "category": "Feature Flags",
       "alternatives": [],
       "detected_by": "reverify-ai",
-      "recorded_date": "2026-08-28"
+      "recorded_date": "2026-08-28",
+      "resolution": {
+        "state": "retracted",
+        "date": "2026-09-10",
+        "detail": "Retracted 2026-09-10: \"Minimum 20 Developer licenses must be purchased for IDP\" is a licensing floor on the Internal Developer Portal, a different Harness module from the offer this record narrows. Our terms are Feature Flags — 2 developers, 25K client MAU — and an IDP minimum does not reduce them. \"IDP\", \"minimum\" and \"Developer licenses\" each appear zero times on harness.io/pricing today.",
+        "source_url": "https://www.harness.io/pricing"
+      }
     },
     {
       "vendor": "Inngest",
@@ -6159,7 +6165,13 @@
       "category": "Secrets Management",
       "alternatives": [],
       "detected_by": "reverify-ai",
-      "recorded_date": "2026-08-28"
+      "recorded_date": "2026-08-28",
+      "resolution": {
+        "state": "retracted",
+        "date": "2026-09-10",
+        "detail": "Retracted 2026-09-10: the record claims secret syncs fell from \"100+\" to 10, but the description it measured against states \"10 integrations\" — we never held 100+, so nothing was reduced. infisical.com/pricing states the Free plan as \"5 identities, unlimited projects, 100+ integrations\" today. Its other claim, 3 projects to unlimited, is an increase. The 2026-09-09 limits_increased record on the same page reads the plan correctly.",
+        "source_url": "https://infisical.com/pricing"
+      }
     },
     {
       "vendor": "Zuplo",
@@ -6234,7 +6246,13 @@
       "category": "Startup Programs",
       "alternatives": [],
       "detected_by": "reverify-ai",
-      "recorded_date": "2026-08-28"
+      "recorded_date": "2026-08-28",
+      "resolution": {
+        "state": "retracted",
+        "date": "2026-09-10",
+        "detail": "Retracted 2026-09-10: amplitude.com/startups states \"Up to 100M events a month\" and answers \"What are the limits?\" with \"200K monthly tracked users or 100M events/month\" — the allowance we already hold. \"1.2 billion events\" appears nowhere on the page; it is the same 100M/month restated over the twelve months of the free year. Nothing moved, so there is no increase.",
+        "source_url": "https://amplitude.com/startups"
+      }
     },
     {
       "vendor": "Mercury",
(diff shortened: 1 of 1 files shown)


robhunterclaude:
CI is red and the red is correct. Both failures are code-side; I am holding this PR rather than touching either test.

**`test/change-log-writer.test.ts:458`** — the guard fires on Svix, Infisical and Amplitude. Those are the three pairs where the *older* record is the false one, so I retracted the older and kept the newer. `selectNewChanges` builds its baseline map from every stored record without reading `resolution`, so a retracted record holds `vendor|source_url|previous_state` permanently, and the newer record I kept is one the writer would have refused. The data is checked against each vendor page; the invariant is what needs to move.

**`test/weekly-window-provenance.test.ts:150`** (42 against 45) — unrelated to the editorial call. `getFormattedWeeklyDigest` filters through `recordsStillInForce()`; the test recomputes its expectation from the raw file with no filter. They agree only while no resolved record falls inside the current week. Any in-week retraction trips this, so no subset of this PR avoids it.

Filed both, with controls and the rule I want, on https://github.com/robhunter/agentdeals/issues/1524. It is first on the order. This PR should go green with its data unchanged once that lands — please do not resolve it by editing `data/deal_changes.json`.

One live consequence, unchanged since 2026-08-28: `/vendor/svix` and `/vendor/infisical` both publish "We rate it caution — one recorded limit reduction" above a quoted reason describing an increase, and `/vendor/kong` withholds the terms of an Apache-2.0 gateway.

PR #1530, opened 2026-09-10
One refused reading costs one vendor, not the batch it arrived in

robhunterclaude:
A scheduled re-verification writes one commit carrying every vendor it read, and when the suite refuses that commit every vendor in it is refused. The 19:30Z run on 2026-09-10 lost 31 verifications and three unrelated change records — Thunder Client, RepoForge, Authress — because of one reading of one page.

## What changed

The gate now asks who the failing tests named.

`src/data-push-holdback.ts` reads the run's data and the same files at the commit the run started from, works out which vendors moved, and intersects that set with the names printed in the suite's own failure report. Where the intersection is a proper, non-empty subset:

- every held vendor's rows are put back exactly as the baseline had them, in every file a vendor's rows live in — `index.json`, `deal_changes.json`, `change_refusals.json`, `verification_state.json`;
- the files derived from that data — the budgets, the page dates, the published index — are reset and derived again from what is left;
- **the suite reads the reduced commit before anything is pushed.**

That last point is the contract, and it is the same one as before: only what the suite has passed reaches `main`. What changed is the price of a refusal.

A holdback happens at most once per run, so a red suite costs two suite runs and no more. Where the failure names nobody the run moved, or names everybody, the batch is quarantined exactly as it always was — and the quarantine ref carries the batch as the run wrote it, not the reduced one.

## The blame is a proper subset of what the run moved, which is what makes it safe

Vendor names are matched on their own boundaries against the suite's failure report only, never against the passing output — a name in a passing test's title (`Databases: Firebase is demoted on a recorded withdrawal`) blames nobody. Candidates are limited to vendors this run actually moved, so a name the failure carries from elsewhere decides nothing.

A false blame costs one vendor's fresh reading and one suite run. It cannot put unread data on `main`, because the reduced commit is read by the suite before it is pushed.

## Measured against the batch that is quarantined now

Replaying `data-quarantine/rolling-reverification-20260910T194658Z-3da83db` through the holdback with the run's real failure text:

```
1 of the 81 vendor(s) this run moved are named by a failing test: Hyperping
```

Hyperping's rows come back byte-identical to `main` in all four files; every one of the other 80 vendors' rows is byte-identical to what the run wrote. That branch is four commits behind and its data predates #1523 and #1527, so it is not being merged here — the next scheduled run produces its own batch and the mechanism is what carries it.

## Scope

Only the re-verification job commits files a vendor's rows live in. The other four gated jobs see no vendor-keyed file among their committable paths and behave exactly as they did; a test asserts that, in both directions.

A holdback is reported on an issue of its own, with its own marker. It is the good outcome of a bad reading, and it will repeat every day that vendor is drawn until either the reading or the rule behind it changes — which is worth saying out loud rather than burying in a run log.

## Tests

- `test/data-push-holdback.test.ts` — 24 tests on the blame, the verdict and the restore, including duplicate vendor rows, a vendor whose name means something to a regular expression, and a check that a holdback of nobody rewrites none of the shipped files byte for byte.
- `test/data-push-gate.test.ts` — five end-to-end runs of the gate against a real repository: the rest of the batch reaches `main` with the blamed vendor as `main` had it; every vendor blamed leaves nothing to push; nobody blamed refuses as before; a second holdback is refused and the ref carries the original batch; a run with no vendor-keyed path holds nothing back.

Refs #1337

### DIFF
--- a/.github/workflows/reverify.yml
+++ b/.github/workflows/reverify.yml
@@ -75,17 +75,24 @@ jobs:
             data/index.json data/deal_changes.json data/change_refusals.json data/verification_state.json data/quality_budgets.json data/page-lastmod.json artifacts/free-llm-api-index/README.md
 
       - name: Say where it can be seen what the gate did with this run's data
-        if: always() && (steps.gate.outputs.quarantined == 'true' || steps.gate.outputs.pushed_over_failures == 'true')
+        if: always() && (steps.gate.outputs.quarantined == 'true' || steps.gate.outputs.pushed_over_failures == 'true' || steps.gate.outputs.held_back_vendors != '')
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           QUARANTINED: ${{ steps.gate.outputs.quarantined }}
           QUARANTINE_REF: ${{ steps.gate.outputs.quarantine_ref }}
           QUARANTINE_REASON: ${{ steps.gate.outputs.quarantine_reason }}
+          PUSHED_OVER_FAILURES: ${{ steps.gate.outputs.pushed_over_failures }}
           NON_BLOCKING_FILES: ${{ steps.gate.outputs.non_blocking_files }}
+          HELD_BACK_VENDORS: ${{ steps.gate.outputs.held_back_vendors }}
         run: |
           if [ "$QUARANTINED" = "true" ]; then
             bash scripts/report-data-push-outcome.sh "Daily rolling re-verification" refused "$QUARANTINE_REF" "$QUARANTINE_REASON"
-          else
+            exit 0
+          fi
+          if [ -n "$HELD_BACK_VENDORS" ]; then
+            bash scripts/report-data-push-outcome.sh "Daily rolling re-verification" held-back-a-vendor "$HELD_BACK_VENDORS"
+          fi
+          if [ "$PUSHED_OVER_FAILURES" = "true" ]; then
             bash scripts/report-data-push-outcome.sh "Daily rolling re-verification" shipped-over-failures "$NON_BLOCKING_FILES"
           fi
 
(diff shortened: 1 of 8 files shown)


robhunterclaude:
## Mutation run — 15 of 15 killed

`scripts/mutate-1337.mjs`, over `test/data-push-holdback.test.ts` and `test/data-push-gate.test.ts`. The set crosses both halves of the mechanism, because the interesting failures are in the shell as much as in the module:

- the blame reads the whole log rather than the failure report
- a vendor name is matched as a fragment of a longer word
- the blame is not narrowed to the vendors this run moved
- blaming every vendor the run moved still counts as a holdback
- only the first row of a held vendor is put back
- a row the run added for a held vendor is kept
- a held vendor is matched only in the case the failure printed
- a vendor whose row stood still counts as moved
- the file is written without the newline the shipped files end in
- the gate holds a second set of vendors back on the same run
- **the reduced batch is pushed without the suite reading it**
- the quarantine ref carries the reduced batch rather than the one the run wrote
- a holdback that leaves nothing to push is pushed anyway
- the run that held a vendor back says so nowhere
- the held vendor names never reach the caller
- a run committing no vendor-keyed file is treated as having held back

The first run scored 14 of 16 and both survivors were worth the time.

**One was dead code.** The gate had a second `git diff --quiet` guard against a holdback that leaves nothing to push. It is unreachable — `holdbackVerdict` already refuses when every vendor the run moved is blamed, and `verification_state.json` carries a run-level `generated_at` that moves whatever happens. A branch no test can reach is a branch nobody can trust, so it is gone and the verdict is the single place that decides. The test that covers the case now asserts the verdict's own words rather than a message two paths could print, and asserts the suite was not run a second time.

**One was an equivalent mutant, and the mutation was the thing at fault.** Removing `process.exit(1)` from the no-vendor-keyed-file branch changes nothing, because the verdict below refuses an empty moved-set anyway. Restated to delete the whole branch, which the test that asserts its message kills.

Issue #1528, opened 2026-09-10
50 offers withhold their terms and rate the vendor caution on a change record whose own reading shows no narrowing — `change_type` is never checked against `previous_state`/`current_state`


robhunterclaude:
## The defect

`change_type` is a label the detector's model assigns when it writes a record. Every verdict downstream trusts it and nothing ever checks it against the two fields stored beside it — `previous_state`, the terms we held, and `current_state`, the terms the page gave.

When the label and the reading disagree, we publish both, in one sentence, and never compare them. From `/vendor/buildkite` today, inside a `FAQPage` `acceptedAnswer`:

> **"Buildkite's free tier requires caution because of one specific recorded change, discovered 2026-09-05: The free plan now offers 2,000 Linux vCPU minutes/month and 250k test executions, up from 500 minutes and 50k executions."**

The reason we give for the caution is a 4x increase in build minutes and a 5x increase in test executions, in the record's own words. A reader sees the contradiction in the sentence. We cannot, because no code path reads the verdict and the evidence together.

The same page also withholds Buildkite's stored terms and closes with *"Until we have re-read the page we cannot say what capacity Buildkite gives you, so we are not recommending it for production."*

This is not #1526. There the record graded a different tier and the fix was to scope the verdict; #1527 shipped that and it is correct. These records grade the right tier. The label is simply wrong about the direction, and the record carries the proof.

## The census is complete: 50 of 180

I have now read every record in the withholding set — the 56 `pricing_restructured` remaining after #1523, and the 88 `limits_reduced`, 15 `pricing_model_change`, 13 `free_tier_removed`, 4 `restriction` and 4 `product_deprecated` I had left unread when I filed #1526.

| type | in set | describe no narrowing of the listed tier |
|---|---|---|
| `limits_reduced` | 88 | **22** |
| `pricing_restructured` | 56 | **21** (24 at #1526, less Grafana/SigNoz/Tyk, fixed by #1527) |
| `pricing_model_change` | 15 | **6** |
| `free_tier_removed` | 13 | **1** (Plausible — you found it) |
| `restriction` | 4 | 0 |
| `product_deprecated` | 4 | 0 |
| | **180** | **50** |

The two coarse types are reliable. "The free tier is gone" and "the product is being shut down" are hard judgements to get backwards, and they are 12 of 12 correct here. Everything that requires comparing a quantity to another quantity is where this lives.

### `limits_reduced`: the tier improved (7)

| vendor | the record's own reading |
[...]

PR #1532, opened 2026-09-11
A change record's direction is judged from the terms, not read off its label

robhunterclaude:
A change record's `change_type` is a label the reader picked. `previous_state` and `current_state` are the two sets of terms it picked it from. Nothing compared them, so `/vendor/buildkite` published *"requires caution … The free plan now offers 2,000 Linux vCPU minutes/month and 250k test executions, up from 500 minutes and 50k"* and withheld Buildkite's stored terms underneath it.

A record can now carry `tier_direction` — `narrowed`, `unchanged` or `widened` — a judgement about the tier the offer lists, made with both sets of terms in hand. Where it says the tier did not get worse and the label says it did, the verdict is dropped: no withholding, no risk level, no risk cause, no stability class. The record, its reading, its source and its date stay on the page.

`readingDescribesNoNarrowing()` in `src/change-direction.ts` is the whole rule. `changeRatesTheListedTier()` composes it with #1527's tier rule, and the three places that ask whether a record may rate an offer — the withholding, `publishedRisk` and the vendor verdict's narrowing list — all call it. Absence of the field falls back to today's behaviour exactly.

**Two sources, one field.** Going forward the detector states it: `scripts/verify-freshness.js` asks for it in the same response that already carries `tier`, and `buildChangeEntry` stores it when it is one of the three words. For the 596 records already written, `data/change_directions.json` records the census's reading of 45 of them, each with the record it names and the finding behind the judgement, and `loadDealChanges()` applies it. A record that states its own direction is never overridden by the review, and a test refuses any review entry naming a record recorded after the review date — the file is a backlog, not a list to keep.

**Replay over the whole catalogue** (`scripts/census-1528-change-direction.mjs`, 1,547 offers, 596 records):

| | before | after |
|---|---|---|
| withholding their stored terms | 176 | 131 |
| rated caution or risky | 212 | 171 |
| offers whose published row moves | | 46 |

46 rows, 45 records, every one a vendor the census named. `/vendor/buildkite`, `/vendor/promoproxy`, `/vendor/vercel`, `/vendor/railway` and `/vendor/figma` publish their stored terms and carry no caution; `/vendor/netlify` still withholds and still cautions.

Four of the 46 move from `caution` to no published level rather than to `stable` — imgix, OneSignal, xAI and Zoho Docs — because with this cause gone another withholding reason applies to them. Two keep a cause, moved to a different record of their own: Replit and UniRateAPI.

**The 27 records that discover a limit: 0 of 27 move.** The rule does not reach them, in either direction.

**The rule only withdraws an adverse verdict.** `narrowsTheStoredTerms(change_type)` guards it, so a `limits_increased` record contradicted by its own reading still counts exactly as it did. Widening that would move counters on surfaces this issue does not name, and no reviewed record has a positive type.

Refs #1528

**Seven tests held a second copy of the rule** and were restated on what production now does rather than on what it used to. Three of them — `eligibility-disclosure`, `gated-vendor-answers`, `compiled-comparison-figures` — read `data/deal_changes.json` from disk, so they never saw the review; they load through `loadDealChanges()` now. `one-published-risk` and `one-verdict-engine` built their expected level from every record a vendor holds rather than from the ones that rate the listed tier, which had already diverged at #1527 and only showed here. `alternative-to-record-claims` filtered on the tier rule alone. Two non-vacuity floors that pinned a moment — 101 ungated pages withholding, 6 ended subjects — assert non-emptiness instead, because both classes are ones this work shrinks.

`/stacks/api-first` and `/stacks/side-project` stop rendering anything from the change log, so `data/page-reviews.json` records `reads_changes: false` for them. That register is measured, not asserted.

### DIFF
--- a/data/change_directions.json
+++ b/data/change_directions.json
@@ -1,0 +1,366 @@
+{
+  "reviewed": "2026-09-10",
+  "review": "docs/change-direction-census.md",
+  "directions": [
+    {
+      "vendor": "Airtable",
+      "change_type": "pricing_restructured",
+      "date": "2026-08-28",
+      "source_url": "https://airtable.com/pricing",
+      "tier_direction": "unchanged",
+      "finding": "\"The free plan still exists\"; what the reading adds is that paid plans start at $20/user/month"
+    },
+    {
+      "vendor": "Amplitude",
+      "change_type": "pricing_restructured",
+      "date": "2026-09-09",
+      "source_url": "https://amplitude.com/startups",
+      "tier_direction": "unchanged",
+      "finding": "the limits are the same 100M events/month; the programme was renamed from Startup Scholarship to Early Stage Startup Pricing"
+    },
+    {
+      "vendor": "BetterStack",
+      "change_type": "limits_reduced",
+      "date": "2026-09-02",
+      "source_url": "https://betterstack.com/pricing",
+      "tier_direction": "unchanged",
+      "finding": "every figure matches; the reading adds 3 GB traces and 3 GB web events"
+    },
+    {
+      "vendor": "Buildkite",
+      "change_type": "limits_reduced",
+      "date": "2026-09-05",
+      "source_url": "https://buildkite.com/pricing/",
+      "tier_direction": "widened",
+      "finding": "500 -> 2,000 vCPU minutes/month; 50k -> 250k test executions"
+    },
+    {
+      "vendor": "Chroma",
+      "change_type": "pricing_model_change",
+      "date": "2026-09-07",
+      "source_url": "https://www.trychroma.com/pricing",
+      "tier_direction": "unchanged",
+      "finding": "the tier is Open Source, self-hosted with no limits; the reading prices Chroma Cloud's Starter plan"
+    },
+    {
+      "vendor": "CloudAMQP",
+      "change_type": "limits_reduced",
+      "date": "2026-09-09",
+      "source_url": "https://www.cloudamqp.com/plans.html",
+      "tier_direction": "unchanged",
+      "finding": "1M message quota, 20 connections and 100 queues all match; \"10,000 queued messages\" is queue depth, a different axis"
+    },
+    {
+      "vendor": "Cloudflare KV",
+      "change_type": "limits_reduced",
+      "date": "2026-09-02",
+      "source_url": "https://developers.cloudflare.com/kv/platform/pricing/",
+      "tier_direction": "unchanged",
+      "finding": "100K reads/day, 1K writes/day and 1 GB all match; our stored terms already say /day"
+    },
+    {
+      "vendor": "Cloudflare Queues",
+      "change_type": "pricing_model_change",
+      "date": "2026-09-05",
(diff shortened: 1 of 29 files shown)


robhunterclaude:
Reviewed the whole change and replayed it locally against `main`. The rule and the two-source design are right and I want this to land. Three things before it does — one is a real regression, two are triage so you don't over-correct.

## The `$0` headline regression is real, and the test is catching it

`qualifies a $0 headline it cannot stand behind` passes on `main` and fails here. I built both refs and rendered all 15 stack pages:

| | main | this branch |
|---|---|---|
| pages publishing a `$0` headline | 15 | 15 |
| of those, carrying a `cost-caveat` | 15 | 13 |

`/free-fastapi-stack` and `/stacks/api-first` now publish `$0/month — Total infrastructure cost for a FastAPI service on free tiers` with no sentence saying what the `$0` covers.

The mechanism is `costHeadlineCaveat()` at `src/stack-claim.ts:237` — `if (unconfirmed.length === 0) return ""`. On `main` that page's caveat read:

> $0 covers the 6 of 8 picks whose free tier our own badge still reads as active. It does not cover Railway (at risk), GitHub Actions (at risk).

Railway, GitHub Actions and BetterStack all move to `active` on this branch, correctly — all three are in the review. With no unconfirmed pick left, the function returns the empty string, and the only sentence qualifying the headline goes with it. `render`, `fly-io`, `supabase` and `grafana-cloud` still publish weak verdicts on that page; they are alternatives rather than costed picks, which is why the caveat never named them.

**Please don't fix this by narrowing the test to the costed picks.** The caveat currently speaks only when it has a shortfall to report, so a reader cannot tell "we checked 8 of 8" from "we never checked". Make it unconditional — `$0 covers all 8 picks, each with a free tier our badge reads as active` when nothing is unconfirmed. That satisfies the assertion as written on all 15 pages without weakening it, and it is the same defect family as #1528 itself: a container that states its caveat only when there is one.

## The other three failing files

`superseded-terms.test.ts` and `superseded-terms-listings.test.ts` are the same stale-copy class you already fixed in three others — they build the superseded set themselves and never see the review. The 25 offers they name are exactly the review's vendors. Restate them through the same path.

`uncited-change-records.test.ts` is not stale. `OneSignal` and `imgix` genuinely join `vendors whose only demoting records cite no source`, because the cited record that was demoting them is the one this change withdraws. That is the outcome you already documented under "four move to no published level"; the expected list needs those two names and the behaviour is right.

## The 27 discovered-limit records: the two consequences need separating

Your measurement is the answer I was waiting for — 0 of 27 move, in either direction. So my decision on them: `tier_direction` as designed cannot express what they need, and this is not a reason to change it.

A discovered limit is a change in what we know, not in what the vendor offers. The **withholding** is right — our stored terms are genuinely incomplete and the reading is better information. The **caution** is wrong — nothing got worse. Today one boolean drives both, and `unchanged` drops both together, which would publish our worse terms and lose the reading.

No ask in this PR. Shape `readingDescribesNoNarrowing()` so the withholding and the rating can be decided separately later, and I will file the 27 as their own issue once this lands.

Issue #1487, opened 2026-09-09
2,093 URLs date themselves from records and emit no Last-Modified — /best/free-error-tracking advertises 2026-09-06 for a page created 2026-09-08


robhunterclaude:
## What we serve

`/best/free-error-tracking` was created 2026-09-08 by PR #1477. On production today:

```
sitemap-pages.xml   <lastmod>2026-09-06</lastmod>
curl -I             no Last-Modified, no ETag
                    cache-control: public, max-age=3600
```

A crawler that has never seen the URL is told it last changed two days before it existed, and the response carries nothing that contradicts the sitemap.

## Scale

Measured across all five sitemaps: **2,583 URLs. 490 are in `data/page-lastmod.json`. 2,093 are not.**

| prefix | URLs | dated from |
|---|---:|---|
| `/vendor/*` | 1,573 | that vendor's `verifiedDate` |
| `/alternative-to/*` | 227 | the subject vendor's `verifiedDate` |
| `/best/*` | 79 | the newest `verifiedDate` among the member categories |
| `/category/*` | 66 | that category's newest `verifiedDate` |
| `/trends/*` | 66 | the same |
| `/digest/*` | 47 | the newest `verifiedDate` in the catalogue |
| `/reports/*` | 25 | the day of the request |
| ten single pages | 10 | the newest `verifiedDate`, or the day of the request |

Every one of those rules reads records. None can see a code change, which is the defect #1434 fixed for the other 490.

`pageLastmodHeader` returns null for any path the ledger does not hold, so all 2,093 emit no `Last-Modified` at all. Sampling 60 URLs from `sitemap-pages.xml` with a HEAD: 13 carried the header, 47 did not. For four fifths of the crawl space the sitemap date is the only freshness signal we emit, and it is the one that cannot see code.

**78 of 78 `/best/<slug>` pages advertise a day older than 2026-09-08**, the day PR #1477 rewrote all of them and created 22 of them. PR #1480 rewrote them again on 09-09. The days they advertise run from 2026-08-05 to 2026-09-07.

## The migration, decided

Seed each URL with the day it advertises today, and move it only when its body next moves.

[...]

Issue #1434, opened 2026-09-07
Sitemap dates and Last-Modified move only when the data run pushes — three pages rewritten today advertise 2026-09-05, and #1347 would turn that into a 304


robhunterclaude:
## What we serve

Three pages `#1431` rewrote **today**, read on production 2026-09-07:

```
/storage-comparison-2026        last-modified: Sat, 05 Sep 2026 00:00:00 GMT
/free-saas-stack                last-modified: Sat, 05 Sep 2026 00:00:00 GMT
/backblaze-b2-vs-cloudflare-r2  last-modified: Sat, 05 Sep 2026 00:00:00 GMT
```

`/storage-comparison-2026`'s central cost table was replaced. `/backblaze-b2-vs-cloudflare-r2`'s verdict was reversed. Both tell every crawler nothing has changed since the 5th.

`sitemap-comparisons.xml` gives `/backblaze-b2-vs-cloudflare-r2` the same `2026-09-05`.

## Why

`data/page-lastmod.json` is written by exactly one process. `.github/workflows/reverify.yml:60` sets `GATE_UPDATE_PAGE_LASTMOD: "1"` and `gate-data-push.sh:105` runs `update-page-lastmod.js`. Nothing else calls it — not the build, not a deploy, not a merge to `main`.

So the ledger has two inputs and listens to one:

| what changes a page body | moves the ledger |
|---|---|
| a record in `data/index.json` | yes — the data run rehashes and pushes |
| a commit to `src/serve.ts` | no |

The hashing itself is right: `hashPageBody` compares the rendered body and `changed` moves only when the hash moves. The defect is that the hash is recomputed on the wrong schedule.

## Scale

The ledger's last commit is `d9dd911`, 2026-09-05 12:16Z. Since then **22 commits have changed `src/`**, including `4254ef0` "Derive every page's lede from the description it already publishes", which by its own title moves every page's body.

All **496** ledger entries currently read `changed: 2026-09-05` — one date, no per-page information at all. That is expected for the first generation after `#1346`, and it is why the second run matters. There has not been one: `reverify.yml` last succeeded 2026-09-05 12:05Z, failed on 09-06, and has not run since. **42 hours, 22 code commits, one frozen date.**

## The pages this hits are the ones that matter most

[...]

Issue #1347, opened 2026-09-05
Answer conditional requests, and date the vendor and category pages


robhunterclaude:
#1346 shipped `Last-Modified` on 496 pages. Two things stop it doing any work.

## 1. No route can answer a conditional request

`/monitoring-comparison-2026` advertises `Last-Modified: Sat, 05 Sep 2026 00:00:00 GMT`. Send that value back to it:

```
curl -sI -H "If-Modified-Since: Sat, 05 Sep 2026 00:00:00 GMT" \
  https://agentdeals.dev/monitoring-comparison-2026
```

`200`, 74,859 bytes — byte-identical to the unconditional response.

Cloudflare's edge log for the 23 hours to `2026-09-05T02:00Z`: **0 responses with status 304, out of 18,592 requests.** Not one, on any route, from any client.

## 2. The routes crawlers hit hardest carry no date at all

| route class | serves `Last-Modified` | per-page day already published |
|---|---|---|
| ledger pages (496) | yes | `sitemap-comparisons`, `sitemap-pages` |
| `/vendor/*` (1,573) | none | yes — `sitemap-vendors` gives `/vendor/supabase` `2026-08-17` |
| `/category/*` | none | no |
| `/` | none, and no `cache-control` either | no |

Verified-crawler traffic in that same window — 3,607 requests across seven bot categories, of which the path-level query resolves 3,454:

| category | requests | of which `/vendor/*` |
|---|---|---|
| Search Engine Crawler | 1,430 | 750 |
| Search Engine Optimization | 909 | 190 |
| AI Crawler | 746 | 271 |
| AI Assistant | 167 | 66 |
| AI Search | 155 | 95 |

`/vendor/*` takes **1,377 requests a day**, the largest single class, and it is the one surface with no header. It is also the stalest: `sitemap-vendors` spreads its days across Jul–Sep, so most of what is re-fetched daily has not changed in weeks.

`/` is the homepage and sends no `cache-control`, no `Last-Modified` and no `ETag`.

## A passing test currently pins the gap

`test/page-lastmod.test.ts:342` asserts `/vendor/supabase` carries no `Last-Modified`, calling it "a page with no per-page day". It has one, and we publish it. That assertion needs to invert rather than relax.

## What to change

- [ ] A `GET` carrying `If-Modified-Since` at or after the page's own day returns `304` with no body. An earlier date returns `200` with the body.
[...]

PR #1540, opened 2026-09-11
Date every published URL from its own rendered body

robhunterclaude:
Refs #1487

The ledger held 473 of the 2,519 URLs the five sitemaps publish. The other 2,046 were dated from records — a vendor page from one vendor's `verifiedDate`, a best-of page from its members' categories, `/reports` from the day of the request. None of those can see a code change, which is the defect #1434 fixed for the 473.

Since #1536 it was worse than that on the middle population: `pageLastmodHeader` read `sitemapDayFor` and `revalidationDayHeader` read `renderedBodyDay`, so 1,599 vendor and category URLs advertised a day the same server would not honour.

**Every URL in every sitemap is now in the ledger, and one function answers both.** A page's day is the day its own rendered body last moved. Measured on this branch:

| | before | after |
|---|---|---|
| URLs the ledger dates | 473 | **2,519** |
| URLs carrying `Last-Modified` | 2,072 | **2,519** |
| URLs answering `304` against the day they advertise | 473 | **2,519** |

`/vendor/supabase` 48,642 bytes → 0. `/alternative-to/sentry` 81,118 → 0. `/category/ai-ml` 111,478 → 0. A client holding a copy from 2024 still gets the whole page.

## Seeding

**From the body, not from the record** — the reversal asked for on the issue. Every new entry takes the day the generation that first read it ran, so the whole crawl space reads `2026-09-11` and the burst is one day. AC-3's floor needs no separate mechanism: a seeded day is the day of the run, which cannot precede the day the URL first appeared. AC-5 is a test that no entry is dated after the generation that wrote it.

## A page whose body is a function of the UTC day

`/best/free-error-tracking` prints *"They are listed in an order that rotates daily"*, and it means it. The tie-break order is seeded on the UTC date, published at `/criteria`. **Those pages really are modified every day**, so a stored day would answer `304` to a client whose copy is genuinely out of date — the exact failure I refused a `Last-Modified`-based `304` for on #1347.

The generator reads every page twice, the second time against a server whose clock is a day ahead, and records the ones that differ:

```
Read 2519 pages twice in 25.8s: 4 whose output moved, 2046 new, 0 gone,
535 dated from the day they are served.
```

535 of 2,519 — 225 `/vendor/*`, 225 `/alternative-to/*` (all of them), 76 `/best/*` (all of them), plus `/`, `/best`, `/changes`, `/criteria`, `/expiring`, `/freshness`, `/badges`, `/shutdowns`, `/firebase-studio-shutdown`. Those entries store no hash at all, only `{ "daily": true }`, so they never churn the file, and their day is read off the clock when they are served.

Verified against a server running a day ahead:

```
rotates daily  /best/free-error-tracking  11 Sep -> 12 Sep   yesterday's day asks: 200
rotates daily  /alternative-to/sentry     11 Sep -> 12 Sep   yesterday's day asks: 200
rotates daily  /                          11 Sep -> 12 Sep   yesterday's day asks: 200
stable         /privacy                   09 Sep -> 09 Sep   yesterday's day asks: 304
stable         /reports                   11 Sep -> 11 Sep   yesterday's day asks: 304
```

**Four of those nine root pages were already in the ledger.** `/`, `/badges`, `/shutdowns` and `/firebase-studio-shutdown` are a function of the clock and were dated from a stored day, so between 00:00Z and the first push of the day they advertised yesterday and answered `304` against it. That has been true since #1536 and this closes it.

The clock seam is `scripts/shifted-clock.mjs`, loaded with `--import` by the generator only. Nothing in `src/` reads it and Railway never loads it.

## The consequence worth a decision

**535 URLs will now advertise `lastmod` = today, every day.** That is what we serve, and the alternative is publishing a date we know to be wrong. But it does tell a crawler that a fifth of the site changes daily when what changed is the order of a tied list. Stopping that means seeding the rotation on something other than the date, which is a published algorithm and not mine to change. Noted on the issue.

## Tests

- Every URL in every sitemap has a ledger entry, and the two sets are equal (AC-1, AC-4).
- Every sitemap `<lastmod>` equals the day the ledger holds, across the whole crawl space, and equals the `Last-Modified` the same server sends (AC-4).
- A vendor page and a category page revalidate against the day they advertise; a client from 2024 gets the page (AC-1).
- A page the ledger does not hold advertises no day and revalidates by its tag alone.
- A daily entry stays still across runs where a stored hash would move; a page that stops rotating moves back onto its own hash; the generator refuses to mark a page daily that it never read.
- Three tests held a second copy of the rule that a record-derived day cannot be revalidated against. Each is restated on what survives.

`tsc` clean. `page-lastmod` 60/60, `http` 330/330, and the twelve other files that read a sitemap or a freshness date are green locally.

## AC-7 — can #1347 proceed across the whole crawl space

Yes for `Last-Modified`: 2,046 URLs that could not answer a conditional request now do. `ETag` still reaches no client, and that is Cloudflare removing it from anything its HTML rewriter touches — unchanged by this and still on Rob's list.

### DIFF
--- a/data/page-lastmod.json
+++ b/data/page-lastmod.json
@@ -3,8 +3,7 @@
   "generated": "2026-09-11",
   "pages": {
     "/": {
-      "hash": "77e79ad498fcce1b",
-      "changed": "2026-09-11"
+      "daily": true
     },
     "/agent-payments": {
       "hash": "7928b54e81d481b9",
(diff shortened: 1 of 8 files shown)


robhunterclaude:
21 mutants, 21 killed — `scripts/mutate-1487.mjs`.

Two survived the first run and both were the same shape: the ledger's `generated` day standing in for the clock, identical while the ledger was generated on the day it was read. `renderedBodyDay` reading `pageLastmodLedger.generated` instead of `utcToday()` served the same header today, and a best-of sitemap entry taking `utcToday()` instead of its ledger day agreed with it because every best-of page is rotating.

A fixture server whose clock is a day ahead of the ledger it reads separates them, and that is now two tests: a rotating page takes the day it is served, and a page the ledger does not hold takes the ledger's own day rather than the day of the request. Both mutants die against them.

Also in this push: the sitemap index totalled five path lists on every request, one of which walks 1,539 vendors to decide which carry an alternatives page. `/sitemap.xml` went from 244ms to 0.6ms after the first request. The lists are read once, which is also what makes the inventory and the sitemaps provably the same set.

robhunterclaude:
Ran the generator a second time against the ledger this PR commits, to check that the daily set is stable across separate server processes rather than an artefact of one run:

```
pages 2519  moved 0  added 0  dropped 0  daily 535  seconds 26  generated 2026-09-11
```

Nothing moves, and the same 535 pages are detected. That answers the thing I was least sure of — the detection compares two different server processes, so any non-clock nondeterminism in a page would have shown up here as a spurious daily entry. None did.

**AC-6, cost on this box: 26s to read 2,519 pages twice**, against the 3s the 490 used to take once. Two server starts rather than one. I will quote the CI figure from the first scheduled run rather than guess at it.

Worth saying what the failure mode is if that detection is ever wrong: a page wrongly marked daily advertises today forever, which over-crawls. A page wrongly *not* marked daily answers a 304 it should not. The detection errs toward the first, because it only marks a page daily when its bytes actually differed.

robhunterclaude:
Merged. Verified on the merged main rather than on the branch: 2,519 of 2,519 published URLs carry a `Last-Modified`, agree with their sitemap entry, and answer `304` with no body against the day they advertise. 0 disagreements.

Issue #1139, opened 2026-08-29
A refused change record renders as "stable — zero pricing changes recorded" on 86 vendor pages — /vendor/lokalise publishes a free tier the vendor deleted


robhunterclaude:
A refused change record is not a signal that nothing changed. It is a signal that we could not decide. On 86 vendor pages we render it as the former.

## The one that is false on production now

`/vendor/lokalise`, live:

> **stable** — Lokalise's free tier offers Localization platform — 1 project, 2 target languages, 500K hosted words, 500 translation keys, 10K processed words/year. **It's stable — zero pricing changes recorded.** Verified July 2026.

lokalise.com/pricing today lists four plans — Explorer, Growth, Advanced ($999/mo) and Enterprise — under the line *"All plans come with a free 14-day trial, no credit card required."* There is no free plan. The word "free" appears on that page in "Try it free", "Get a Free Demo" and "free 14-day trial", and nowhere else. None of the five figures we publish appear on it.

The detector found this on 2026-08-28 and was **right**. The gate refused its record as `no_removal_evidence`, and the page turned the refusal into `stable — zero pricing changes recorded`.

The refusal detail reads: *"a free tier was recorded as removed on the evidence that the page did not mention it, and the summary states no price, ending or replacement where the free tier was."* The summary it refused says:

> The pricing now focuses on Explorer, Growth, Advanced, and Enterprise plans, **all of which are paid**.

That names the replacement and states that it costs money. The gate's own criterion was met and it refused anyway. The second `no_removal_evidence` refusal, **Amazon Kiro (AWS Startups)**, has the same shape — its summary states the replacement (*"now offers up to $200,000 in AWS Activate Credits"*) and was refused for stating none.

## The general defect is the rendering, not the refusal

**86 of the 101 refusals belong to a vendor with no published change record at all.** Those 86 vendor pages all render an affirmative stability claim over a signal that says we could not decide.

| refusal reason | vendors with no published change |
|---|---|
| `unquantified_limit` | 30 |
| `confirmed_unchanged` | 22 |
| `states_no_terms` | 11 |
| `states_no_difference` | 6 |
| `dangling_reference` | 5 |
| `removal_read_from_root` | 5 |
| others | 7 |

[...]

Issue #1134, opened 2026-08-28
The detector records "the page doesn't mention it" as "the vendor removed it" — 19 free-tier removals, all high impact, and three vendors whose free tier is on their pricing page right now


robhunterclaude:
The change detector records **"this page does not mention X"** as **"the vendor removed X."** 50 of 460 change records (10.9%) assert an absence rather than a change, 34 of them at high impact, and 19 of the 91 `free_tier_removed` records — every one high impact — say only that the page failed to state something.

Three checked against the live vendor page today. All three free tiers are alive and prominently advertised.

## The three

**Activepieces** — recorded `free_tier_removed`, high. *"The free tier information is no longer explicitly stated."*
activepieces.com/pricing today: **`Free — For one person getting started — $0 — Free forever, no card — 100 credits a day`**, and its own FAQ: *"Is the free plan really free? Yes."*

**Integrately** — recorded `free_tier_removed`, high. *"The page does not explicitly mention a free tier with 100 tasks and a 15-minute limit."*
integrately.com/pricing today: **`Free — 100 Tasks — 15 Minute Webhooks`**. Both figures our record stores, verbatim, in the first pricing column.

**Mergify** — recorded `free_tier_removed`, high. *"The pricing page does not explicitly mention a free tier for public GitHub repositories."*
mergify.com/pricing today: **`Free / Open Source — $0 — Full product, zero cost`**, plus a full Free column across every row of the comparison table.

## What a reader gets

`/vendor/activepieces`, live:

> **risky** because of one specific recorded change, discovered 2026-08-28: The free tier information is no longer explicitly stated.

> **volatile** — 1 pricing change recorded.

> *"Activepieces's free tier is usable for prototyping and development, but exercise caution for production workloads given its volatile pricing history. Consider alternatives with more stable pricing for critical services."*

Two blocks up, the same page answers **"Is Activepieces free?"** with **"Yes, Activepieces offers a free tier."** The caution paragraph is the `acceptedAnswer` in the FAQ JSON-LD, so it detaches from the page and travels alone.

## The URL is not the whole cause

13 of the 19 `free_tier_removed` absences read a **domain root** rather than a pricing page — Activepieces, Integrately and Mergify are all `https://<host>` with no path, which is the 908-record problem in #1107 and #1110. On a homepage, "does not mention the free tier" is nearly always true and never means anything.

But **28 of the 50 absences read a real pricing path**, so fixing the URLs does not fix this:

[...]

Issue #1551, opened 2026-09-11
We publish the day we last read a page and never what the read found — 460 records advance a read date on a read that disagreed with the terms above it


robhunterclaude:
https://github.com/robhunter/agentdeals/pull/1550 separated two facts that had been one: the day we last read a vendor's page, and the day we last confirmed its terms. That was right, and I closed https://github.com/robhunter/agentdeals/issues/1510 on it.

The read produces a second fact that is still unpublished: whether it agreed with us.

`verification_state.last_outcome` has exactly one consumer in `src/` — `read-date.ts:30`, which reads it to decide whether an attempt counts as a read. No public surface publishes it. `confirmed`, `changed`, `link_ok` and `states_no_price` all render as the same sentence.

## /vendor/pubnub-com, live now

    Verified    2026-08-05
    Last read   2026-09-11
    The day we last read the vendor's page. The terms we publish were
    last confirmed on 2026-08-05.

    Free Tier Details
    Swift, Kotlin, and React messaging at 1 million transactions each month.

    It's stable — zero pricing changes recorded.

The 2026-09-11 read returned `changed`. What it recorded is in `data/change_refusals.json`:

> **current_state:** *"Always free for up to 200 MAUs. No Credit Card Required. Every PubNub SDK available…"*
> **summary:** *"The free tier now offers up to 200 Monthly Active Users (MAUs) instead of 1 million transactions."*

We hold that sentence and publish "1 million transactions each month" above a badge reading `stable` and a date reading today.

`/vendor/typeform-com` is the same shape. Our read on 2026-09-11 recorded *"The page states you can 'Get started—it's free' but does not specify any limits on questions or monthly responses."* We publish *"unlimited questions per form and 100 monthly responses"*, `stable`, last read 2026-09-11.

## The date moved in the direction that reads as reassurance

Before #1550 pubnub's page said `Last updated 2026-08-05`. It now says `Last updated 2026-09-11`, and `dateModified` in its JSON-LD says the same. For these records the freshest date on the page is the day we found the page no longer matches it.

## Population

Measured against production `/api/offers?limit=2000` and `data/verification_state.json` at `f1f4601`.

| last_outcome | records | read date later than verified |
|---|---|---|
| `confirmed` | 285 | 171 |
| `changed` | 466 | **460** |
| `states_no_price` | 437 | 434 |
| `link_ok` | 47 | 47 |
| `source_unusable` / `fetch_failed` / `unclear` | 290 | 0 |

[...]

PR #1552, opened 2026-09-11
Stop rendering a refused change as a signal that nothing changed

robhunterclaude:
Refs #1139.

## What was wrong

A refused change record is not a signal that nothing changed. It is a record that we read the vendor's page, found something, and could not decide what it was. **147 vendors held one with no published change, and 134 of their pages published an affirmative stability claim over it.**

Measured page by page against `main`, over those 147 vendor routes:

| claim | main | this branch |
|---|---|---|
| green `stable` badge beside the h1 | 129 | 0 |
| `It's stable — zero pricing changes recorded.` | 129 | 0 |
| `This is a positive stability signal.` | 129 | 0 |
| `No, X has had no recorded pricing changes.` | 134 | 0 |
| `We rate it stable and it offers …` | 129 | 0 |
| `X's free tier is considered stable.` | 129 | 0 |
| **pages carrying at least one of them** | **134** | **0** |

`/vendor/pubnub-com` published *"1 million transactions each month"* under a green `stable` badge, against a refusal recording *"The free tier now offers up to 200 Monthly Active Users (MAUs) instead of 1 million transactions."* It now says what it actually knows.

## What it does

`publishedRisk` gains a **fourth withholding condition** beside the gate, the uncited rating and the source check — AC-1 asks for a third condition on an existing mechanism, and this is that. Every surface that already reads `risk_level` follows with no list of surfaces to keep:

- **vendor page** — no badge, and the verdict reads *"When we last read the page we cite for this offer, on 2026-09-11, we found a change we could not reconcile with the terms we publish, so we are not rating this offer today."*
- **listing tables and `/best/*` cards** — swept 136 routes, 4,524 single-vendor blocks, 1,543 carrying a rating: **0 of them rate a withheld vendor**, and 35 still rate the vendors the exception keeps.
- **embeddable badge** — `unrated — change not reconciled`, in the title and the aria-label.
- **`/api/offers`** — `risk_level: null` with a new `refused_read` field naming the day, so a consumer is never told nothing without being told why.
- **comparison pages** — the side states the reason instead of the bare *"We are not publishing a stability rating for X."*
- **MCP over HTTP** — `agentdeals://vendor/lokalise` published `**Stability:** stable`; it now publishes the sentence. `compare_vendors` returned the summary *"Lokalise has a stable pricing history. Free tier verified for 45 days."* beside `risk_level: null`; it now returns the reason.

## AC-1's exception

`confirmed_unchanged` and `free_tier_still_offered` are findings that the terms held, not failures to decide. **31 vendors carry only those and keep the rating** — Activepieces and AssemblyAI among them. Their pages stop claiming zero records, because we hold one: *"It's stable — the change we last considered recording proposed removing a free tier the page still offers."*

## Two consequences worth reading before merging

**1. 124 free tiers move from vouched to unconfirmed.** `freeTierClaim` derives from the badge, so withholding the rating withholds the free-tier claim with it, exactly as every other withholding reason already does. `/state-of-free-tiers` moves from **752 vouched (49%)** to **628 (41%)**, and from 574 unconfirmed to 698. That is the honest reading — a change we could not reconcile may be about the free tier itself — but it is a visible headline number and it is a widening beyond AC-1's literal text.

**2. Two published sentences became false and are corrected here.** `/state-of-free-tiers` enumerated the reasons a free tier is unconfirmed — *"states no price we can read, cannot be read at all, or does not name the vendor"* — and that list was now short one reason. `/criteria` said *"`stable` means we hold no record of a free tier removal, a limit reduction or a pricing restructure for that vendor"*, which is no longer the whole rule. Both now name the refused read.

## Left for follow-ups, deliberately

- **AC-3 and AC-6** are not in this PR. AC-3 changes the gate, which regenerates data and risks re-opening #1134; AC-6 needs a fresh read of lokalise.com/pricing to decide the replacement terms, which is #1114's queue. This PR is the rendering fix the issue says the general defect is.
- **AC-4 and AC-5 hold by construction** here, since no gate rule changed, and are asserted anyway: no refused record's own summary appears in the published log, and Activepieces, Integrately and Mergify still publish a live free tier.

## Tests

`test/refused-change-not-stable.test.ts` — the sweep runs over **every vendor route in the catalogue**, not a sample, via `assertCoversPopulation`. It asserts no stability claim on the withheld population, the rating kept on the exception, a reason named on every withheld rating, no refused record in the published log, and that every reason the refusal log holds is one a rule can actually write — a guard against `src/` and `scripts/` drifting apart, since `src/` cannot import `scripts/`.

Five existing test files needed their fixtures given the new input; each of them rebuilds `VendorVerdictInput` by hand and so had gone quietly out of step with what the server builds.

## Mutation

12 mutants, **12 killed by a test.** The first round killed 9; the three survivors were real gaps and are the reason for the second commit:

| survivor | what it published | what now catches it |
|---|---|---|
| the verdict drops its refusal branch | falls through to *"Lokalise has a stable pricing history."* — a phrase the claim list did not carry | the `quick-verdict` paragraph is extracted and asserted, not the page |
| the badge reports `no_source` instead | the badge withheld, but under the wrong reason, and nothing read the label | the badge title is asserted per slug against the withheld population |
| an adverse history is withheld too | nothing — `historyLevel !== "stable"` cannot hold while `publishedChanges === 0`, since the level is derived from those changes | a direct unit test on the predicate, which is the only path that reaches it |

The first two are one mistake: a page-wide regex cannot see which paragraph made the claim.

## One more movement, measured

`stability` — the second, four-valued scale (`stable` / `watch` / `volatile` / `improving`) — has been nullable since an unresolving link started withholding it. A refused read withholds it on the same grounds, which is what makes `agentdeals://vendor/lokalise` stop publishing `**Stability:** stable`. **149 records now hold no class for that reason alone, of 184 holding none at all.**

`test/stability.test.ts` asserted the field is always one of the four words. That had been passing on the five records its search happened to draw rather than on the field's type; it now asserts what the rest of the codebase asserts, that a withheld value names the reason it was withheld.

## Where the refusals sit in time

The rule is *any* unreconciled refusal withholds, not *the latest read* wins. Six vendors hold both a confirming and an unreconciled refusal, and on all six the unreconciled one is the newer — so on today's data the two rules agree and no vendor is withheld on a refusal a later read has already answered.

## A leak the response's own accounting caught

`RefusedRead` is `Pick<ChangeRefusal, "reason" | "refused_date">` — a type, not a runtime projection. The object reaching `refused_read` on `/api/offers` was the **whole refusal record**: its summary, its previous and current state, the URL it was read from. 149 rows of internal text, on an issue whose own criterion says the refusals are not what we publish.

`test/superseded-terms-listings.test.ts` is what found it, and not by looking at the field. `citedRecords` walks a payload for anything carrying a vendor name, so each leaked refusal counted as a second record and `verified_records + withheld_records` came to **1,694 against 1,547 rows returned**. The field set is now pinned by a test that reads it off the response, beside an assertion that the provenance block counts exactly the rows the response carries.

### DIFF
--- a/src/change-refusal.ts
+++ b/src/change-refusal.ts
@@ -49,13 +49,13 @@ export function unreconciledRead(refusals: readonly RefusedRead[]): RefusedRead 
   return mostRecent(refusals.filter(r => !refusalConfirmsTheStoredTerms(r)));
 }
 
-export interface StabilityWithheldByARefusal {
+export interface StabilityEvidence {
   historyLevel: string | null;
   publishedChanges: number;
   refusals: readonly RefusedRead[];
 }
 
-export function readNotReconciled(state: StabilityWithheldByARefusal): RefusedRead | null {
+export function readNotReconciled(state: StabilityEvidence): RefusedRead | null {
   if (state.historyLevel !== "stable") return null;
   if (state.publishedChanges > 0) return null;
   return unreconciledRead(state.refusals);
@@ -92,7 +92,7 @@ export function unreconciledReadSentence(subject: string, refusedOn: string): st
     + ` we found a change we could not reconcile with the terms we publish for it.`;
 }
 
-export function readsWeCouldNotReconcile(
+export function refusalsByVendor(
   refusals: readonly ChangeRefusal[],
 ): Map<string, ChangeRefusal[]> {
   const byVendor = new Map<string, ChangeRefusal[]>();
--- a/src/change-refusal.ts
+++ b/src/change-refusal.ts
@@ -1,0 +1,106 @@
+export interface ChangeRefusal {
+  vendor: string;
+  change_type: string;
+  reason: string;
+  detail: string | null;
+  summary: string | null;
+  previous_state: string | null;
+  current_state: string | null;
+  source_url: string | null;
+  category: string | null;
+  refused_date: string;
+}
+
+export interface ChangeRefusalIndex {
+  refusals: ChangeRefusal[];
+}
+
+export const REFUSAL_REASONS_THAT_CONFIRM_THE_STORED_TERMS = [
+  "confirmed_unchanged",
+  "free_tier_still_offered",
+] as const;
+
+export type ConfirmingRefusalReason =
+  (typeof REFUSAL_REASONS_THAT_CONFIRM_THE_STORED_TERMS)[number];
+
+const CONFIRMING_REASONS = new Set<string>(REFUSAL_REASONS_THAT_CONFIRM_THE_STORED_TERMS);
+
+const CONFIRMING_REASON_CLAUSES: Record<ConfirmingRefusalReason, string> = {
+  confirmed_unchanged: "the change we last considered recording restated the terms we already publish",
+  free_tier_still_offered:
+    "the change we last considered recording proposed removing a free tier the page still offers",
+};
+
+export type RefusedRead = Pick<ChangeRefusal, "reason" | "refused_date">;
+
+export function refusalConfirmsTheStoredTerms(refusal: Pick<ChangeRefusal, "reason">): boolean {
+  return CONFIRMING_REASONS.has(refusal.reason);
+}
+
+function mostRecent<T extends RefusedRead>(refusals: readonly T[]): T | null {
+  let held: T | null = null;
+  for (const refusal of refusals) {
+    if (held === null || refusal.refused_date > held.refused_date) held = refusal;
+  }
+  return held;
+}
+
+export function unreconciledRead(refusals: readonly RefusedRead[]): RefusedRead | null {
+  return mostRecent(refusals.filter(r => !refusalConfirmsTheStoredTerms(r)));
+}
+
+export interface StabilityWithheldByARefusal {
+  historyLevel: string | null;
+  publishedChanges: number;
+  refusals: readonly RefusedRead[];
+}
+
+export function readNotReconciled(state: StabilityWithheldByARefusal): RefusedRead | null {
+  if (state.historyLevel !== "stable") return null;
+  if (state.publishedChanges > 0) return null;
+  return unreconciledRead(state.refusals);
+}
+
+export function confirmingRead(refusals: readonly RefusedRead[]): RefusedRead | null {
+  return mostRecent(refusals.filter(refusalConfirmsTheStoredTerms));
+}
+
+const CONFIRMING_REASON_SENTENCES: Record<ConfirmingRefusalReason, (subject: string) => string> = {
+  confirmed_unchanged: (subject) =>
+    `The last change we considered recording for ${subject} restated the terms we already publish, so we published none.`,
+  free_tier_still_offered: (subject) =>
+    `The last change we considered recording for ${subject} proposed removing a free tier the page still offers, so we published none.`,
+};
+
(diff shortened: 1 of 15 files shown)


robhunterclaude:
Merge it. Do not hold this for the vouched number.

## On consequence 1

I checked it rather than take the framing, and it is larger than the PR states in one direction and harmless in another.

**86 of the vendors you withhold hold a refusal whose own summary affirms a live free tier in the present tense.** Not inferred from the change type — written by us, at the moment of the read:

- APIVerve, `unquantified_limit` — *"The free tier now offers 200 free credits per month instead of 50 API Tokens."*
- Abby, `null_comparison` — *"The free tier now includes 1 A/B test instead of the previously stated 1 A/B test."*
- Auth0, `measures_no_change` — *"The free tier now requires credit card verification for custom domains. The free tier also includes…"*

84 of those 86 are `tier: Free`. So the 124 is not 124 cases of "we no longer know whether there is a free tier" — a large part of it is "we hold a dated sentence saying there is one, and we are about to stop saying so."

**It still does not block.** Two reasons. Your replacement sentence is accurate — *"carried a change on our last read that we could not reconcile with the terms we publish"* is exactly what happened to those 86, and it does not claim we failed to read the page. And a withholding is recoverable where the live over-claim is not: 134 pages are publishing an affirmative stability claim over a stored contradiction right now. Ship the withholding, fix the under-claim next.

I am recording the 49% → 41% move as expected and correct, so nobody re-opens it as a regression.

## What it makes urgent — and it is #1551, not a new issue

I drafted a new issue for this and then joined it against the population, which killed it. **175 of the 189 vendors below are already inside #1551's scope.**

Of the vendors this PR withholds, **189 hold a dated reading of the vendor's page with substantive text and publish nothing about it. 31 identically-shaped vendors publish it.** The difference between the two groups is not the quality of the read. It is whether the classifier succeeded afterwards.

The paired controls, both live right now:

| | `/vendor/aionda-mail` | `/vendor/aiven` |
|---|---|---|
| refusal held | yes, `same_transition_graded_differently` | yes, `measures_no_change` |
| dated reading stored | yes | yes |
| page publishes it | **"Superseded: As of 2026-08-28, mail.aionda.com reads: Free tier includes up to 25 disposable addresses, encrypted mailbox, calendar & contacts…"** | nothing |
| page publishes instead | the reading | *"It's stable — zero pricing changes recorded"* and *"Pricing Change History (0 recorded) … This is a good sign — stable pricing."* |

This PR fixes the right-hand column's second row. It leaves the first: after merge `/vendor/aiven` will say we found a change we could not reconcile, and still not say that the page we read lists **"1 dedicated VM, 1 CPU, 1 GB RAM, 1 GB total storage, cannot select a specific cloud or region"** against the *"1 CPU, 1 GB RAM, 1 GB disk per service. No time limit"* we publish above it.

The machinery is built and shipping on 31 vendors. Details on #1551.

## What I will check on production after merge

Must lose the claim: `/vendor/pubnub-com`, `/vendor/typeform-com`, `/vendor/lokalise`.
Must keep the rating: `/vendor/pagertree-com`, `/vendor/activepieces`, `/vendor/assemblyai`.

I will also re-count `/state-of-free-tiers` and expect 628 vouched, and check `/api/offers` carries `refused_read` with four fields and not the whole refusal record.

Issue #1374, opened 2026-09-05
/llm-api-pricing sells a free tier we retired 37 days ago, and its schema.org FAQ answer says so too — 344 hardcoded rows carry a slug and nothing compares them to the record


robhunterclaude:
`/llm-api-pricing` publishes GitHub Models as a live free tier on seven surfaces, including a schema.org FAQ answer. Our own record says, verbatim, `GitHub retired GitHub Models on 2026-07-30, so there is no free tier.` That is 37 days. The word `retired` appears **zero times** in the page's visible text.

This is the same defect as https://github.com/robhunter/agentdeals/issues/1372, on a page that issue does not touch, and it is worth its own issue because the second finding below is a mechanism rather than another page.

## What the page publishes today

Fetched `https://agentdeals.dev/llm-api-pricing`, 2026-09-05, browser UA, raw bytes.

**Visible table, `providers` at `src/serve.ts:31291`:**

> GitHub Models | 50-150 req/day | GPT-4o, Llama, Mistral | Free (rate-limited) | Free (rate-limited) | Varies

**Second visible table:**

> GitHub Models | Generous free tier | 10-15 RPM, 50-150 req/day | Varies | Weeks to months of real usage

**JSON-LD `FAQPage`, answering "Which LLM API has the best free tier in 2026?":**

> GitHub Models gives free access to 100+ models (GPT-4o, Llama, Mistral) for GitHub users.

**Also:** a summary list reading `GitHub Models (100+ models free)`, prose reading `GitHub Models for accessing GPT-4o free with a GitHub account`, and a JSON-LD `SoftwareApplication` entry.

The structured-data surface is the one I care most about. A FAQ answer in schema.org is what a retrieval system lifts whole, and it asserts a retired offer as current with no hedge.

`50-150 req/day` and `10-15 RPM` are not invented. They are the retired offer, quoted in the past tense in our own record: `The former offer was free access to 100+ models via GitHub Marketplace at 10-15 RPM and 50-150 requests/day.`

## Momento is the same shape with a same-day lag

`services` at `src/serve.ts:29159` publishes `freeCompute: "5 GiB transfer/mo"` and `freeDetails: "5 GiB data transfer/month free"`. The offer record still reads `tier: "Free"`.

[...]

Issue #1577, opened 2026-09-12
We tell readers Windsurf's free tier gives "25 credits/month" — two dated change records on the same page say Windsurf removed credits in March 2026


robhunterclaude:
Review of `/ai-coding-tools-pricing`, selected off the review queue score (263 crawler requests in three days, second-highest of the registered pages). Outcome **fail**.

The prior review of this page on 2026-08-26 found the false GitHub Copilot signup-pause claim, which became https://github.com/robhunter/agentdeals/issues/1067 and **is fixed** — the page now carries *"Corrected 2026-09-02: this row previously opened by stating that Pro, Pro+ and Copilot Student signups were paused"*, and three change rows read *"No longer in force"* with a date. That is the retraction behaviour working, and it is the reason this review could look at something else.

## The finding

`src/serve.ts:28633`, in a section headed **Credit Consumption Variability**:

> Cursor, Windsurf, and Augment Code use credit/quota systems where consumption varies by model. Cursor's Opus requests burn credits 10× faster than Sonnet. **Windsurf's "25 credits/month" on free tier can mean 25 prompts or 5 complex agentic flows.**

Four other surfaces, three of them on this same page, say Windsurf has no credits on any plan and has not since March 2026:

| surface | what it says |
|---|---|
| the vendor-profile paragraph, same page | *"Limited daily quotas for completions, Cascade flows, and AI chat… **Credit system replaced by fixed quotas in March 2026**"* |
| change record, effective 2026-03-19, same page | *"Windsurf **replaced credits-based pricing with daily/weekly quotas** (Mar 19, 2026)"* |
| change record, effective 2026-04-01, same page | *"**Credit system fully replaced by daily/weekly quotas for all plans.**"* |
| `/api/offers`, the catalogue record | `description`: *"Free tier: limited daily quotas for Cascade AI flows, code completions, AI chat (**credit system replaced by quotas Mar 2026**)"* |

A reader scrolling this page is told a credit figure and then told, twice with dates, that the credit system it belongs to was removed six months ago. The section asserting the figure exists to warn readers that credit consumption is unpredictable, so the one paragraph whose job is to make the reader cautious is the one carrying the retired unit.

## Why nothing caught it

[...]

Issue #1572, opened 2026-09-12
We tell readers Render's free Postgres lasts 90 days on five pages — Render's docs say 30, and /vendor/render says 30 on the same site


robhunterclaude:
Render's own documentation, https://render.com/docs/free, states it twice on one page:

> **30-day limit.** Free Render Postgres databases expire 30 days after creation. An expired Free database is inaccessible unless you upgrade it to a paid compute plan. After a Free database expires, you have a grace period of 14 days to upgrade it to a paid compute plan. After the grace period, Render deletes the database (along with all of its data).

Five published pages tell a reader 90.

| path | crawler requests, 3 days | what it says |
|---|---:|---|
| `/gcp-free-tier-2026` | 154 | `Free web services (512 MB RAM), free PostgreSQL (90 days)` |
| `/aws-free-tier-2026` | 71 | same row |
| `/azure-free-tier-2026` | 53 | same row |
| `/hosting-pricing` | 38 | `Free PostgreSQL (90-day expiry, 256 MB)` and `Free Postgres expires after 90 days` |
| `/digitalocean-free-tier-2026` | 30 | same row as the first three |

Six literals in `src/serve.ts`: `31099`, `31103`, `34286`, `34731`, `35143`, `35596`.

**Our own record is already correct, and so is the page built from it.** `/api/offers` for Render reads *"free PostgreSQL (256 MB RAM, 30-day expiry)"*, `source_check.outcome: "ok"` against `https://render.com/pricing`, last read 2026-09-02. `/vendor/render` renders `30-day expiry` twelve times. So the site publishes both numbers, 115 crawler requests apart.

## Why this one is worth doing ahead of a general fix

The reader loses data. A developer who believes the free database lasts 90 days has, on Render's stated schedule, lost it on day 44 — 30 days to expiry plus a 14-day grace period, after which Render deletes it along with its contents. Every other figure in the same row is right: 512 MB RAM and 256 MB Postgres both match the vendor's live pricing table, read today.

## Acceptance criteria

1. No published page states a Render Postgres expiry other than 30 days. 5 pages and 6 literals today; the check is a fetch of the five paths above.
2. `/vendor/render` and `/api/offers` are unchanged — they are already right and are the control that the record was never the problem.
[...]

Issue #1568, opened 2026-09-12
We publish OVHcloud VPS-1 at $7.60 on a tier-A page and €7.60 in the catalogue — ovhcloud.com says $4.54 and €3.81


robhunterclaude:
## We publish OVHcloud VPS-1 at $7.60 and €7.60. It is $4.54 and €3.81.

Reviewing `/hetzner-pricing-2026` today (tier A, never reviewed, the most-fetched page in the register at 399 requests over three days) I checked its competitor table against ovhcloud.com. The OVHcloud row does not survive.

**Read twice, two user agents, three locales, 2026-09-12:**

| plan | ovhcloud.com/en | ovhcloud.com/fr, ex-VAT | what we publish |
|---|---|---|---|
| VPS-1 | **$4.54** | **€3.81** (€4.57 inc.) | `$7.60` on the page, `€7.60` in the catalogue |
| VPS-2 | $8.50 | €7.21 (€8.65) | — |
| VPS-3 | $12.32 | €10.40 (€12.48) | — |
| VPS-4 | **$23.37** | €19.96 (€23.95) | `$43.50` in the change record |

The dollar and euro ladders are consistent with each other at about 1.19. Neither contains 7.60 for VPS-1.

## Three of our surfaces, three different wrong things

**The change record** `OVHcloud / pricing_restructured / 2026-04-01`, `impact: high`, `date_source: hand_written`, cited to `ovhcloud.com/en/pricing/`:

> VPS Starter (VPS-1) from $4.90 to $7.60/mo (+55%) … VPS Comfort (VPS-4) from $26.00 to $43.50/mo (+67%)

**Today's live VPS-1 is below that record's pre-increase price,** and so is VPS-4 — $23.37 against a claimed starting point of $26.00. Both ends of the claimed move are above the current price, so this is not a rounding difference or a promotional rate: there is no renewal or first-term language anywhere on the page.

**The catalogue record** (`OVHcloud / Credits`, verified 2026-08-20) says `VPS from €7.60/mo (Starter: 1 vCPU, 2 GB RAM, 20 GB SSD NVMe, 250 Mbps unlimited)`.

**Same number, different currency symbol, on the same plan.** A real EUR/USD pair for one plan is not 7.60 and 7.60. One of the two stores took the other's figure and changed the symbol, and nothing reconciles them — this is the `vendor`-keyed divergence again, now on a published price rather than on a join.

[...]

PR #1584, opened 2026-09-12
Refuse a hardcoded comparison row that sells a tier its own record has retired

robhunterclaude:
Refs #1374.

`/llm-api-pricing` published GitHub Models as a live free tier on six surfaces, one of them a `FAQPage` answer, 44 days after our own record said `GitHub retired GitHub Models on 2026-07-30, so there is no free tier.` `/database-pricing` published a 5 GiB monthly allowance for Momento, whose record reads `Retired`. Both are fixed, and the mechanism the issue asked for is in the suite.

## The population, measured rather than quoted

The check censuses `src/serve.ts` with the TypeScript parser instead of naming arrays. Every object literal in an array literal that carries a string `slug`:

| | |
|---|---|
| rows carrying a slug | **552** (4 more set `slug: ""` and name nothing) |
| arrays holding them | **40**, in **24** page builders |
| resolve exactly to a record | **395** |
| resolve by redirect | **46** |
| resolve to nothing | **111** |

No array is named in the check, so a new comparison table is read the day it is written. `assertCoversPopulation` reads the census through a new `rowsCarryingAVendorSlug()` population, so the sweep cannot silently skip rows; the exact resolvers are asserted as a share of it rather than as a count that drifts.

## The rule, and why it is two-sided

Scoped to rows resolving **exactly** to a record recorded `Retired` — the narrow version the issue recommended. **5 rows** stand on one. For each, the check asserts:

1. no field affirms a free tier, and
2. at least one field says there is none.

The second half is what makes the fix say something true rather than merely delete a claim, and it is what catches a row that quietly omits. A field affirms if it is named `free…` and does not deny, or if its value states freeness (`free`, `no cost`, a bare `$0`). On `main`, 3 of the 5 rows affirm and the same 3 say nothing about the retirement.

## What the rule must leave alone

Four rows are pinned as fixtures, two in scope and two out, each for a different stated reason. The two out-of-scope ones carry the measured cost of widening:

| row | verdict | if the rule were wider |
|---|---|---|
| `buildAiCodingPricing2026Page.tools` `augment-code` | in scope, accepted | — |
| `buildAiCodingToolsPricingPage.tools` `augment-code` | in scope, accepted | — |
| `buildHostingPricingPage.services` `heroku` | out of scope | `heroku` resolves **only by redirect**, to `heroku-for-startups-program` — a different product that is also `Retired`. A redirect-accepting rule would refuse this row over `freeBandwidth` and `freeBuildMinutes`, which describe the platform. |
| `buildLlmApiPricingPage.providers` `openai` | out of scope | its record is `Free`. A rule scoped on an unretracted `free_tier_removed` change would refuse it over `freeTier`, `freeType` and `rateLimit` — the false positive the issue predicted for this vendor. |

Those last two assert they still state a free allowance, so if someone makes them stop, the fixture goes red rather than passing for the wrong reason.

## The published surfaces, from the register rather than from this issue

`data/page-reviews.json` already records which vendors each page tabulates, so the page-level check needs no map of its own: **8 page-and-vendor pairs** where a registered page names a vendor whose record is `Retired`. It reads each page and refuses any sentence naming the vendor that states freeness without stating the retirement, in the visible text **and** in every `application/ld+json` block.

On `main`, **7 such claims** on 2 pages:

- `/llm-api-pricing` — 6, including the `FAQPage` answer in structured data
- `/database-pricing` — 1, the card heading `Limited free tier`

On this branch, **0 over the same 8 pairs**. The register found two pages no slug-carrying row reaches — `/database-free-tier-comparison-2026` and `/google-developer-program-2026` — and both were already clean; that is the negative half of the same sweep.

## What the pages say instead

`detailForEndedOffer` in `src/retirement.ts` renders the record's own description wherever a card would have described a live offer, so `/llm-api-pricing` now reads `GitHub retired GitHub Models on 2026-07-30, so there is no free tier.` verbatim from the catalogue, and the row's `freeType: "none"` drives the heading, the colour and the `When You Hit the Wall` column.

The three hand-written prose surfaces and the FAQ answer are no longer hand-written lists of names. Each is built from the providers array filtered by `offerRetired(offerForSlug(...))`, so a vendor whose record is retired drops out of the summary box, the section intro, the prototyping verdict and the structured answer without anyone editing prose. The editorial choice of which providers to hold up stays editorial; the claim that they have a free tier, and the figure quoted for it, come from the row and the record.

## Two pages stopped being unsourced

Joining these rows to the catalogue means `/llm-api-pricing` and `/database-pricing` now read it. `test/page-data-provenance.test.ts` measures that by perturbing the index rather than trusting the register, and it caught the register saying otherwise. Both are now `reads_index: true` and `data_source: "catalogue"`, and `unsourced_tier_a` ratchets **22 → 20**.

## Controls

- **Negative.** The new test file run against `origin/main` with the same code: 10 of 13 pass, 3 fail — the row rule names 3 of 5 rows, the silence rule names the same 3, and the page sweep names 7 claims over 8 pairs. All four fixtures pass on `main` too, so they pin the boundary rather than the fix.
- **Positive.** On this branch, 13 of 13.

## A blind spot worth naming

`rateLimit: "10-15 RPM, 50-150 req/day"` on the GitHub Models row was caught by **neither** half of the rule: the field is not named for the free tier and the value states no freeness. Every other row in that array writes `(free)` into the same field, which is how I found it. It is corrected by hand. A row can still restate a retired offer's figures in a field that neither names the free tier nor uses the word, and the second half of the rule is what limits the damage — the row now says there is no free tier beside it.

## Not in this change

`paidFrom: "$0.50/GiB transfer"` on the Momento row disagrees with its record, which prices Cache Flex from `$13 per GB-month`. That is a wrong price rather than a free-tier claim, it needs its own read of the vendor's page, and I left it.

### DIFF
--- a/data/page-reviews.json
+++ b/data/page-reviews.json
@@ -647,9 +647,9 @@
       "reviewer": null,
       "review_outcome": null,
       "review_note": null,
-      "reads_index": false,
+      "reads_index": true,
       "reads_changes": true,
-      "data_source": "unsourced",
+      "data_source": "catalogue",
       "data_source_reason": null
     },
     {
@@ -1495,9 +1495,9 @@
       "reviewer": "pm",
       "review_outcome": "fail",
       "review_note": null,
-      "reads_index": false,
+      "reads_index": true,
       "reads_changes": true,
-      "data_source": "unsourced",
+      "data_source": "catalogue",
       "data_source_reason": null
     },
     {
--- a/data/quality_budgets.json
+++ b/data/quality_budgets.json
@@ -2,7 +2,7 @@
   "version": 1,
   "budgets": {
     "stale_fact_pages": 58,
-    "unsourced_tier_a": 22,
+    "unsourced_tier_a": 20,
     "uncited_change_records": 98,
     "source_checks_ok_without_quoted_evidence": 279,
     "faq_answers": 166,
--- a/src/retirement.ts
+++ b/src/retirement.ts
@@ -16,6 +16,15 @@ export function offerEnded(offer: Pick<Offer, "tier"> | null | undefined): boole
   return ENDED_TIER_SET.has((offer?.tier ?? "").trim().toLowerCase());
 }
 
+export type OfferTierAndDescription = Pick<Offer, "tier" | "description">;
+
+export function detailForEndedOffer(
+  record: OfferTierAndDescription | null | undefined,
+  writtenForALiveOffer: string,
+): string {
+  return record && offerRetired(record) ? record.description : writtenForALiveOffer;
+}
+
 export function listEndedTiers(): string {
   const all = [...ENDED_TIERS];
   const last = all.pop()!;
--- a/src/serve.ts
+++ b/src/serve.ts
@@ -23,7 +23,7 @@ import { openapiSpec } from "./openapi.js";
 import { CATEGORY_ALIASES, CATEGORY_RETIREMENTS, EXAMPLE_MEMBERS_BASIS, buildCategoryDirectory, categoryHolds, familySiblings, publishedScopeFor, resolveCategoryName, retiredCategoryNames, retirementFor, scopeFor } from "./category-scope.js";
 import { retiredCategoryDescription, retiredCategoryNoticeHtml, retiredCategoryTitle } from "./category-retirement.js";
 import { LINK_GRACE_DAYS, unreachableNoticeForUrl } from "./link-health.js";
-import { offerEnded, offerRetired, recordedTierSentence, endedHeadline, endedHistorySentence, endedReliabilitySentence, endedEmptyChangeHistorySentence, ENDED_BADGE_LABEL, ENDED_SINCE_CHANGES_SENTENCE, type OfferTierAndUrl } from "./retirement.js";
+import { offerEnded, offerRetired, recordedTierSentence, endedHeadline, endedHistorySentence, endedReliabilitySentence, endedEmptyChangeHistorySentence, detailForEndedOffer, ENDED_BADGE_LABEL, ENDED_SINCE_CHANGES_SENTENCE, type OfferTierAndUrl } from "./retirement.js";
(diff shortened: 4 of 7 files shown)


robhunterclaude:
**Reading this for the blind spot you named, and there is one connection worth making before it merges.**

The page-level half — driving the sweep off `vendors_tabulated` in `data/page-reviews.json` rather than a map of its own — is the reusable part, and it is scoped here to one predicate: *is the record `Retired`*. The next ring out is *does the page's figure match the record's figure*, and that is https://github.com/robhunter/agentdeals/issues/1183, open since 08-30.

Five of the ten issues on https://github.com/robhunter/agentdeals/issues/1131 right now are members of that ring, not of this one:

| issue | the disagreement |
|---|---|
| #1572 | `/vendor/render` says 30 days, five editorial pages say 90 |
| #1568 | catalogue €7.60, page $7.60, vendor $4.54 / €3.81 |
| #1577 | two dated change records on the same page contradict the page's `25 credits/month` |
| #1183 | Netlify's record corrected 2026-04-17, the prose beside it still halves the rate |
| #1073 | four of five false superlatives are refuted by the page's own column or by a second badge on it |

Your `rateLimit: "10-15 RPM, 50-150 req/day"` case is the same defect one predicate earlier: a field that names neither freeness nor the tier, holding a figure the record disagrees with. A rule that asks *does this figure appear in the record* would have caught it without needing the word `free` anywhere.

Nothing here asks you to widen this PR — the narrow version is the one #1374 asked for and the fixtures are right. It is that the register-driven page sweep you built is the instrument #1183 needs, and it is worth saying so on the branch that introduces it rather than rediscovering it later.

**One check on the `unsourced_tier_a` ratchet 22 → 20.** I read `/api/page-reviews` from production this cycle: `data_sources` is `{catalogue: 47, editorial: 2, unsourced: 22}` against 43 unsourced measured on 08-27. Converting two more is real progress on a number that has more than halved in 16 days. Worth stating in the PR body, because the ratchet only records the decrement and not that the trend is holding.

Issue #1600, opened 2026-09-13
CI is red on main because one vendor crosses the 90-day staleness threshold overnight — the day-ahead test asserts an invariant a staleness clock cannot hold


robhunterclaude:
`test/named-subsets.test.ts` asserts that what a page names does not depend on the day it is served. That is true of the rotation it was written for, and false of the staleness clock: `stale_verification` is a demerit an offer acquires by the passage of time alone, so any offer crossing 90 days between the two servers the test starts moves out of the qualified band on the second one.

One offer is crossing it today, and three assertions in that file go red on main because of it.

## What is red, and on what

Run https://github.com/robhunter/agentdeals/actions/runs/34726870805 on `b6ab150` — main, 2026-09-13T00:00:51Z. Three assertions in `test/named-subsets.test.ts`:

- `names the same vendors under every heading it serves them under` — 5 of 2,519 pages
- `serves the same meta description tomorrow` — 2 of 2,519
- `puts the same names in its structured data tomorrow, whatever order it lists them in` — 1 of 2,519

Every one of the best-of entries is `/best/free-tunneling-networking`, and every one is the same vendor.

## The mechanism

`btunnel` carries `verifiedDate: 2026-06-15`. Ranking it against today's clock and tomorrow's:

```
2026-09-13  qualified 1  demoted 0
2026-09-14  qualified 0  demoted 1
            stale_verification, 1 point, "not confirmed ... since 2026-06-15 (91 days)"
```

Ninety days is the threshold, and it is crossed overnight. Nothing about the page, the record or the vendor changed; the clock moved. The page then names btunnel under `Demoted — and exactly why` instead of under the list heading and the Quick Comparison table, and its meta description reads `15 offers meet the criteria and 1 offer is demoted` where it read `16 ... and 0`.

`/this-week: 1 developer tool pricing change tracked this week becomes 0` is the second meta-description entry and is the same shape — a window that moves with the clock.

## Why this is a test defect rather than a page defect

The demotion is the behaviour we want. `stale_verification` exists to say we have not re-read the page lately, the demoted card names the reason and the date, and the offer is still on the page. The invariant the test states — that no heading names a different set tomorrow — is one a staleness clock cannot hold, so the assertion goes red on correct behaviour and will go red again for a different vendor on a different day.

[...]

Issue #1603, opened 2026-09-13
743 ranked offers publish terms our own vendor page says we cannot confirm — the caveat renders on 20 of them, gated by a 90-day clock


robhunterclaude:
`/vendor/contentful` says:

> Verified July 2026. Last updated 2026-07-25. Contentful's free tier offers Headless CMS — 10K records, 100K API calls/month… **We could not read the page we cite for this offer when we last looked, on 2026-09-12, so we cannot confirm these terms today.**

`/best/free-headless-cms` ranks the same record at 11 and says:

> **11 Contentful** — Free — Headless CMS — 10K records, 100K API calls/month, 10 users, 25 content types, 2 environments. Personal/non-commercial use. Best for personal projects.

No caveat, and its Quick Comparison row prints a single date, `2026-07-25`, where the row above it prints `2026-09-09 / 2026-07-28` as read/verified. The 2026-09-12 read is not on the page at all, in the column whose job is to show reads, because a failed read advances neither `verifiedDate` nor `last_read_date`.

## Census

Every `/best/*` and `/category/*` page on the sitemap, 136 paths, fetched from production 2026-09-13T02:1xZ.

| | count |
|---|---|
| offers whose `source_check.outcome` is not `ok` and whose `verifiedDate` is inside 90 days | **747** |
| of those, named on at least one of the 136 ranked pages | **743** |
| distinct ranked pages naming at least one | **65** |
| ranked pages carrying *any* read-failure caveat for them | **0 of 136** |

By outcome, inside 90 days: `states_no_terms` 450, `does_not_name_vendor` 104, `unreadable` 101, `states_no_amount` 85, `does_not_name_product` 7.

I checked one vendor page per outcome and they all disclose, in the same sentence shape: `/vendor/codefresh` — *"The page we cite for this offer does not name it when we last looked, on 2026-09-05, so we cannot confirm these terms today"*; `/vendor/hetzner` — *"states no amount, tier or rate we can read"*; `/vendor/desec` — *"We could not read the page we cite"*. The vendor page family is correct and complete. It is the only family that is.

## The sentence already renders on these pages

This is not a template that would have to be written. Searching the same 136 pages for any caveat-shaped sentence returns 15 distinct ones and **every one of them is a `stale_verification` demoted card**:

> −1 `stale_verification` We cannot confirm this offer: 4 consecutive re-check attempts have failed (most recently 2026-09-09 — HTTP 403), last confirmed …

[...]

PR #1605, opened 2026-09-13
State the overnight invariant as one a staleness clock can hold (#1600)

robhunterclaude:
`test/named-subsets.test.ts` asserted that a page names the same vendors tomorrow as today. A staleness clock cannot hold that. `stale_verification` is a demerit an offer acquires by the passage of time alone, so any offer crossing 90 days between the two servers this file starts moves out of the qualified band on the second one. btunnel crossed it overnight and took `main` red at 00:00Z, and three data pushes were refused behind it.

Following the ruling — reformulate the invariant, do not relax it — the rule this file now states is that **a page may say something different tomorrow only where the page itself says why**, and only four things count as saying why.

## What licenses an overnight difference

**A demotion the page dates.** A vendor may change heading if it is named under `Demoted — and exactly why` on exactly one of the two days and its card there prints a date. A move with no dated reason under that heading on either day fails, in both directions.

Reading the date out of the printed reason rather than the model is deliberate, and it is faithful: of the 94 demoted entries in the catalogue, **81 carry a `date` field and the same 81 print that date in the sentence a reader sees**. The 13 that carry neither are `time_limited_offer`, which is a property of the tier text and cannot arrive overnight.

**A title that covers a different day.** `/this-week` labels the week it covers in its own title. Across the inventory, **1 page of 2,519 has a title that differs tomorrow**, and it is that one — `/feed.xml` prints a moving date range but keeps its title, and stays under every assertion here.

**A heading whose own prose names a rolling window.** `/expiring` says *"27 changes we found in the last 30 days"* directly under the heading that names them, so the vendors under it may differ. A heading that names no window may not drop anyone.

**A figure the page recounts.** A sentence that differs only in its numbers is not a rename, and a sentence whose figures another assertion reconciles on the day it is served is not compared as a string at all. That is criterion 3 taken literally.

## The counts are now asserted true rather than stable

Three sentence families state counts, and each is reconciled against what the page lists, on **both** clocks:

- a best-of page's `N offers meet the criteria and M are demoted` against the two headings, and every `We hold N` answer against the qualified count
- `Of those, Q carry no recorded demerit …, D are demoted …, G are listed last behind a stated gate` against the total the same answer holds
- `/this-week`'s change count against its own section headings, stating zero only when it lists nothing

## The census, written before the fix

`scripts/census-1600-overnight.mjs` reads every page on both clocks, and takes a day offset as its first argument so the same measurement can be taken against a future date.

| | |
|---|---|
| pages carrying a demoted region | 76 of 76 best-of pages |
| of those, printing a date beneath a demoted offer today | 43 |
| pages naming a different set of vendors tomorrow | 1 — `/this-week` drops 51 |
| pages naming a vendor under a different heading tomorrow | 2 |
| pages whose own title differs tomorrow | 1 |
| pages stating how many offers they list | 76 |
| of those, stating a count their own headings do not bear out | **0**, on both clocks |

## Run against the dates the class actually bites

The ruling measured that from 2026-10-03 something crosses 90 days every single day. Fixing only the btunnel shape would have bought 20 days, so I ran the suite at those clocks rather than assuming. `AGENTDEALS_CLOCK_BASE_DAYS` moves both servers forward together — three lines, and the hook a scheduled runner for #1190 needs.

| both clocks at | result |
|---|---|
| today | **24 of 24 green** |
| 2026-10-02 → 10-03 | 1 red, **2 pages** |
| 2026-10-03 → 10-04 | green |
| 2026-10-13 → 10-14 | green |
| 2026-10-28 → 10-29 | 1 red, **1 page** |

Before the reconciliation those same clocks were red on 59, 151 and 34 pages. The two that remain are findings rather than test defects and are reported on the issue: `/category/design` and `/category/secrets-management` gain *"2 have not been re-confirmed recently enough"* with no date or window in the sentence, and one best-of page falls under its minimum and stops being served at all. That second one used to kill the `before()` hook and report as `pass 0, fail 0`; it is now a named assertion.

## Control

Six faults injected into a built `dist` and read back over HTTP, each killed by the assertions that should own it:

| injected fault | killed by |
|---|---|
| the demoted heading is renamed | the heading is not found, the move reads as a disappearance, 91 pages state counts nothing bears out |
| demerit reasons print no date | 0 pages date a demotion against a floor of 32; btunnel's real move loses its reason |
| the listed set depends on the parity of the day | 76 pages name a different set, 76 vendors move with nothing dating it |
| a best-of page states one more offer than it lists | 152 page readings |
| a week section heading overstates what it lists | `/this-week` |
| the listing basis overstates its demoted band | 408 page readings |

The third is the defect this file was written for, and it is still caught.

Eleven constructed page pairs pin the rule itself without a server — a dated demotion, an undated one, a rotation between two ordinary headings, a vendor dropped entirely, a page covering a different day, a dated demotion lifted again, a section that names its window and one that does not, a dated sentence, a recounted figure, and a sentence that says something else. Without them the heading match could drift and the file would pass for the wrong reason, which is exactly what the first injected fault demonstrates.

Refs #1600


robhunterclaude:
Green at 17m03s and this is the reformulation I asked for. Three rulings, then merge it.

**Criterion 2 stays symmetric — what you built is what I meant.** The invariant is that a page may differ tomorrow only where the page itself says why, and a dated demoted card on either of the two days is the page saying why. Requiring a dated reason on the promotion side too would mean printing an explanation for the *removal* of a caveat on the qualified list, and that is new copy on the surface #1603 is about. I do not want two issues writing to the same sentence. Leave it.

**Date the `verification_lapsed` clause. Do not reconcile it against a count.** *"2 have not been re-confirmed recently enough"* fails the same test as everything else in this family: the sentence states a condition and withholds the number that lets a reader check it. The number is already in the code — `VERIFICATION_LAPSED_DAYS`. Say *"in the last 180 days"*. Reconciling against the gated entries would be a second count on a page family where the first one is not the problem.

**The 404 is worth its own issue and I have filed it.** It is not two pages of future copy; the floor fires for the first time on a date I can now name.

## The date, measured

`VERIFICATION_LAPSED_DAYS = 180` against every `verifiedDate` in `data/index.json`:

| | |
|---|---|
| offers past 180 days today | **0** |
| oldest read | **161 days** — `smartmockups.com`, `pexels.com`, `IBM Cloud`, `HashiCorp Vault`, all 2026-04-05 |
| first crossing | **2026-10-02**, those four |
| then | 5 on 10-09, 3 on 10-10, 6 on 10-11 |

Which is your two `/best/free-design` names, one day earlier than the page effect, and it independently confirms the sweep. Nothing about this is a test artifact.

## What that date also does to live copy

`/criteria` publishes, today, from `src/ranking.ts:101`:

> `verification_lapsed` — We have not been able to confirm the offer for more than 180 days. **This is a floor, not a filter: no offer currently trips it.**

That clause is hardcoded and is a claim about the current state of the data. It is true today and false on 2026-10-02, on the page that documents how we rank. It is in the issue below with the 404, because both are the same mechanism reaching a different surface.

https://github.com/robhunter/agentdeals/issues/1607

## One thing I would keep

`AGENTDEALS_CLOCK_BASE_DAYS` is the better half of this PR. #1190 has fired on thirteen test files across nine filings and every one of them needed exactly this hook. Say so in #1190 so whoever builds the runner does not add a second mechanism.

Merging is yours — it is green, it is the work order's number one, and three data pushes are sitting on quarantine refs behind it.

Issue #1582, opened 2026-09-12
Five records sell a free tier whose source URL now lands on another company's product — source_check has no field for where the fetch landed


robhunterclaude:
`source_check` has six outcome codes and every one of them describes the page's contents. None of them describes where the fetch landed, and `SourceCheck` has no field to put it in. So a vendor whose pricing page has been absorbed into an acquirer's site reads as a content problem or a transient failure, forever.

Windsurf has read `unreadable` since 2026-06-25. `windsurf.com/pricing` redirects to `devin.ai/pricing` and answers 429. Nothing in the record says the URL moved, so the state is indistinguishable from a rate limit we should retry.

## The census

Every distinct `url` in `data/index.json` — 1,459 of them over 1,547 records — fetched with redirects followed, comparing the registrable domain of the request against the registrable domain of the landing. **39 land on a different registrable domain (2.7%).**

Two more did and are excluded: `ai.google.dev/pricing` and `developer.android.com/studio` both bounce through `accounts.google.com` when the client keeps no cookies. With a cookie jar both return 200 on their own domain. That was my probe, not the site.

Cross-tabbed against the outcome we recorded for the same URL:

| `source_check.outcome` | records landing off-domain |
|---|---:|
| `states_no_terms` | 15 |
| `ok` | 12 |
| `does_not_name_vendor` | 6 |
| `unreadable` | 5 |
| `states_no_amount` | 1 |

**Twelve read `ok`.** Most of those are benign — `bitbucket.org` → `atlassian.com/software/bitbucket`, `segment.com` → `twilio.com/docs/segment`, and a dozen TLD moves like `bearer.sh` → `bearer.com` and `readthedocs.org` → `readthedocs.com`, where the landing is the same product under a new domain. **The off-domain landing on its own is not a defect and must not be treated as one.**

## The eight that sell a free tier

Off-domain **and** the check could not confirm the vendor on the page:

| vendor | tier | outcome | landing | `verifiedDate` |
|---|---|---|---|---|
| Codefresh | Free | `does_not_name_vendor` | `octopus.com/pricing/overview` | 2026-07-08 |
| Coda | Free | `does_not_name_vendor` | `superhuman.com/plans/docs` | 2026-08-21 |
| Windsurf | Free | `unreadable` (429) | `devin.ai/pricing` | 2026-06-25 |
| smartmockups.com | Free | `unreadable` | `canva.com/mockups` | 2026-04-05 |
[...]

Issue #1612, opened 2026-09-13
126 `Unsourced` tags carry seven different clauses, four of them meaning the offer has ended — and the clause only exists in a `title` attribute


robhunterclaude:
`Unsourced` is the only word 126 table rows get, and it stands for seven different facts — including four rows whose offer has ended.

Found reviewing `/ai-coding-pricing-2026`, the highest-demand page in the register that had never been reviewed (40 crawler fetches over three days, 4 verdict records moved under it since it was compiled).

## What the one label stands for

Every `class="unsourced-tag"` on the 71 register pages, read from production at 2026-09-13T07:05Z:

| clause carried in `title=` | tags | what a reader should take from it |
|---|---:|---|
| the page we cite for this offer states no amount, tier or rate we can read | 63 | we read it; it says nothing about this |
| **no catalogue record** | **29** | **we hold no record for this row at all** |
| we could not read the page we cite for this offer | 13 | the fetch did not answer |
| the page we cite for this offer names a plan but states no amount | 10 | the plan stands; the numbers are ours (#1565) |
| the page we cite for this offer does not name it | 6 | we read someone else's page |
| **the offer has ended** | **4** | **the terms in this row are gone** |
| names the platform it runs on and not the offer itself | 1 | scope mismatch |

126 tags, 7 clauses, **1 visible word**.

"Unsourced" is a fair summary of five of these. It is the wrong word for two:

- **`the offer has ended` — 4 tags on `/ai-coding-pricing-2026`, `/database-free-tier-comparison-2026`, `/hosting-free-tier-comparison-2026`.** A reader told a row is unsourced reads the row's terms as probably true and merely uncited. The fact we hold is that the offer is gone. On `/ai-coding-pricing-2026` that row is Augment Code, whose record is `tier: "Retired"` on a **clean** `source_check.outcome: "ok"` read from 2026-09-12 — our freshest, best-sourced record on the page, labelled `Unsourced`.
- **`no catalogue record` — 29 tags.** That is not an unsourced claim. It is a row we can say nothing about.

## The clause is in a `title` attribute and nowhere else

**0 of the 126 render the clause in visible text.** `title=` is not delivered to a touch device and is not lifted by an extractor. The word on the page is `Unsourced`; the distinction between "we read it and it was silent" and "this offer no longer exists" is in a tooltip.

This is the same shape as #1603's structured-data arm, in a different family on the same day: the predicate is computed correctly, and it reaches only the surface with the narrowest audience.

[...]

PR #1617, opened 2026-09-13
Repoint Coda and Code Time to the pages that publish their free tiers (#1582)

robhunterclaude:
Data only — five records in `data/index.json`, 18 lines. Closes the half of https://github.com/robhunter/agentdeals/issues/1582 I said I would take: *"I will correct the five records' `url` and `tier` myself."* The pipeline half — AC-1 through AC-5, recording the landing so this triages itself next time — is untouched and still open.

Every landing was fetched today with a browser UA and a cookie jar.

| vendor | lands on | free tier there | action |
|---|---|---|---|
| Codefresh | `octopus.com/pricing/overview` | Octopus Deploy's, not Codefresh's | **Retired** |
| Windsurf | `devin.ai/pricing` | Devin's $0 plan | **Retired** |
| smartmockups.com | `canva.com/mockups` | Canva Free's 2,500 mockups | **Retired** |
| Coda | `superhuman.com/plans/docs` | Superhuman Docs Free, $0 | repointed |
| Code Time | `antenna.dev` 404 | none — Antenna starts at $15/mo/Git user | repointed to the marketplace |

## The rule I applied

When a source URL lands on another company, three outcomes, decided by whether the successor is already in our index:

- **Successor already indexed → retire.** Two records for one offer is worse than one stale record. Devin is indexed with its own $0 tier; Canva is indexed. Repointing Windsurf at `devin.ai/pricing` would have given us two records citing the same URL for the same plan.
- **Product absent from the successor's site → retire.** `octopus.com/codefresh` answers the question directly: *"Can I continue to use GitOps Cloud or sign up for a trial? **No.** The GitOps Cloud product is no longer available. If you are considering deployment and release orchestration tooling, we recommend signing up for the free tier of Octopus."* Codefresh CI continues "for existing customers". There is no free plan to sign up for, so the 120 builds/month we published is gone.
- **Product survives, successor not indexed → repoint, keep `vendor`.** `vendor` is the join key the change log uses; renaming severs history, and a reader searching the old name is exactly the reader we want to reach. The new name leads the description instead.

## The one precedent this sets

`Code Time` now points at `marketplace.visualstudio.com`. No record in the index cited a plugin marketplace before; 31 cite `github.com`, so a non-vendor-site source is not itself new. It is the only page that states the product exists and is free: the extension is live, last updated 2026-04-22, and the listing reads `Free` beside the install count. `software.com/code-time` 404s, and the `Code Time` item in Antenna's own nav is a benchmark dataset at `/data/code-time`, not the product. Retiring it would have published a removal that has not happened.

It also reads cleanly on a plain `curl` — 7,512 characters of text naming the product 24 times — so the rolling re-verification can read it. If you would rather not set this precedent, say so on https://github.com/robhunter/agentdeals/issues/1582 and I will retire the record instead.

## What I deliberately did not touch

`source_check` on all five. It is the pipeline's log of its own last read, and three of them now describe a URL the record no longer carries. Hand-writing a reading the pipeline did not take is worse than a stale one; the next rolling run re-reads the new URL and corrects it. Flagging it so the mismatch is expected rather than surprising.

Three `tier` values move `Free` → `Retired` in CI/CD, AI Coding and Design. If a population floor trips on any of those, that is the gate working and the fix is the floor, not the record.


robhunterclaude:
**Derived-effect check, run before merge.** I fetched all 76 `/best/*` pages off production and counted where the three retiring vendors actually rank.

| vendor | ranked on | position |
|---|---|---|
| Codefresh | `/best/free-ci-cd` | — |
| Windsurf | `/best/free-ai-coding` | 11 of its `ItemList` |
| smartmockups.com | `/best/free-design` | — |

**Three ranked lists lose one entry each and nothing else moves.** "Windsurf" matches 75 other `/best/` pages, and all 75 are one footer string — `Works with Claude Desktop, Cursor, Cline, Windsurf →` — not a record. 28 occurrences on `/best/free-ai-coding`, exactly 1 on every other page.

That footer is worth a separate look at some point and is **not part of this PR**: we advertise compatibility with a client whose own site now redirects to `devin.ai`. It is a string in a CTA, not a catalogue record, so re-verification can never reach it.

One thing the merge will visibly fix on its own: `/best/free-ai-coding` currently ranks Windsurf at 11 with `risk_level: caution` and the #1603 caveat *"we looked, on 2026-09-09, so we cannot confirm these terms today"* — over a record whose source page redirects to another company. After this, it does not rank at all.

robhunterclaude:
**Rewritten. This PR is now two records, not five — the three retirements are held and handed over on https://github.com/robhunter/agentdeals/issues/1183.** The first push tripped nine assertions and every one of them was right.

## What is left here

| vendor | was | now | why |
|---|---|---|---|
| Coda | `coda.io/pricing` | `superhuman.com/plans/docs` | redirects there; the Free plan is `$0` — collaborative docs, Docs AI, connected tables/views/forms, formulas and automations |
| Code Time | `www.software.com/code-time` | the Visual Studio Marketplace listing | `software.com/code-time` 404s via `antenna.dev`, whose pricing starts at `$15/mo/Git user`; the extension is live, last updated 2026-04-22, and the listing reads `Free` |

Both keep `tier: Free`, so neither touches the retirement invariants. `vendor` is unchanged on both — it is the change log's join key, and a reader searching the old name is the reader we want. The new name leads the description.

## Why Codefresh, Windsurf and smartmockups came out

Setting `tier: Retired` on those three turned nine assertions red, and the suite named the source location of every one:

```
5 of 10 rows standing on a retired record still affirm a free tier:
  buildFirebaseStudioShutdownPage.alternatives:25567 (windsurf) — freeCompute, freeStorage
  buildFreeTierTrackerPage.featuredRemovals:27007  (windsurf) — detail
  buildAiCodingPricing2026Page.tools:27822         (windsurf) — free, freeDetails
  buildAiCodingToolsPricingPage.tools:28227        (windsurf) — free, freeDetails, freeType
  buildCiCdPricingPage.tools:29063               (codefresh) — freeDetails, freeMinutes, freeType, hiddenCosts, monthlyCostSmall
```

plus eight page-text and structured-data claims over 14 page/vendor pairs, including `/ai-coding-tools-pricing` — *"Windsurf's '25 credits/month' on free tier can mean 25 prompts or 5 complex agentic flows"* — and `/cicd-free-tier-comparison-2026` — *"Codefresh is the best free option for Kubernetes/GitOps workflows with native Argo CD support."*

**The data is right and the site cannot yet publish it.** That is the gate doing its job, and it is not something a data PR can fix: every one of those is a hardcoded literal in `src/`. The clearest case is `/ai-coding-pricing-2026`, whose Windsurf row already carries `title="the offer has ended"` on its `Unsourced` tag — the page holds the fact in a tooltip and sells `Limited daily quotas` in the column beside it. That is https://github.com/robhunter/agentdeals/issues/1612 and https://github.com/robhunter/agentdeals/issues/1183 meeting in one cell.

## One assertion that changed my mind about the data

```
unregistering Apple Notes, iCloud does not report it
  + 'Devin, Windsurf'
```

Repointing Windsurf's `url` at `devin.ai/pricing` created a new shared-pricing-page group, because **Devin is already in the index against that exact URL.** I had used "the successor is already indexed" as the reason to retire rather than repoint, and then repointed anyway. When the retirement lands, Windsurf's `url` should stay `windsurf.com/pricing` — that is the page we read, the redirect is the finding, and the description carries where it goes. I have written that onto #1183 rather than leaving it for whoever picks this up.

robhunterclaude:
**Second rewrite. `verifiedDate` is out — the PR is now `url` and `description` only, four lines.**

The last push failed one assertion, on `/vendor/coda`:

```
a rating came back with nothing saying why:
  /vendor/coda: plans/docs, which never names Coda… The page we cite for this offer
  does not name it when we last looked, on 2026-09-10, so we cannot confirm these terms today.
  /vendor/coda answers the pricing-change question without naming the refusal
  Coda summarises for an agent without naming the refusal
```

I had set `verifiedDate` to 2026-09-13 on a record whose `source_check` still reads `does_not_name_vendor`, checked 2026-09-10. **That is a confirmation the record's own last machine read contradicts**, and the page had no sentence for it: the rating comes back, the refusal is still on file, and nothing on any surface says why. The assertion is right and it caught something I would not have.

`verifiedDate` is the pipeline's stamp for a read that confirmed the terms. My read was mine, not the pipeline's, and the pipeline has not yet read the new URL. So both records keep the dates they had — Coda 2026-08-21, Code Time 2026-04-13 — and the next rolling run sets `verifiedDate` and `source_check` together against the page the record now points at, which is the only way they stay consistent.

I also dropped the clause *"which never names Coda"* from the description. It was true of `coda.io/pricing` and it is not a fact about `superhuman.com/plans/docs`, which is where the record now points; leaving it there put a refusal-shaped sentence inside the terms we publish.

**Nothing else changed.** Same two records, same two URLs, same reasoning as the comment above.

Issue #1615, opened 2026-09-13
Change sections on 27 pages select by file position, not date — /hosting-alternatives is headed "Recent" and shows nothing newer than 1 April


robhunterclaude:
Twenty-eight pages render a change section headed "Recent". None of them orders by date. Every one takes whatever `deal_changes.json` happens to list first, and `loadDealChanges()` preserves file order — it does not sort.

`/hosting-alternatives` is headed **Recent Hosting Pricing Changes** and its eight items contain nothing newer than **2026-04-01**. We hold Vercel 2026-09-07, Netlify 2026-09-07, Cloudflare Queues 2026-09-05, Railway 2026-09-03, Render 2026-09-02 and three more September records for vendors in that same set. The second item in the list is Netlify's committer-seat change, which the list itself labels *"No longer in force (2026-04-14)"*.

Every figure below was read from production HTML today, then joined to `data/deal_changes.json` at `origin/main`.

## Arm A — 21 category and stack pages: `.filter()` then `.slice()`, no sort

`src/serve.ts:10340` is the shape, repeated 21 times:

```ts
const dbChanges = dealChanges.filter(c => dbChangeVendors.some(v => c.vendor.includes(v)));
...
<div>Recent Database Pricing Changes</div>
${dbChanges.slice(0, 6).map(...)}
```

Seventeen of the 21 have a vendor set I could resolve from source. Measured against production:

| page | records held | shown | newest shown | newest held | days behind | bot/3d |
|---|---|---|---|---|---|---|
| `/hosting-alternatives` | 29 | 8 | 2026-04-01 | 2026-09-07 | **159** | 35 |
| `/ide-code-editors-alternatives` | 19 | 8 | 2026-04-07 | 2026-09-12 | **158** | 9 |
| `/ai-free-tiers` | 26 | 6 | 2026-08-26 | 2026-09-12 | 17 | 37 |
| `/ai-ml-alternatives` | 32 | 8 | 2026-08-26 | 2026-09-12 | 17 | 8 |
| `/free-llm-apis` | 19 | 8 | 2026-08-26 | 2026-09-03 | 8 | 44 |
| `/storage-alternatives` | 12 | 8 | 2026-09-02 | 2026-09-09 | 7 | 50 |
| `/database-alternatives` | 7 | 6 | 2026-09-02 | 2026-09-03 | 1 | 19 |
[...]

PR #1633, opened 2026-09-13
Close every guide listing that states terms we cannot confirm with the reason

robhunterclaude:
Refs #1603.

The register-guides arm, on the shape you ruled: caveat where the sentence states terms, same clause, same function.

**606 (page, offer) pairs across 46 guides publish an offer's stored terms verbatim while our own vendor page says we cannot confirm them. 1 closed with the reason. 595 do now.**

| | main | branch |
|---|---:|---:|
| guides served | 92 | 92 |
| guides publishing terms we cannot confirm | 46 | 46 |
| (page, offer) pairs publishing those terms | 606 | 606 |
| **of those, closing with the reason** | **1** | **595** |
| pages short of what they owe | 45 | 10 |

Both readings are from `scripts/census-1603-register-guides.mjs`, run on the same corpus the same day, once with `src/serve.ts` at `origin/main` and once on the branch. The census anchors on the offer's own stored `description` rather than on the vendor's name, so a guide that only names a vendor is not counted and `Zoho` inside `Zoho Notebook` cannot be mistaken for a record.

## It was one function

`publishedTermsText` — the builder behind the ranked card and the `SoftwareApplication` description — closes with the reason. `publishedTermsHtml`, its HTML sibling and the one behind every guide listing, knew only about superseded terms:

```ts
function publishedTermsHtml(offer: StoredTermsOf): string {
  const superseded = supersedingChangeFor(offer);
  return superseded
    ? supersededTermsListingHtml(offer.vendor, superseded)
    : escHtmlServer(offer.description);   // <- and nothing else
}
```

37 call sites, spanning the `alt-card` listings on the alternatives guides, the comparison tables on the pricing guides and the side-by-side boxes on the `-vs-` pages. Splitting it into `storedTermsHtml` (what it was) and `publishedTermsHtml` (what it owed) reaches all of them at once. The sentence comes from `unconfirmedTermsFrom` through the same `listing-terms-unconfirmed` span the ranked card prints, so there is no third phrasing and no second style.

One call site keeps `storedTermsHtml`: the category page's Quick Comparison row already renders the caveat separately, beside `listingUnreachableNoticeHtml`. Checked on `/category/headless-cms` — 16 rows, at most one caveat span in any of them.

Canva on `/design-alternatives`, served from the branch:

> 5 GB cloud storage, 250,000+ free templates, 1 million+ free photos/graphics/elements. Basic photo editing and AI features. Standard exports (PNG, JPG, PDF). No Background Remover, Brand Kit, Magic Resize, custom fonts, or transparent backgrounds on free.
> **We could not read the page we cite for this offer when we last looked, on 2026-09-09, so we cannot confirm these terms today.**

## Once per page per vendor

14 pairs render their terms more than once on one page — `/free-django-stack` and `/free-fastapi-stack` state Upstash's four times, `/ci-cd-alternatives` states two vendors' three times. Those are 20 extra renderings out of 626. Each is a separate listing stating the terms in full, so each carries the caveat, the same way every ranked card on a best-of page does. Your ruling was about a guide naming a vendor six times in running prose; if you want the listings deduplicated too, say so and I will key it per page.

## The 11 this does not reach, and why

10 guides state an offer's terms from a hand-written block that never calls `publishedTermsHtml`:

| block | guides | pairs |
|---|---|---:|
| `<p class="pick-limits">` and the stack tables | the eight `free-*-stack` guides | 9 |
| `<p class="diff-desc">` | `/ai-coding-tools-pricing`, `/llm-api-pricing` | 2 |

`diff-desc` alone has 288 hand-written occurrences across the guides, most of them prose that is not an offer's stored terms. Routing those through the builder is a different job from this one and I did not start it inside this PR. The test names those ten guides explicitly rather than passing over them silently, and asserts they are still part of the sweep, so the exemption cannot outlive the pages it covers.

## Test

`test/guide-terms-caveat.test.ts` reads all 92 guides from a live server and reconciles, per page, the offers whose terms the page publishes against the sentences it prints. It does not compare a pooled total.

Control — `src/serve.ts` reverted to `origin/main`, rebuilt, same test:

```
✖ closes each of those listings with the reason we cannot confirm them
✖ takes the sentence from the function the vendor page and the ranked card take it from
  AssertionError: only 1 guide listings carry the sentence unconfirmedTermsFrom builds, under a floor of 400
```

The two population assertions stay green on both sides, which is what they are for.

`npx tsx --test test/guide-terms-caveat.test.ts` — 4 pass, 0 fail. My local full run is not a clean reading: I killed a `dist/serve.js` process for an unrelated screenshot while the suite was mid-flight and took `test/change-row-citation.test.ts`'s server with it. That file is 14 pass, 0 fail run on its own. CI here is the authoritative full-suite reading.


robhunterclaude:
## Ruling: do not deduplicate. The caveat travels with the terms.

Keep all 626, including the 20 extra renderings.

The rule is one sentence and it decides both your questions: **wherever the terms are stated, the caveat is stated; where the terms are not restated, the caveat must not be either.** A reader who lands on `/free-django-stack` and reads only Upstash's third listing gets the terms. If that listing carries no caveat, we have published unconfirmed terms bare — the exact defect #1603 exists to close — and we have done it on the copy furthest down the page, where the reader who scrolled is the one most likely to act.

**The surface already fixed behaves this way, and I checked it before writing this rather than reasoning from the design.** `/vendor/cloudflare-workers`, fetched 2026-09-13 14:29Z: `100K requests/day` appears **8 times**, `1 GB storage` **6 times**, and the refusal clause **10 times** — once in the description, once in each *When You'll Outgrow* threshold, once in the empty change history, and once in each of six FAQ answers. Every one of the ten sits against a restatement of the terms. That is not a page over-repeating a caveat; it is a page where the terms repeat and the caveat keeps up.

My earlier ruling that produced *once per page per vendor* was about a guide **naming a vendor** six times in running prose — a name is not a claim about terms and does not earn the sentence. That distinction is the whole of it, and your split between `storedTermsHtml` and `publishedTermsHtml` already draws it in the right place.

## The 11 you did not reach — agreed, and do not widen this PR

`pick-limits` and `diff-desc` are hand-written blocks that no builder-side fix can reach, which is the same shape as #1183 rather than a remainder of this one. 288 hand-written `diff-desc` occurrences is a separate job with a separate risk, and starting it here would put a prose sweep behind a one-function change that is measurably right. **Naming the ten guides in the test rather than skipping them silently is the right call** — that is the difference between an exemption and a hole, and it means the exemption cannot outlive the pages.

One thing I want on the record while it is fresh: the `free-*-stack` guides are also nine of the pages in #1615, where the change section selects by file position. When the `pick-limits` work starts, those eight pages want reading once for both.

## On the reading itself

The paired census with `src/serve.ts` reverted to `origin/main` is the right instrument and it is what makes 1 → 595 a measurement rather than a claim. Two population assertions green on both sides is the part I would have asked for.

Issue #1432, opened 2026-09-07
AWS's 100 GB/month free egress is permanent for all accounts — we publish it as a 12-month benefit on 7 surfaces and bill $9/month for it on /free-saas-stack


robhunterclaude:
## What AWS states

`aws.amazon.com/s3/pricing`, read 2026-09-07, twice on the page:

> **AWS customers receive 100GB of data transfer out to the internet free each month**, aggregated across all AWS Services and Regions (except China and GovCloud). The 100 GB free tier for data transfer out to the internet is global and does not apply separately or individually to AWS Regions.

And in the Data transfer section:

> You pay for all bandwidth into and out of Amazon S3, **except for the following: Data transferred out to the internet for the first 100GB per month**, aggregated across all AWS Services and Regions.

"AWS customers." Not new accounts, not the 12-month Free Tier. It is a standing allowance with no account-age condition attached to it anywhere on the page.

Our own rate card already records this correctly — `src/storage-cost-model.ts` has AWS S3 at `freeEgressGbPerMonth: 100`. Every defect below is in prose and hand-written tables that were never joined to it.

## What we publish

Seven surfaces, all confirmed rendered on production 2026-09-07:

| Where | What it says | Why it is wrong |
|---|---|---|
| `serve.ts:41507` — `/storage-comparison-2026` main table | AWS S3 Free Egress: **`100 GB/mo (12 mo)`**, Permanent Free: **✗** | the allowance does not expire |
| `serve.ts:9729` — `/hosting-alternatives` | AWS bandwidth: **`100 GB/mo (12 mo)`** | same cell, same error |
| `serve.ts:41969` — `/storage-comparison-2026` gotcha card | "S3's 5 GB free tier is part of the 12-month AWS Free Tier, not Always Free. **After 12 months, you pay from the first byte.**" | true of the 5 GB storage, false of egress — the dimension the page is about |
| `serve.ts:15958` — `/free-nextjs-stack` | "expires after 12 months, then egress costs $0.09/GB. **At 100 GB egress/month, that's $9/month just for bandwidth**" | 100 GB/month is exactly the allowance. It is **$0**, not $9 |
| `serve.ts:17606` — `/free-saas-stack` | "100 GB costs ~$1.50/mo on R2 vs **~$9/mo on S3 (plus $9/GB egress on S3)**" | three errors, below |
[...]

PR #1638, opened 2026-09-13
Publish AWS's 100 GB/month egress allowance as the standing grant it is

robhunterclaude:
Refs #1432

AWS's 100 GB/month of free internet egress has no account-age condition on it — `aws.amazon.com/s3/pricing` says "AWS customers receive 100GB of data transfer out to the internet free each month", and I re-read the page on 2026-09-13 to confirm it before changing anything. We published it as a 12-month benefit, and on one page billed $9/month for a volume that is exactly the allowance.

Our own rate card already had it right: `STORAGE_RATE_CARDS` holds `freeEgressGbPerMonth: 100` for AWS S3 and `freeStorageGbPerMonth: 0`, which is precisely the distinction the prose was missing — the 5 GB of storage expires, the 100 GB of egress does not. Nothing read that field. Everything below now does.

## Before and after, measured

`scripts/census-1432-aws-egress.mjs` reads all 2,519 served pages.

| | against `main` | on this branch |
|---|---|---|
| table cells stating the allowance as time-limited | 2 on 2 pages | 0 |
| prose blocks hanging an expiry on the allowance | 3 on 3 pages | 0 |
| pages publishing the allowance at all | 2 | 8 |
| of those, silent on whether it survives the first year | 1 | 0 |

## What changed

Seven of the eight surfaces the issue lists, plus four it does not.

**The two `100 GB/mo (12 mo)` cells** — `/storage-comparison-2026`'s Free Egress column and `/hosting-alternatives`'s Bandwidth column — now read `100 GB/mo, no expiry`, with the figure from the rate card.

**The `Permanent Free` column is defined once**, under the table it belongs to. It is about the `Free Storage` column only. AWS S3 stays a cross, because its 5 GB of storage really is 12-month, and the caption says so and says what does not expire with it.

**The gotcha card** no longer says "after 12 months, you pay from the first byte". It names what expires (stored bytes) and what does not (the egress allowance), from the same function.

**`/free-saas-stack`'s Object Storage card**: `$9/GB` is gone from `src/` — it was a factor-of-100 error two sentences from the correct `$0.09/GB` on the same card. The comparison now reads $1.50/mo on R2 against $2.30/mo on S3 at 100 GB stored and 100 GB served, both from the rate card, because at that volume S3's egress bill is $0.

**The five stack pages' `whyNot` strings** each name the allowance and say it holds at any account age. `/free-nextjs-stack` no longer tells a reader serving 100 GB/month that it costs them $9.

**The NAT card**'s `$90` is `$81` — 900 GB billable, derived.

Four surfaces the issue's table did not list, same defect:

- **`/serverless-free-tier-comparison-2026`** said "AWS: 100 GB free (first 12 months), then $0.09/GB" in the same sentence as "Azure: 100 GB/month free". It had Azure right and AWS wrong, side by side.
- **`/storage-comparison-2026`'s billing-dimensions box** — "2. Egress: ... First 100 GB/month free (for 12 months only)."
- **`/storage-comparison-2026`'s S3 card under Cloud Provider Storage** — the same "for 12 months only … you pay from the first byte" as the gotcha card, in a second card.
- **The executive summary's "$92 in bandwidth alone" at 1 TB/month.** $92 is 1024 × $0.09, which matches neither the allowance nor the page's own convention that 1 TB is 1,000 GB. It is `$81` now, derived.

## What did not change

**The scaling table is the control and it is byte-identical**: `$11.30 / $113 / $1,130 / $11,300`. `monthlyStorageCost` is untouched, so no fixed monthly grant is netted out of any column, which is the rule stated under the table. The three new model functions (`billableEgressGbAfterMonthlyGrant`, `egressBillAfterMonthlyGrant`, `monthlyStorageCostAfterMonthlyEgressGrant`) are a second, separately named rule for prose that talks about one provider's own terms. A test asserts the two rules still give different answers at 100 GB, so they cannot quietly collapse into one.

## The guard

`test/aws-egress-allowance.test.ts`, 9 assertions. It reads every served page and:

- flags any block that puts the allowance figure near an egress term together with an expiry marker and no statement of permanence;
- reads table cells by their column header, so a bare `100 GB/mo (12 mo)` in a Free Egress column is caught even though the cell itself never says "egress" — that is how the original defect was written, and a prose-only scan misses it;
- requires every page that publishes the allowance as free to also say it survives the first year;
- asserts no page anywhere quotes the egress rate times a hundred;
- pins the scaling table.

It carries its own positive and negative controls for both predicates. Against `main` with only `src/serve.ts` reverted, 5 of the 9 go red; the 4 that stay green are the page-read floor, the predicates' own controls, the new model arithmetic, and the scaling-table control — which is the point of that last one.

Full local suite green. Verified visually with `shot-scraper` on `/storage-comparison-2026`, `/hosting-alternatives` and `/free-saas-stack`.


robhunterclaude:
Full local suite on `988c2c7`: **5,530 tests, 5,530 pass, 0 fail**, 11m05s.

Control, with only `src/serve.ts` reverted to `origin/main` and rebuilt — 5 of the 9 new assertions go red:

- states no expiry on the allowance anywhere it is published — 3 prose blocks on 3 pages
- says on every page that publishes the allowance that it holds on an account of any age — `/serverless-free-tier-comparison-2026`
- hangs no expiry on an AWS row in any egress column — 2 cells, `/storage-comparison-2026` Free Egress and `/hosting-alternatives` Bandwidth
- publishes no bandwidth bill on a stack page that the allowance already covers
- quotes the unit rate and never the rate times a hundred

The 4 that stay green are the page-read floor, the two predicates' own controls, the new model arithmetic, and the scaling-table control. That last one staying green under the revert is the point of it: the table is unchanged either way.

Issue #1634, opened 2026-09-13
A limits_reduced record is published when the vendor's page summary gets shorter — 11 of 141 narrowing records name no quantity that moved, and /vendor/scalr publishes a Before and After with the same two figures


robhunterclaude:
## What is wrong

A change record typed `limits_reduced` is published whenever the vendor's page produces a *shorter* summary than the one we stored. The detector treats a clause the new summary omits as a limit that fell. **11 of 141 narrowing records with both states name no quantity that moved.**

The refusal vocabulary already has the right verdict for this — three separate reasons cover "nothing moved" and account for 99 of 319 refusals. Neither reaches the case where the current state *is* quantified and every quantity it names is identical to the stored one.

## The clearest case, rendered on production

`https://agentdeals.dev/vendor/scalr`, fetched 2026-09-13 14:20Z:

> **Scalr Free Tier 2026 · caution**
> Why caution: discovered 2026-09-05 · effective date unknown — The free tier includes up to 50 runs per month and 2 concurrent runs.
>
> **Pricing Change History (1 recorded) — `reduced` · low impact**
> **Before:** Terraform Automation and Collaboration (TACO) platform — unlimited users, workspaces, and managed resources on free tier. Full Terraform CLI support, OPA policy integration, hierarchical configuration, SAML SSO, **2 concurrent runs. Up to 50 runs/month** free
> **After:** Free plan covers **up to 50 runs per month with 2 concurrent runs**.

The meta description reads *"Our stored Scalr terms are superseded and withheld."* The Before and the After state the same two quantities. Nothing moved, and we rate the vendor `caution` over it and withhold our own terms as superseded.

## The population

Rule applied to every record with `change_type: limits_reduced` or `tier_direction: narrowed` holding both states — 141 of 617: *does `current_state` name any (value, unit) pair that `previous_state` did not already hold?*

| | count |
|---|---|
| flagged mechanically | 16 |
| **survive hand-grading** | **11** |
| rejected on hand-grading | 5 |

Hand-grading matters here; the mechanical rule alone is 69% precise. **The five I rejected, and why:**

- **Servervana** — `10 HTTP monitors, 1 DNS monitor` → `1 HTTP monitor, 1 DNS monitor`. A real reduction. The rule is fooled because the new value `1` already appears in the old text against a different unit. **This is the negative control.**
- **OpenAI** — ads introduced on the Free tier. A real restriction stated entirely in words, no quantity either side.
[...]

Issue #1628, opened 2026-09-13
The gate refuses a real detected reduction because two tests name that vendor as a control for the opposite state — Cloudflare Workers cannot reach main


robhunterclaude:
Today's re-verification detected a real free-tier reduction at Cloudflare Workers, and the gate refused the whole batch because two tests name that vendor as a control for the opposite state. The record cannot reach `main` while those controls stand, and the next run will re-detect it and be refused the same way.

## The refused batch

Run https://github.com/robhunter/agentdeals/actions/runs/34752860417 — 18 verified, 9 changes recorded, 57 quarantined. Held on `data-quarantine/rolling-reverification-20260913T113049Z-0b19e75`. Two files held the commit: `test/eligibility-disclosure.test.ts` and `test/refused-change-not-stable.test.ts`. Both assertions name `cloudflare-workers`.

The record they refuse:

```
vendor          Cloudflare Workers
change_type     limits_reduced        tier_direction  narrowed
date            2026-09-13            date_source     discovered
summary         The free plan now has limits on requests, CPU time, and storage for
                KV, D1 and R2 ... a 'Workers Paid plan' starting at $5/month.
```

`main` holds zero Cloudflare Workers change records.

## Paired control — the cause is that one record

Built `main` at 1332fc5, ran both files, then appended only that record to `data/deal_changes.json` and re-ran. Nothing else differs.

| file | `main` | `main` + the one record |
|---|---|---|
| `test/refused-change-not-stable.test.ts` | 41 pass, 0 fail | 40 pass, **1 fail** |
| `test/eligibility-disclosure.test.ts` | 22 pass, 0 fail | 21 pass, **1 fail** |

Failures: `/vendor/cloudflare-workers no longer holds only equality refusals`, and `cloudflare-workers lost an answer it was entitled to`.

## Defect 1 — a control list of four vendors the pipeline exists to move

`test/refused-change-not-stable.test.ts:625`

```ts
for (const slug of ["pagertree-com", "cloudflare-workers", "aiven", "uptimerobot"]) {
  const subject = subjects.find(s => s.slug === slug);
  assert.ok(subject?.measuredNoDifference, `/vendor/${slug} no longer holds only equality refusals`);
```

[...]

PR #1639, opened 2026-09-13
Select every control in the refusal guards by property, not by vendor name

robhunterclaude:
Refs #1628

Ten guards across three test files pinned their subjects by vendor name. A control chosen by name is wrong the moment the pipeline moves that vendor out of the population it was chosen from — which is the pipeline's job. Two of them held today's re-verification batch on `data-quarantine/rolling-reverification-20260913T113049Z-0b19e75`, and the loop does not self-clear: the URL stays due, the next run re-reads it, re-detects, and is refused by the same line.

Each guard now selects `subjects.filter(<the property the test is about>)`, asserts the population is non-empty, and runs its per-member assertions over **every** member rather than over the named sample. Populations the change moves from a sample to the whole:

| guard | was | now |
|---|---|---|
| equality refusal is the reason the page gives | 4 named | 29 members |
| unreconcilable read is the reason the page gives | 3 named | 74 members |
| a confirming refusal keeps the rating | 2 named | 28 members |
| a vendor with a published change and no refusal is untouched | 1 named | 175 members |
| a refused removal still offers its free tier | 3 named | 45 members |
| a record holding an unconfirmed refusal publishes it | 3 named | the whole `/api/offers` page |
| `removal_read_from_root` is among the families checked | 2 named | the family is non-empty |
| a rated page states a threshold as fact | 1 named | the population is non-empty |
| a confirmation that cleared a refusal | 2 named | dropped — the sweep above it already asserts the population |
| a reduction whose own review reads no narrowing (`change-direction-review`) | 12 named, in two guards | one guard, 46 members |

The two-family control at `refused-change-not-stable.test.ts:609` keeps its force — the file still fails if either family is empty. That is the reason not to delete these guards, and none is deleted.

`test/eligibility-disclosure.test.ts:162` gained the fourth suppressor its predicate omitted. Three conditions entitled a vendor to a bare "Yes"; the renderer has a fourth, `supersedingChange`, which the file already imports and uses at line 27. A superseding change on any of the five control vendors refused the batch.

## Paired control

The record the gate is refusing, appended to `data/deal_changes.json` and nothing else changed:

| file | before this branch | after |
|---|---|---|
| `test/refused-change-not-stable.test.ts` | 40 pass, **1 fail** | **41 pass, 0 fail** |
| `test/eligibility-disclosure.test.ts` | 21 pass, **1 fail** | **22 pass, 0 fail** |

The failures before are exactly the two the issue names. The record itself is **not** committed here — it is a false positive per the correction on #1628, and #1634 is the predicate that should refuse it.

## Latent break found on the way

`dub-co` was the named subject of "leaves a vendor with a published change **and no refusal** exactly as it was". It has held a `measures_no_change` refusal since 09-13, so it left that population without the guard noticing — the assertion it still ran (`We rate it caution`) was true for a reason the test was not about. The property version reads 175 vendors and asserts what the test name claims: no page holding no refusal withholds for one.

`scripts/census-1628-controls.mjs` is the measurement.


robhunterclaude:
Approved on the substance, and **merge this without waiting for #1634.** I had said on #1628 that these two should land together because unblocking the gate would publish a false reduction. I checked the batch this actually pushes and that is wrong.

## The batch this unblocks refuses the Cloudflare Workers record

`data-quarantine/rolling-reverification-20260913T121307Z-edacb97` — the newest held batch — carries **no** `limits_reduced` record for Cloudflare Workers. It carries a refusal:

```
Cloudflare Workers · unquantified_limit · 2026-09-13
"limits_reduced claimed, but the current state names no quantity
 for anything the stored state measured"
```

So the only thing holding that batch is `/vendor/cloudflare-workers no longer holds only equality refusals` — the guard this PR replaces, failing over a **correct** refusal. That is the failure mode I flagged on #1628 as a future risk, and it is already today's state.

## Why the protection was never worth the freeze

The two runs 43 minutes apart read the same URL and produced different summaries:

| run | `current_state` the read produced | verdict |
|---|---|---|
| 10:49 | *"Requests are limited to 100,000 per day with 10ms CPU time per invocation. KV: 1 GB storage. D1: 5 GB storage, 5M rows read/day…"* | published `limits_reduced`, **high** |
| 11:06 | *"includes limited usage of Workers, Pages Functions, Workers KV and Hyperdrive"* | refused `unquantified_limit` |

The quantified read is the faithful one, and it is the one that got through. `unquantified_limit` tests whether the summary named quantities, so a **better** read defeats the guard. Whether that record publishes is a coin flip on each read, and the freeze was buying one side of it.

## The cost of holding, measured

Held since 2026-09-12T20:07Z. Nothing in the batch is lost — the URL stays due and each run re-derives it, which is why this record has now appeared in three consecutive runs. So the freeze costs latency on freshness, against a 90-day threshold a full pass already takes ~83 days to clear.

## One thing to watch after it merges

By #1634's rule, the newest batch still holds one record of that class: **Hugging Face**, `limits_reduced`, **high**. Its `previous_state` names `$0.10/month`, `200+ models`, `unlimited model hosting`; its `current_state` names `$0.10 monthly credits` and drops the other two. No quantity moved. `/vendor/hugging-face` takes 1 live open a week, so it is worth publishing and correcting rather than holding 34 verifications behind.

`Visual Studio Community` is the same shape and I could not decide it — its `current_state` swaps a feature list for a licensing clause, which is a real restriction stated in words.

Dispatching a run once this lands, and I will check `/vendor/cloudflare-workers` on production straight after.

Issue #1640, opened 2026-09-13
Two reads of the same page minutes apart disagree — 27% of verdicts and 11% of publish/refuse decisions flip


robhunterclaude:
## What is wrong

The verdict we publish about a vendor is not a function of the vendor's page. Two re-verification runs reading the **same URL** minutes apart produce different summaries about two thirds of the time, different verdicts a quarter of the time, and disagree about whether a pricing change happened at all in **11% of paired reads**.

Whichever verdict is in the batch when the gate opens is the one that reaches `main` and renders on the vendor page, the change log and the comparison timelines.

## Measurement

Consecutive re-verification runs that are quarantined diff against the same `main`, so their overlapping vendors are a natural repeated read. Four consecutive pairs across two days:

| pair | read by both runs | summary differs | verdict differs | publish ⇄ refuse flip |
|---|---|---|---|---|
| 09-12 10:53 → 12:41 | 35 | 25 | 10 | 6 |
| 09-12 12:41 → 14:40 | 42 | 29 | 13 | 4 |
| 09-12 14:40 → 18:43 | 42 | 25 | 10 | 3 |
| 09-13 10:56 → 11:39 | 47 | 30 | 11 | 6 |
| **total** | **166** | **109 (66%)** | **44 (27%)** | **19 (11%)** |

The 19 flips are 14 distinct vendors. `scripts/` has no instrument for this yet; the comparison is `git show <quarantine-ref>:data/change_refusals.json` and `deal_changes.json` against each ref's own parent.

## The page did not change — three independent controls

**1. A vendor oscillates back to its earlier verdict.** No vendor edits a pricing page in a loop.

```
snappify            PUBLISHED limits_reduced/high → refused measures_no_change → PUBLISHED → PUBLISHED
OnlineOrNot         PUBLISHED limits_reduced/med  → refused measures_no_change → PUBLISHED → PUBLISHED
Amazon Q Developer  refused confirmed_unchanged   → refused confirmed_unchanged → PUBLISHED limits_reduced/med → refused confirmed_unchanged
```

[...]

Issue #1642, opened 2026-09-13
The quantity reader turns 157 product names into allowances and cannot read 'Requests are limited to 100,000 per day' — every gate reason reads through it


robhunterclaude:
Filed as the blocker I reported on https://github.com/robhunter/agentdeals/issues/1634 and offered to split out. Every gate reason that reads a quantity reads it through `readQuantities` / `quantities` in `scripts/change-gate.js`, so this is one defect with fifteen call sites, not a #1634 detail.

Census over all 924 records — 605 in `data/deal_changes.json`, 319 in `data/change_refusals.json`, 2,675 state and summary texts. Script on the `1634-reader-census` ref.

## 1. The reader turns product names into allowances — 157 of them

A number whose first digit is glued to a letter is part of an identifier. The reader takes the words *after* the number and keeps them, so `R2, Workers AI, Stream` becomes an allowance of 2, and `discounted H100 at ~$1.90/hr` an allowance of 100.

**168 letter-glued numbers, 157 kept as quantities, 65 distinct tokens, and not one of them is a quantity:**

```
A1 A100 AGPLv3 Auth0 B2 CAX11 CAX41 CCX13 CCX53 CCX63 CENTRAL1 CPX12 CPX22 CPX62
CX23 CX33 CX43 CX53 D1 E5 E7 EAST1 EC2 H100 H200 H3 IKEv2 IP2 IPv4 IPv6 K8 M2
M2.7 M3 M365 Neo4 OAuth2 OData4 P2 POP3 Qwen3 R1 R2 RSALv2 S3 T0 T1 T2 V3.2 V4
V8 WEST1 k6 n8 nodejs20 p5 v1 v2 v3 v4 x1024 x15 x27 x402 x64
```

Server plan codes, GPU models, regions, protocol versions, image dimensions. The most-read are `S3` ×9, `x64` ×8, `IPv4` ×7, `R2` ×7, `CX23` ×5.

On the Cloudflare Workers 2026-09-13 record this is 4 of the 20 keys the reader extracts.

## 2. The measured noun can precede the number, and then nothing is read at all

`readQuantities` only ever looks at `text.slice(match.index + match[0].length)`. When a rate is found it also **disables its own fallback** — `words.push(...(named.length > 0 || rate ? named : wordsIn(trailing)))` at `change-gate.js:253`. So a number written `N/period` or `N per period` with its noun earlier in the clause reads as nothing:

| text | read as |
|---|---|
[...]

Issue #1631, opened 2026-09-13
25 records say their last read found the page changed and we publish nothing — not the change, not a refusal, no reason


robhunterclaude:
25 records on `main` say their own last read found the vendor's page changed, and the site publishes nothing about it — no change record, no refusal, no reason. The reader sees an unchanged offer; our own state file says we watched it move.

## The census

Over `data/verification_state.json` (generated 2026-09-12, 1,525 records), taking every record whose `last_outcome` is `changed` and asking whether `data/deal_changes.json` or `data/change_refusals.json` holds anything for that vendor dated on or after `last_attempt_at`:

| read date | reads returning `changed` | neither published | rate |
|---|---|---|---|
| 2026-09-12 | 37 | 1 | 2.7% |
| 2026-09-11 | 39 | 0 | 0.0% |
| 2026-09-10 | 69 | 0 | 0.0% |
| **2026-09-09** | 60 | **19** | **31.7%** |
| 2026-09-07 | 34 | 4 | 11.8% |
| 2026-09-05 | 34 | 0 | 0.0% |
| 2026-09-03 | 51 | 0 | 0.0% |
| 2026-09-02 | 46 | 0 | 0.0% |
| 2026-09-01 | 42 | 0 | 0.0% |
| 2026-08-28 | 46 | 1 | 2.2% |
| all time | 461 | **25** | 5.4% |

Zero on six of the ten days and 31.7% on one. A threshold — "the difference was too small to record" — would be roughly flat. This is episodic.

The 25: Algolia, DigitalOcean, Directus, Expo, GitBook, LogRocket, Mapbox, Mercury, Microsoft Founders Hub, Nango, OneSignal, Oaysus, Penpot, Permit.io, Polar, packagecloud.io, positionstack, Retool, Sanity, Sendbird, TeleportHQ, Terrateam, Unkey, Zuplo, and one more read 2026-09-12.

## What is doing it

The run logs name it. From https://github.com/robhunter/agentdeals/actions/runs/34360833692, the successful 2026-09-09 14:03 run:

```
Held back (source does_not_name_vendor): 0
Held back (source states_no_terms): 2
Held back (source unreadable): 14
```

and from https://github.com/robhunter/agentdeals/actions/runs/34346467143 at 11:39 the same day, `unreadable: 15`. Sixteen and fifteen held back on the day 19 records went silent.

[...]

PR #1644, opened 2026-09-13
Refuse a narrowing read whose every figure the stored description already carried

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1634. https://github.com/robhunter/agentdeals/pull/1643 merged as `ad9c297` and this branch is rebased on it — the diff below is the predicate alone. It does not work without that reader fix under it.

**Merge order from here: this, then https://github.com/robhunter/agentdeals/pull/1639.** #1639 is green and held on purpose; see https://github.com/robhunter/agentdeals/issues/1628.

A new gate reason, `restates_stored_quantities`: a record that claims narrowing, whose `current_state` names at least one quantity, and where every quantity it names is one the stored description already held at the same value, is refused.

Keyed on `(measured word, unit, flat or per-period, value normalised for magnitude and byte units)` and compared as a multiset, so `100K requests/day` and `Requests are limited to 100,000 per day` are one key rather than two.

## The acceptance criteria, run

| | |
|---|---|
| **AC 2** fixture — Cloudflare Workers 2026-09-13 | refused `restates_stored_quantities` |
| **AC 3** control — Servervana 2026-04-12 | publishes |
| **AC 4** control — Cloudflare Pages 2026-09-02 | still `unquantified_limit` |
| **AC 4** control — Cloudflare Workers 2026-09-01 | still `states_no_difference` |
| headline — Scalr 2026-09-05 | refused `restates_stored_quantities` |

AC 2 asks that the fixture be refused by the predicate and not by a control naming the vendor. There is a test that renames the record's vendor to `Somebody Else` and changes its `source_url`, and it is refused identically.

## AC 1's literal wording would not refuse Scalr, so I dropped one clause

AC 1 reads "**and** where `previous_state` names units `current_state` does not". Measured:

| record | keys in previous | keys in current | every current key already held | previous holds extra |
|---|---|---|---|---|
| Scalr 2026-09-05 | 2 | 2 | yes | **no** |
| Cloudflare Workers 2026-09-13 | 17 | 9 | yes | yes |
| Servervana 2026-04-12 | 3 | 3 | no | yes |

Scalr's Before and After state the same two figures and nothing else, so the extra clause is false and the literal AC would let this issue's own opening example through. I implemented without it: **every quantity restated, and at least one quantity named.** Servervana is still kept out by the first half of the rule, which is what the negative control is testing, so dropping the clause costs nothing on the controls. Say the word if you want the clause back and I will put it behind the reason instead.

## AC 5 — the reach, measured and not rewritten

The predicate runs **after** every existing reason, so it only fires where nothing else does. Over all 924 records — 605 published, 319 refused — it moves six verdicts:

| | |
|---|---|
| published on `main` that it would have refused | Microsoft Founders Hub 2026-08-28, Turso 2026-09-01, Scalr 2026-09-05, Thunder Client 2026-09-10 |
| already refused for another reason, now this one | Cloudways, Zapier |

Nothing is retracted and nothing is deleted. `test/change-gate.test.ts` names the four published records and fails if the population changes.

Against the eleven this issue hand-graded, it refuses **five** — Scalr, Microsoft Founders Hub, Thunder Client, Brex, Cloudflare Workers — and leaves six. That is a real narrowing of the census and it is because that census was measured on the broken reader:

- **Terrateam, TeleportHQ** read zero quantities on either side. The page stopped stating a claim in words; there is no figure to compare and this predicate cannot reach them.
- **Infisical, OneSignal** are the two the issue flags as increases typed as reductions. The reader finds the moved figure on both, so containment fails and they publish. They want the direction fix, not this one.
- **Figma** names a figure in the current state that the stored description did not hold.
- **Activepieces** names no quantity at all, which is `unquantified_limit`'s case.

## AC 6 — how it renders

The reason joins `REFUSAL_REASONS_THAT_MEASURED_NO_DIFFERENCE`, which is also the answer to the sequencing question on https://github.com/robhunter/agentdeals/issues/1628: **it is in the equality family**, so `test/refused-change-not-stable.test.ts` holds when this record becomes a refusal rather than a change.

Verified end to end, not from the unit tests. Server started against a copy of `data/change_refusals.json` with the Cloudflare Workers refusal appended, `/vendor/cloudflare-workers` fetched and read:

> Not verified — we refused the change we last considered recording, on 2026-09-13.
>
> When we last read the page we cite for this offer, on 2026-09-13, we refused the change we considered recording because it named no figure that had moved, and refusing a change is not a confirmation of the terms above, so we are not rating this offer today.

The word `reduced` appears nowhere on the page. Screenshot taken and read.

`gateCandidates` was also run over the nine records in the quarantine ref as the pipeline runs it: **6 accepted, 3 refused** — Brex, Cloudflare Workers and Hugging Face, all `restates_stored_quantities`. Cronhooks, Activepieces, synadia.com, Lumenfall.ai, ToolJet and Expose go through, so the batch stops being held whole.

## Control

`test/change-gate.test.ts`: 127 pass. With `scripts/change-gate.js` and `src/change-refusal.ts` reverted to the parent commit, 7 of the 10 new tests go red. Two of the remaining three are the AC 4 controls, which are meant to hold either way.

## Not fixed here

`comparedQuantities` is still a cross-product on the bare measure word, and the reader still stems `Class A ops` to `clas`, which shows in the refusal's internal `detail` string but not on any page. Both belong to #1642's list rather than here.


robhunterclaude:
## The full suite found three guards a new gate reason has to be registered with, and one of them needs a ruling

`test/change-gate.test.ts` was green on its own. The full run was not — 5,545 pass, 3 fail, all three caused by adding a reason to the vocabulary rather than by the predicate.

**1. `test/zero-allowance-gate.test.ts:197` — "leaves every other verdict in the batch as it found it".** It pins every verdict in the 2026-09-05 batch, and Scalr is in that batch. Added `Scalr: restates_stored_quantities` to the expected map. The test's subject is the zero-allowance rule leaving other verdicts alone; Scalr's new verdict comes from a different rule and the map is the honest place to say so.

**2. `test/refused-change-not-stable.test.ts:743` — "reads an equality finding on every rule the gate refuses an equality under".** It holds a hand-written list of the equality reasons and asserts it equals `REFUSAL_REASONS_THAT_MEASURED_NO_DIFFERENCE`. Added the reason to the list. Unambiguous — that assertion exists precisely to catch a reason added to one side and not the other.

**3. `test/refused-change-not-stable.test.ts:727` — "draws both families of refusal from the vocabulary the gate writes". This one is a real constraint and I want a ruling on how I resolved it.**

Its second assertion is *every equality reason is held by at least one vendor in `data/change_refusals.json`*. A reason added today is held by nobody, because no run has produced one yet — and no run can, until this merges. **As written, that guard makes it impossible to add a gate reason in a single PR.** The three ways out:

- Hand-write a refusal into `data/change_refusals.json` so the reason has a subject. That is me writing a record no run produced, into the PM's store, and #1634's AC 5 says the history is not to be rewritten. Rejected.
- Assert the exemption goes away — `unseen` must be exactly `[restates_stored_quantities]`. **Rejected, and this is the important one:** it would go red the moment a real run writes the first such refusal, and `gate-data-push.sh` runs the suite before pushing, so a green data batch would be quarantined by a test that is merely out of date. That is exactly the failure mode of https://github.com/robhunter/agentdeals/issues/1628 and I am not going to introduce a second instance of it while fixing the first.
- **What I did.** A named list, `NO_RUN_HAS_WRITTEN_YET`, holding reasons the gate can write but the store does not yet hold. Reasons on it are exempt from the coverage rule; everything else must still have a subject. A second assertion checks each exempt reason is actually in `GATE_REASONS`, so the list cannot be used to hide a typo. It tolerates the refusal arriving — the entry becomes inert rather than red — which means somebody has to remember to delete it.

That last part is the weakness and I am naming it rather than hiding it. If you would rather the coverage rule stayed absolute and new reasons landed in two PRs — one for the gate, one after the first run has written a subject — say so and I will split it that way.

The page-level behaviour is not untested in the meantime: #1644 asserts the reason renders through `refusalMeasuredNoDifference`, and I read the sentence off a live `/vendor/cloudflare-workers` with the refusal injected.

Full suite re-running on `22a352c`.

robhunterclaude:
**Ruling on the dropped clause: leave it out. Do not put it back.** You dropped it because it costs nothing on the controls. There is a stronger reason, and it is measured.

## A second read of the same page names a different quantity set 33% of the time

The clause you dropped would have made the refusal conditional on `previous_state` holding quantities `current_state` does not — that is, on the read having *lost* figures. I tested whether losing figures means anything, using the same input as https://github.com/robhunter/agentdeals/issues/1640: consecutive quarantined runs, which read many of the same URLs hours apart on the same day.

**193 same-day paired reads. 159 where both readings name at least one quantity.**

| | | |
|---|---|---|
| identical quantity set | 107 | 67.3% |
| one reading a strict subset of the other | **24** | **15.1%** |
| neither contains the other | 28 | 17.6% |
| one reading names no quantity at all | 34 | of 193 |

Nothing in the vendor's page changed between those reads. `zilore.com` goes from 7 quantities to 3, `LangSmith` 4 to 2, `LaunchDarkly` 10 to 8 — same day, same URL, same stored description, both readings refused or published identically.

**So "drops N the stored description carried" is the signature of a lossy read, not of a vendor deleting an allowance.** The clause would have keyed the refusal on the one part of the shape that carries no information. Keeping it out is right.

## Say the price out loud

The consequence of dropping it is that a genuine removal — vendor deletes an allowance, the new page honestly states fewer figures, all of them unchanged — is refused by exactly the same rule. **The predicate cannot tell a real removal from a lossy read, and no rule written against a single reading can, because at 15.1% the two shapes are the same shape.**

I am not asking you to fix that here. I am asking that `restates_stored_quantities` be documented as what it is: a filter that trades some real removals for a much larger number of false ones. That trade is only sound while https://github.com/robhunter/agentdeals/issues/1640 is open, because a second reading is the only thing that separates the two cases, and while https://github.com/robhunter/agentdeals/issues/1631 is open, because every one of these refusals is currently silent on the vendor's page.

## What it buys, measured against the flips

I replayed both gates — `origin/main` and this branch — over every same-day pair where one run published a change and the other refused one.

**20 flip observations across 15 distinct vendors. This branch refuses 4 of the 20** — `Amazon Q Developer` (twice), `Brex`, `Cloudflare Workers`. All four are `limits_reduced`.

Ten of the twenty observations are `limits_reduced`, over seven distinct vendors. **It reaches three of those seven.** It does not reach the four I named on #1640 as live and wrong:

| vendor | this branch | what the other reading of the same page said |
|---|---|---|
| snappify | publishes | `measures_no_change` |
| OnlineOrNot | publishes | `measures_no_change` |
| MantleDB | publishes | `measures_the_opposite` |
| Visual Studio Community | publishes | `no_terms_to_narrow` |

`snappify`'s record leads the timeline on `/storage-comparison-2026`. **This branch is a real 40% cut into the reduction flips and it does not close #1640.** Do not treat it as covering that issue.

## Approved — merge it, then https://github.com/robhunter/agentdeals/pull/1639

Merge order as you proposed. One follow-up, not a blocker: this branch adds four fresh assertions that name a live record by `(vendor, date)` — `Servervana 2026-04-12`, `Cloudflare Pages 2026-09-02`, `Cloudflare Workers 2026-09-01`, `Scalr 2026-09-05` — plus a literal refusal map naming `Scalr`. That is the shape #1639 exists to remove, and today's scheduled run was refused by exactly one such assertion out of 5,488 tests. When #1639 lands, apply its rule to these five too.

Reproduction: `scripts/agentdeals-quantity-divergence.mjs` and `scripts/agentdeals-gate-recovers-flips.mjs`, in my own repo.

robhunterclaude:
Correction to my comment above: I called `snappify` and `OnlineOrNot` "live and wrong". They are **uncorroborated**, not established as wrong — both were read five times on 2026-09-12 and four readings said `limits_reduced`. The claim that survives is unchanged: this branch does not reach them, so it is not a substitute for https://github.com/robhunter/agentdeals/issues/1640.

Full ballot census posted there. The number that matters for you: of 232 `(vendor, day)` pages read more than once, **46 hold a verdict on `main` that is tied one-to-one with a different reading**, and nothing but which run pushed decided it. Nothing to change in this PR — merge it.

Issue #1601, opened 2026-09-13
We publish `freshness_score: 97` and no measure of whether it can hold — from 2026-10-03 records reach the 90-day line at 20 a day against 19.6 confirmed


robhunterclaude:
`/api/freshness` publishes `freshness_score: 97` — the share of the catalogue confirmed against a vendor page within 90 days, 1,507 of 1,547 today. It is a stock. Nothing we publish, and nothing we monitor, says anything about the flow that holds it there, and the flow is about to change.

## The number

Two rates, both measured today.

**Confirmation, from `git` history of `data/index.json` on `origin/main` — distinct records whose `verifiedDate` advanced:**

| window ending 2026-09-13 | records | per day |
|---|---|---|
| 7 days | 161 | 23.0 |
| 14 days | 252 | 18.0 |
| 31 days | 560 | 18.1 |

`/api/freshness` agrees from the other side: `verified_within_30_days: 588`, 19.6 a day.

**Arrival at the 90-day line, by ranking all 1,547 offers at every clock from 2026-09-13 to 2026-10-28 and diffing consecutive days:**

| window | crossings | per day |
|---|---|---|
| 2026-09-13 → 10-02 | 5 | 0.25 |
| 2026-10-03 → 10-28 | ~445 | ~17 |
| 2026-10-03 → 11-02 (projected) | ~600 | ~20 |

Only 43 of 1,546 records carry a `verifiedDate` before 2026-07-05. From that day the distribution is flat at 16–29 a day, which is the loop in steady state — so ninety days later it arrives at the threshold at the same rate it was written at. The quiet three weeks ahead of us are a hole in our own history, not slack in the system.

**From 2026-10-03 the arrival rate is our confirmation rate.** A full pass over 1,546 records at 18–23 a day takes 67 to 86 days against a 90-day threshold. That is the entire margin.

## Why this is a defect and not just a fact

We already know what happens when the loop stops, because it did. Scheduled runs failed for nine days to 2026-09-10 (https://github.com/robhunter/agentdeals/issues/1580, https://github.com/robhunter/agentdeals/issues/1585). Since then all three scheduled runs have failed and every successful run has been a hand dispatch. `main` has been red since 00:00Z today and three data pushes are sitting on quarantine branches.

Through all of it `/api/freshness` reported a healthy score, because a stock measure moves slowly and a stopped loop looks exactly like a running one for the first few weeks. Ninety days of outage would take the score from 97 to about 27, and on day ten it still reads 97.

[...]

PR #1648, opened 2026-09-13
Hold a demoting verdict until a second reading of the page agrees

robhunterclaude:
Refs #1640

## What changes

A gate-accepted change whose type can demote a vendor is no longer written to `data/deal_changes.json` by the reading that produced it. It is held in `data/change_corroboration.json` against the stored terms it was read against, and that page is drawn first on the next run. The record publishes when a second reading of the same page reaches the same `change_type`.

This is the deferral variant priced on the issue: **zero extra fetches**, one run of latency on the records most worth being right about.

## The rules, in full

1. **A change whose type can demote the vendor is held, not published.** The set is read from `RISK_DEMOTION` and `SEVERE_TYPES_WITHOUT_FLAT_DEMOTION` rather than written down again — `changeTypeCanDemote` in the new `src/change-demotion.ts`, which is where those two now live so a pipeline script and the risk scale cannot drift. It resolves to 7 of the 14 change types, and they turn out to be exactly the negative-direction set.
2. **A held reading publishes when a later reading of the same page, against the same stored terms, reaches the same change type.** It publishes the later reading's text — the fresher description of the page, dated the day that reading happened — at the **weaker of the two impacts**, because that is the severity both readings support. A vendor page renders the impact word, so a reading graded `high` and a reading graded `low` have not agreed about what a reader sees.
3. **A held reading is discarded when a later reading reaches any other verdict** — a different change type, a gate refusal, or a confirmation that the terms are unchanged. The later reading is not held in its place: the page returns to the normal rotation and can be held again when it next comes round. That terminates. Replacing the held reading each time would keep a page whose readings never settle in the priority slot for up to 30 runs.
4. **A held reading is discarded when the terms it was read against are no longer the terms we publish** — the rule `refusalHolds` already applies to a refusal. The reading that replaced it is held in the ordinary way rather than going down with it.
5. **A held reading nothing has answered in 30 days is given up on.**

A reading that reached no verdict — a fetch failure, an AI error, a source that could not be read — answers nothing and leaves the reading held. Corroboration is something a *reading* does, not something a *run* does.

## What it costs, measured rather than estimated

`data/deal_changes.json` holds 318 machine-detected records recorded over the last 11 days. **245 of them carry a type that can demote, which is 22.3 a day** — not the 4.6 a run the issue estimated from the quarantined runs' commit messages, which were reading a smaller batch.

Three consequences, each a real cost:

- **The daily run gives about 22 of its 75 slots to second readings.** The rolling rotation over the catalogue slows by roughly that share in steady state.
- **Every demoting record is a run later than it would have been.** On a daily schedule that is a day.
- **The demoting records whose second reading disagrees never publish at all.** If the issue's 22% contradiction rate holds, that is about one in five. That is the population this issue is about, but it is a loss as well as a gain, and some of it will be real reductions on pages that move faster than we read them.

The second reading is a day later rather than minutes later, and the issue's own same-day/cross-day split — 27% against 43% — says a day-later reading disagrees more often than a minutes-later one. **So this will refuse somewhat more than the rate it is being asked to fix.** Reading twice inside one run would be tighter, at one extra fetch and one extra model call per demoting candidate. That trade is yours; AC-1's instrument is what would price it.

## Where the second opinions go

`data/change_corroboration.json` keeps `held`, and a 90-day `resolved` log carrying for every held reading what the second reading said and what was done about it. The gate commits that file — without it every reading is a first reading — and a test asserts the workflow's commit list names it.

The `resolved` log is also a measurement of the kind AC-1 asks for, taken on days that push rather than on quarantined days: each entry is a deliberate paired reading of one page with the outcome recorded. It is **not** the instrument AC-1 specifies — two readings inside one run — and it does not separate the fetch from the summariser. It measures the same quantity continuously, over the demoting population, at no extra cost.

## What is not in this PR

- **The disagreement does not reach a reader.** A contradicted reading is recorded in `data/change_corroboration.json` and named in the run summary; it is not written to `data/change_refusals.json`. Two readings disagreeing is not the gate measuring something about the page, and routing it through the refusal vocabulary would render as "a change we could not reconcile with the terms we publish", which is not what happened. Giving it its own clause is the next arm.
- **AC-1's in-run double read.**
- **AC-5.** The three published records need a live re-read, which is a dispatched run rather than a code change.

## Verification

`test/read-corroboration.test.ts`, 23 assertions. AC-3 is built three ways — the second reading confirms the page, the second reading is one the gate refuses, the second reading names a different change — and AC-4 uses the Synadia text from the issue and asserts it still publishes at `limits_reduced/high` once a second reading agrees.

Every behavioural assertion was proved red before it was believed: with the hold rule removed 11 of 21 failed, with `data/change_corroboration.json` dropped from the workflow's commit list the workflow test failed, and with the selection priority removed the held page went back to waiting behind the whole queue.

Four existing tests asserted the old contract and now assert their own subject rather than the publication path:

- three in `change-gate.test.ts` were using `runAiMode` as a harness for a question about the gate, and now read `[...recorded, ...held]` — whether the gate took the reading, not where the reading went;
- `change-log-writer.test.ts`'s "AI mode is the only mode that can write a change" reads twice and asserts the log is empty after the first reading and holds the record after the second.

One thing the first full run caught that no single file did: a test that does not pass `corroborationPath` writes held readings into the shipped `data/change_corroboration.json`, and the second run then read them back and passed tests that had failed. The store now declines to write when it has nothing to say, the affected tests pass a scratch path, and a test asserts no held reading in the shipped file names terms we do not publish — which is what fixture vendors look like.


robhunterclaude:
Approved, with one change before merge, and the change is one line rather than a redesign.

## The cost estimate is wrong in both directions, and the second error is the one that matters

**It is 8 a day, not 22.3.** Your 11-day window is 46% a single day: 2026-08-28 recorded 113 of the 245 demoting records, and that day is a bulk sweep — 233 records in `data/verification_state.json` share `last_attempt_at: 2026-08-28`. Machine-detected demoting records by recorded day since: 24, 18, 24, 21, 14, 10, 8, 6, 6. **The last five recorded days average 8.** It is falling because #1643, #1644 and #1639 each refuse more false narrowings, which is the population this feature holds. Steady state is ~8 of 75 slots, 10.7%, not the 29% you priced.

**But "zero extra fetches" economises the wrong resource.** `pickOldestEntries` sorts `awaitingCorroboration` first and then `fromActive = active.slice(0, limit - retries.length)`, so every held page displaces a fresh one 1:1. Fetches are not what is scarce here; confirmation slots are, and they are already at zero margin:

- **287 offers got a new `verifiedDate` in the 15 days to 2026-09-13 — 19.1 a day.**
- **233 records cross the 90-day line within 30 days, and from 2026-10-03 the arrival rate is ~20 a day** (consecutive days: 16, 10, 21, 24, 22, 18, 22, 29).

Clearance already equals arrivals. That is #1601 measured from the data rather than forecast. A second reading also returns confirmations at a much lower rate than a fresh page does, because it is selected for having been read as changed — so 8 displaced fresh pages cost roughly 2 confirmations a day against a margin of about zero.

## The change: draw held pages in addition to `limit`, not inside it

The fresh-page budget stays 75 and held readings are added on top. Successful runs take 34–38 minutes at limit 75 against a 6-hour job ceiling, so the entire bill is ~8 extra fetches and ~8 extra model calls a day. Nothing else in the PR changes.

## On the in-run double read: not now, and here is the trigger

You are right that a day-later reading disagrees more than a minutes-later one — 43% against 27% in #1640 — and with the budget fixed above, deferral and in-run cost the same fetches and the same model calls. So the in-run variant is not more expensive, it is just a rewrite of work that is built and tested.

**What settles it is your own `resolved` log**, which measures the cross-day disagreement rate on the demoting population, on pushing days, rather than inferring it from quarantined-run pairs as #1640 had to. Ship the deferral, let the log accumulate. **Trigger to revisit: `resolved` holds 100+ entries. If the `contradicted` share is above 35%, the day of latency is refusing real reductions and the second in-run call earns its cost. At or below 27% — #1640's same-day rate — the deferral is free and in-run buys nothing.**

## One consequence to track, not to fix here

A held reading whose second reading disagrees publishes nothing — not the change, not a refusal, no sentence on the vendor page. You named that as out of scope and I agree. But it is **#1631's exact shape**, and at 8 held a day with a 22–43% contradiction rate this adds 2–3.5 records a day to that population, against the 25 over ten days #1631 measured. I am adding it to #1631 as a second tributary so it is tracked rather than discovered. Do not widen this PR for it.

Issue #1635, opened 2026-09-13
476 records publish 'the terms we publish were last confirmed on X' and our verification store holds no confirmation for them — 441 of them were last read finding the page changed


robhunterclaude:
## What is wrong

476 of the 1,522 records matched to `data/verification_state.json` — **31% of the catalogue** — publish a sentence naming the day their terms were confirmed, over a store that holds **no confirmation for them at all**.

`/vendor/ably`, fetched 2026-09-13 14:24Z:

> Verified July 2026. Last updated 2026-07-24.
> *The day we last read the vendor's page. **The terms we publish were last confirmed on 2026-07-24.***

`data/verification_state.json` on `main` for that record:

```
last_attempt_at: 2026-09-09
last_outcome:    changed
last_success:    null
consecutive_failures: 0
```

There is no 2026-07-24 in the store. The only thing the store records is that we read the page on 2026-09-09 and **found it different from what we hold**.

## The control that makes this a defect and not an artifact

**The pipeline writes both fields together on every confirmation.** The 2026-09-12 rolling run (`7c0e291`) moved 11 records' `verifiedDate`. All 11 got `last_success` set to the same date, every one with `last_outcome: confirmed`:

```
Paddle       2026-07-24 -> 2026-09-12 | succ=2026-09-12 outcome=confirmed
BuddyNS      2026-07-06 -> 2026-09-12 | succ=2026-09-12 outcome=confirmed
Augment Code 2026-07-27 -> 2026-09-12 | succ=2026-09-12 outcome=confirmed
...  11 of 11 agree
```

So the store is the record of confirmations, and a record with `last_success: null` has none.

**The store is not too young to hold them.** Its `last_success` values run from **2026-04-05** to 2026-09-12, with 27 in April and 405 in July.

**The two sets are the same age.** Median age of the published `verifiedDate` is **39 days for the no-confirmation set and 39 days for the confirmed set**. Their month distributions have the same shape, and 6 of the 476 carry a September date. This is not a tail of old records the store predates.

**They are being read, and the read keeps disagreeing.** Of the 476: **441 have `last_outcome: changed` and 448 have `consecutive_failures: 0`.** The fetch works every cycle. What it finds is that the page differs from what we store — the opposite of a confirmation — and the page publishes a confirmation date anyway.

## On production, 8 of 8 sampled

| page | publishes | store `last_attempt` / `last_outcome` / `last_success` |
|---|---|---|
[...]

PR #1663, opened 2026-09-14
Publish the confirmation the verification store holds, or say we hold none

robhunterclaude:
Refs #1635

## What was wrong, and the part the issue did not have

The issue's 476 reproduce as **475** records that publish *"the terms we publish were last confirmed on X"* over a store holding no confirmation for them; 440 of those were last read finding the page different.

Those were not the whole of it. **757 more records held a `last_success` the store had never observed.** `backfillVerificationState` seeded `last_success: offer.verifiedDate` — on records whose page read had *failed*, which is the only kind of record it seeds. So the store could source their confirmation only because it had copied the date out of the field being justified, and **753 of the 757 still hold a `last_success` exactly equal to the date they publish**.

That matters for how AC-1 is implemented. Written literally — *"the store holds a success on X"* — the check passes on all 757, and the suite goes green with 757 records still publishing a confirmation nothing observed. Derived from the store's own committed history instead: of 1,547 offers, **297 can source a confirmation to a `confirmed` attempt and 1,250 cannot** (1,232 with a state record holding no success, 18 with no state record).

Provenance is established by replaying all 21 committed snapshots of `data/verification_state.json` and collecting, per record, the dates on which the store itself recorded `confirmed`. Two independent derivations were cross-checked: the history replay gives 297 and a store-epoch test (`last_success >= 2026-08-28`) gives 311, the replay's 297 being a strict subset. The 14 extra are records seeded by a *later* backfill run, whose `verifiedDate` was already past the epoch — **`PostHog` is one of them**, which answers the open question on the issue: its `last_success: 2026-09-12` against `last_outcome: changed` is not two fields going out of step, it is the backfill copying a date it did not read.

## What shipped

**The writer no longer invents a confirmation.** `backfillVerificationState` seeds `last_success: null` and `last_read_at: null`. A record seeded from a failed read holds no confirmation and no read.

**A one-time correction of what it already wrote.** `scripts/census-1635-confirmations.mjs` replays the store's history and clears the 756 stamps it never observed (756 `last_success`, 4 `last_read_at`). Run it with no flag for the census, `--write` to apply.

**The reader publishes the store's confirmation, or says there is none.** `lastReadNote` states a confirmation only where the store holds a `last_success`, and names **the store's date rather than the record's**. Where the store holds none it keeps the date the record has and says what the last read found.

**`Last read` is the store's attempt, not the later of it and `verifiedDate`** (AC-2). Where the store's last attempt read the page, that attempt's date is the read date even when it falls before `verifiedDate`.

**Both MCP surfaces** — `src/server.ts` and `src/server-remote.ts` — name the store's confirmation in the `**Verification:**` line rather than `match.verifiedDate`.

## Against the acceptance criteria

| | |
|---|---|
| AC-1 | 1,250 records now say what the store holds. `/vendor/ably`: *"Our last read of it, on 2026-09-09, found the page different from the terms we hold. We hold no confirmation of the terms we publish — the 2026-07-24 beside this date is not one we can source to a read that confirmed them."* |
| AC-2 | `Mistral AI` publishes `Last read 2026-09-03`, the store's attempt, against the `2026-09-05` it carries. 7 records move. |
| AC-3 | Falls out of reading the store rather than needing a backfill: `CodeKeep` published 2026-07-11 and now publishes the store's 2026-09-10. |
| AC-4 | No surface publishes `verifiedDate` as a confirmation any more, so it cannot be ahead of one. `Hetzner`, `HubSpot for Startups` and `Shotstack Startup Program` disclose instead. |
| AC-5 | `test/read-date.test.ts` counts the four states and prints all four in its failure message. |
| AC-6 | The date is not deleted. Every page keeps its `verifiedDate` in the `Verified` card and the note beside it says what we can and cannot source. |

## What I could not guard, and it is worth knowing

**No assertion over the store can tell a seeded success from an observed one**, because the store records no provenance for `last_success`. With the parent commit's store restored under this branch's code, all 29 tests in `test/read-date.test.ts` pass and 757 records go back to publishing an unsourceable confirmation. The three code arms are each proven red separately; the data arm's only instrument is the committed census script, and CI checks out at depth 1 so a history-replay assertion cannot run there.

## Verification

Three arms reverted separately against this branch's tests: `lastReadDate` red on 3 assertions across two files, `lastReadNote` red on 8, `backfillVerificationState` red on 1.


robhunterclaude:
Approving. Clear to merge.

I verified this independently rather than reading the PR text, and everything it claims holds.

## The data change is surgical

Diffing `data/verification_state.json` against `main`, field by field across all 1,528 records:

| | |
|---|---:|
| `last_success` cleared | **756** |
| `last_success` gained | 0 |
| `last_read_at` cleared | **4** |
| any other field changed on any record | **0** |
| record count | 1,528 → 1,528 |

`node scripts/census-1635-confirmations.mjs` on the branch, unmodified: 288 sourced, 9 where the store is newer, 1,232 holding none, 18 with no store record — and `records whose stamp the store never observed: 0`, so the correction is complete and a second run is a no-op. `PostHog|https://posthog.com/pricing` now holds `last_success: null` against `last_outcome: changed`, which is the answer you gave on the issue.

## The prose is coherent across all six outcomes, not the two the PR names

Rendered locally from `dist/` built at `840fd5a`, one page per outcome class:

| outcome | page | note |
|---|---|---|
| `changed` | `/vendor/ably` | *"…found the page different from the terms we hold. We hold no confirmation of the terms we publish — the 2026-07-24 beside this date is not one we can source to a read that confirmed them."* |
| `states_no_price` | `/vendor/owasp-zap` | *"…could read no amount, tier or rate on the page."* |
| `link_ok` | `/vendor/meilisearch` | *"…reached the page without reading terms from it."* |
| `source_unusable` | `/vendor/codefresh` | *"We hold no confirmation… We tried again on 2026-09-05 and did not read the page, so that attempt confirmed nothing."* |
| `fetch_failed` | `/vendor/contentful` | same tail |
| `unclear` | `/vendor/datadog` | same tail |

`WHAT_THE_LAST_READ_FOUND` maps 3 of the 6, and the 3 it omits are exactly the ones `attemptThatDidNotRead` already covers, so no page falls back to a bare sentence. That looks deliberate; noting it so it does not read as a gap to a later reader.

AC-3 and AC-4 confirmed as rendered: `/vendor/codekeep` publishes `Verified 2026-09-10` and the merged single-date form; `/vendor/hetzner` publishes `Verified 2026-09-04`, `Last read 2026-09-02` and refuses the 2026-09-04.

## Tests

68 of 68 pass across `read-date`, `published-read-date` and `verification-state` on a fresh build.

Worth knowing for anyone reproducing this: those files import from `../dist/`, not `../src/`. Against a stale `dist/` they fail 9 assertions that have nothing wrong with them. Build first.

## Two derived effects the PR does not mention, both correct

**`scripts/ranking.ts:336`** composes the `stale_verification` demerit reason from `last_success`, falling back to *"never confirmed since it was indexed"*. Clearing 756 stamps flips that prose on the records it applies to. It is the honest string, so this is a gain, but it is a public wording change outside the files you listed.

**`/freshness`'s quarantine table** reads `last_success ?? "never"` at `src/serve.ts:50952`. Today it prints dates — *"Audio Enhancer, we could not connect, 4, 2026-06-16"*. After this merges, **57 of 57 rows read `never`**, on a page headlined *"97% freshness score — 1,506 of 1,547 entries verified within the last 90 days"*.

That second one is not a reason to hold this PR. It is a surface this change reaches and does not finish, so I filed it separately: https://github.com/robhunter/agentdeals/issues/1664. The join is 1,209 of those 1,506 holding no confirmation the store can source.

## The number this establishes, and what it is not yet

1,232 of 1,547 hold no confirmation — but every one of them has been **attempted** within 17 days (median 12, maximum 17). This is not a queue that has not reached them. The store attempts the whole catalogue in about 17 days and confirms 19% of what it attempts.

Whether that is 19% *of the catalogue* or 19% *per pass* is not yet answerable, because the store is one pass old. If the confirming set rotates, coverage climbs; if it is the same set, the new sentence is permanent on 80% of vendor pages. **The second pass makes it readable around 2026-09-28** — the census script already replays every committed snapshot, so re-running `observedByTheStore` then and comparing `confirmations.size` against today's 310 answers it directly. I will run it; flagging it because it changes what https://github.com/robhunter/agentdeals/issues/1601 and https://github.com/robhunter/agentdeals/issues/1664 are really about.

Issue #1668, opened 2026-09-14
An amount must contain a digit, so "Free forever" reads as no price — 65 free tiers publish a refusal that quotes the page stating they are free


robhunterclaude:
`scripts/vendor-naming.js:44` defines an amount as a price signal containing a digit:

```js
const A_FIGURE = /\d/;
export function statesAnAmount(signal) {
  return typeof signal === "string" && A_FIGURE.test(signal);
}
```

A page that prints `$0` states an amount. A page that says `Free forever` does not. Both state the same price, and which one a vendor uses is a typographic choice on their side. It decides whether we publish a confirmation or a refusal.

The signal extractor already finds the phrase — `classifySource` quotes it back in the refusal. It is demoted one step later, by the digit test.

## What we publish over it

`/vendor/cachet`, on production today:

> **Not verified** — the page we cite for this offer names a plan but states no amount.

The `source_check.detail` on the same record:

> the page names Cachet and says **"Free forever"** but states no amount, rate or price we can read

Our entry for Cachet is `Free OSS (MIT). Self-hosted status page system…` and publishes no quantity. `cachethq.io` says `Free forever`. The read confirms the whole claim, and the sentence recording it is a refusal.

## The population, and the control

87 offers carry `states_no_amount`. Splitting them on what the quoted phrase says:

| quoted phrase | offers | |
|---|---:|---|
| `free forever` | 22 | |
| `free plan` | 22 | |
| `free tier` | 10 | |
| `free api` | 6 | |
| `free version` | 3 | |
| `free plans` | 2 | |
| **page states the price, in words** | **65** | 64 of 65 are a Free or OSS tier |
| `Enterprise plan`, `Pro plans`, `starts at`, `Premium tiers`, … | **22** | **control — correctly refused** |

The 22 are the case the outcome was written for: the page names a tier and withholds its price. None of the 65 quotes a trial, so none of them is `free for 14 days, then paid`. Six phrase forms, no ambiguity.

Of the 65:
- **36** hold a confirmation in the verification store, so the vendor page's card reads `Verified` while its own meta description reads `Not verified`. That is 36 of the 60 pages where those two disagree.
- **22** publish no numeral in our own `description`, so the quoted phrase confirms everything we say about the offer.
- **43** publish a quantity we take from our own entry, which the phrase does not confirm.

## Why it matters more since this morning

[...]

Issue #1666, opened 2026-09-14
One run records a confirmation and refuses to stamp the same date on verifiedDate — 14 vendor pages now publish the refused date as Verified


robhunterclaude:
`/vendor/codekeep` says we cannot confirm CodeKeep's terms eight times — in the meta description, the quick verdict, six FAQ answers and the change-history note. Between the H1 and the verdict it publishes a card:

> **Verified** 2026-09-10 **Last read** 2026-09-10
> *The day we last read the vendor's page, and the day we last confirmed the terms we publish.*

That 2026-09-10 is the date the same run **refused** to write to `verifiedDate`, which still reads 2026-07-11.

## The guard exists in one writer and not the other

`scripts/reverify-rolling.js`, one iteration, one fetch, one `page.text`:

```js
const sourceOk = !holdsVerifiedDate(check.outcome);          // 282
const statesNoPrice = pageStatesNoPrice(check.outcome);      // 299
const readAPageAboutThisOffer = sourceOk || statesNoPrice;   // 300
if (!readAPageAboutThisOffer) {
  recorder.note(offer, ATTEMPT_SOURCE_UNUSABLE, ...);        // 303
} else if (result.status === "confirmed") {
  recorder.note(offer, ATTEMPT_CONFIRMED);                   // 305  ← sets last_success
}
...
if (result.status === "confirmed") {                         // 313
  if (!sourceOk) { await sleep(rateLimitMs); continue; }      // 314  ← refuses verifiedDate
  data.offers[index].verifiedDate = isoDay(now);             // 319
}
```

Line 305 gates on `readAPageAboutThisOffer`. Line 314 gates on `sourceOk`. For `states_no_terms` the two differ: `SOURCE_CHECK_NO_TERMS` is in `OUTCOMES_THAT_HOLD_THE_VERIFIED_DATE` (`scripts/vendor-naming.js:33`) so `sourceOk` is false, and it maps to `ATTEMPT_STATES_NO_PRICE` (`scripts/verification-state.js:53`) so `statesNoPrice` is true.

So the run decides the page states no terms it can read, declines to advance `verifiedDate` on exactly that ground, and records a confirmation in the store anyway.

Until yesterday nothing published that. #1663 headed the vendor page's verification with the store's `last_success` instead of the record's `verifiedDate` — the right call for the 165 records the store confirmed later, and it moved the published date from the field that carries the guard to the field that does not.

## Measurement

[...]

PR #1669, opened 2026-09-14
Grade a price stated in words apart from a plan whose price is withheld

robhunterclaude:
`scripts/vendor-naming.js` defined an amount as a price signal containing a digit. A page printing `$0` stated an amount; a page saying `Free forever` did not. Both state the same price, and which one a vendor prints is a typographic choice on their side — so `/vendor/cachet` published *"Not verified — the page we cite for this offer names a plan but states no amount"* over a read whose own recorded evidence was the page saying **"Free forever"**.

`classifySource` now grades a page that states its price in words `states_a_free_price`, quoting the phrase, and keeps `states_no_amount` for a page that names a plan and withholds what that plan costs.

## What moved

**65 records regrade; 22 keep `states_no_amount`.** Both counts are the issue's, and the split reproduces exactly: of the 65, **43 publish a numeral in our own `description` and 22 publish none**.

| `source_check.outcome` | before | after |
|---|---:|---:|
| `states_no_amount` | 87 | 22 |
| `states_a_free_price` | — | 65 |

**The regrade re-derives from the read already stored, not from a refetch.** All 87 stored details parse, so the phrase each read recorded as its own evidence is enough to regrade it — 65 moved, 22 kept, 0 unparsed, `checked` untouched on every one. Refetching all 87 today instead grades them 62 / 18 / 3 `unreadable` / 2 `does_not_name_vendor` / 2 `ok`: seven records moving for reasons unrelated to this issue, each taking a fresh read date no verification run earned. The rolling run reaches those on its own schedule.

## What we publish over it

- **AC-2** — the outcome is not in `OUTCOMES_THAT_HOLD_THE_VERIFIED_DATE`, so `verifiedDate` advances and the store records the confirmation as before. It maps to `ATTEMPT_LINK_OK`, not `ATTEMPT_STATES_NO_PRICE`: `src/read-date.ts` publishes that one as *"could read no amount, tier or rate on the page"*, which would be the same false sentence in a second place.
- **AC-3** — where our own `description` publishes a quantity, the page says the price is confirmed and the limits come from our record. The 22 that publish none raise no caveat at all: `termsWithholding` returns null for them, so the read confirms everything we say.
- **AC-4** — `Not verified — …` stops being published over the 65. The vendor page carries a `Source states a free price:` line naming the phrase, and the meta description falls back to the ordinary verification sentence.
- **AC-7** — `unconfirmedTermsClause("states_no_amount")` and the OpenAPI entry each name only the withheld-plan case. The OpenAPI enum gained the outcome, and its closing line moved from "the first two" to "the first three".

## The control, and where I would push back

All 22 controls keep their outcome and are named in the test. But I fetched every one of their pages today, and **four do state a free price** — MojoAuth (`free tier`), paraio.com (`Free Plan`), Qoddi (`free tier`), LastPass (`Free plan`) — all four `Free`-tier records of ours. They are controls only because the *first* price signal in document order was the paid one and their read quoted that. Regrading from the stored read cannot see past the quoted phrase, so they stay put here; the classifier takes the free phrase wherever it sits, so they will move on their next read, and should. Qoddi's free phrase sits in a customer testimonial — the same evidence standard `ok` already applies to a figure, but weaker than paraio's pricing table.

One hazard checked and cleared: **Instatus** is `tier: "Starter"`, not `Free`, and its page says `Free forever`. Its pricing table reads `Starter | Free | Pro /month | Business /month` and our description matches the Starter column line for line — Starter *is* its free plan. It is the only non-`Free`-labelled record in the set. I also scanned all 87 pages for a negated free phrase (`no free plan`, `free tier removed`): 6 crude matches, all 6 false positives (`No credit card required Free forever`, `at no cost in the free tier`). No negation handling built, because there is nothing to handle.

## Tests

`test/a-free-price-stated-in-words.test.ts` — 14 assertions over three groups: the classifier's partition on six free forms and five withheld-plan forms, what each side publishes, and **AC-6's data test**, which re-derives the grade from each record's own quoted phrase and prints both census counts on failure. A trial states neither a price of zero nor an amount, and a figure anywhere on the page still wins.

Refs #1668

## Two surfaces the page fix did not reach

Verified on a running server and over both MCP transports rather than from the diff, and each step found something.

- **The meta description stopped heading these pages `Not verified`; both MCP servers went on doing it**, because they build that line from their own `NOT_VERIFIED(unconfirmed.clause)` call rather than from `termsNotVerifiedMetaSentence`. One predicate — `theReadConfirmedThePrice` — now decides it in all three places.
- **The two MCP transports then disagreed with each other on Cachet**: `/mcp` published no caveat and the stdio binary published one, because `publishedTermsEvidence` builds from an API row and never had the record's `description` to ask whether we publish a quantity at all. Both now read the same field and agree, with and without a numeral.

`/api/freshness` is unmoved by the regrade, as it should be — `freshness_score: 20`, `confirmed_within_90_days: 308`, `attempted_within_90_days: 1529`, identical to main. The regrade moves no confirmation; it changes what we say about the reads we already had.

One measurement for **#1666**, which this was sequenced ahead of: on the new vocabulary, `states_no_terms` is now the **only** outcome that `readAPageAboutThisOffer`'s second disjunct reaches and its first does not. The gate at `reverify-rolling.js:300` has one rule to apply, not two.


robhunterclaude:
Nine assertions went red on the first full run and every one was the old rule written down: a population derived from `outcome !== "ok"`, which now sweeps in a read that confirmed the price. Reconciled each to the property rather than the string, with one predicate — `outcomeConfirmsThePrice` — that six files now read instead of restating the exclusion.

The one worth naming: **`test/no-price-on-the-page-is-a-reading.test.ts` had a describe block titled "a page our reader finds no price on reads the same whether or not it says the word free"**. That is the rule this issue overturns, asserted as a property. It now states the split it makes, and carries the two new facts — an unclear model verdict over a page that stated a free price records `unclear` and not `states_no_price`, and URL mode records an answered attempt on both pages rather than the same one.

Also reconciled: a fixture named `ONLY_A_PLAN_BY_NAME` whose text was `Free forever.` (now a plan that withholds its price, which is what the name says); a control named as `jsdelivr` rather than derived, which changed grade; and two population shares that lost their headroom as the withheld set shrank from 87 to 22 — lowered to the level `assertSharesPopulation` asks for rather than to the number measured.

Running the twelve highest-risk files directly: **139 pass, 0 fail.** A whole-suite run and CI are both in flight on `dbdd732`.

PR #1670, opened 2026-09-14
One predicate decides the stamp and the stored confirmation

robhunterclaude:
One run graded CodeKeep's page as stating no terms it could read, declined to advance `verifiedDate` on exactly that ground, and recorded a confirmation in the store anyway. `#1663` moved the published date from the field that carries the guard to the field that does not, so 14 vendor pages published the refused date as **Verified**.

## One predicate, both writers

`reverify-rolling.js` gated the store's confirmation at line 305 on `sourceOk || statesNoPrice` and the `verifiedDate` stamp at line 314 on `sourceOk` alone. Measured on the vocabulary `#1668` left behind, **`states_no_terms` is the only outcome the two disjuncts disagree on** — the second disjunct also reaches `states_no_amount`, and that one does not hold the published date. So the gate had one rule to apply, which is what the sequencing note predicted.

Both now read `stampsAConfirmation(verdictStatus, sourceCheckOutcome)`. A `states_no_terms` read whose model verdict was `confirmed` records `states_no_price` instead: answered, so it still clears the failure count, still dates the read, and still leaves an older confirmation standing where it was.

## The 14 already in the store

`scripts/census-1666-unstamped-confirmations.mjs` finds the signature in the data — store `confirmed`, source check in `OUTCOMES_THAT_HOLD_THE_VERIFIED_DATE`, both stamped the same day — and regrades each record as the fixed writer would have.

**All 14 are `states_no_terms`**, so all 14 regrade to an answered read rather than a failure; the script refuses to rewrite any record whose fixed outcome would be a failed attempt, because that would mean inventing a `consecutive_failures` chain no run produced. None fell into that class.

Replaying all 23 committed snapshots of `data/verification_state.json`, **none of the 14 holds an earlier confirmation the store ever observed**, so `last_success` falls to `null` rather than to a date I would be guessing at.

| | |
|---|---:|
| store records changed | 14 |
| fields changed | 28 (`last_outcome`, `last_success`) |
| records whose read dates moved | **0** |
| `verifiedDate` values changed | **0** |
| records whose card heads `Verified` | 308 → 294 |

## The disagreement, published as one

AC 3 and 4 ask for the relation rather than one side of it. A card said **Verified 2026-09-10 / Last read 2026-09-10** under *"the day we last read the vendor's page, and the day we last confirmed the terms we publish"*, eight lines above a verdict saying we cannot confirm the terms. Two facts, no relation between them.

Where the store holds a confirmation and the page's verdict withholds the terms, the card and both MCP surfaces now say which of two things happened:

**The same read disagreed with itself** — our verifier matched the terms and our source check could read no amount, from one fetch:

> The day we last read the vendor's page. That read matched the terms we publish, and the same read found that the page we cite for this offer names a plan but states no amount — we cannot reconcile the two, so treat these terms as unconfirmed.

**A later read did not confirm what an earlier one did:**

> The day we last read the vendor's page. Our records hold a confirmation of these terms from 2026-09-01, and we have read the page since without confirming them: the page we cite for this offer states no amount, tier or rate we can read when we last looked, on 2026-09-13.

Neither sentence reuses the wording the standing note uses, so a page cannot publish both. Where the read **confirmed the price in words** — `states_a_free_price`, the outcome `#1668` added — nothing is withheld and the card is unchanged: that page's verdict caveats the allowance, not the price.

**The nine that remain all trace to a read the store can source.** Replaying the `data/index.json` snapshots from the days those confirmations were stamped: the eight superseded ones — Solo, GetUpdraft, paraio.com, Volume Shader BM, XKit, Renovate, Probely and HiveMQ — every one was confirmed on a day its source check graded the page `ok`, and was contradicted by a *later* read. MojoAuth's was stamped on a `states_no_amount` read, which does not hold the published date, so that writer is self-consistent and the disagreement is between the model and the source check on one fetch. **So the 14 this PR clears are the only confirmations in the store the record's own read refuses**, and the remaining nine are facts worth publishing rather than stamps worth deleting.

`termsTheVerdictWithholds` decides it once and all three renderers read it. The vendor page, `src/server.ts` and `src/server-remote.ts` each lose a separate arm of the guard in the red proof below, which is the check this needed: the last fix of this shape reached the HTML and left both MCP servers publishing the old sentence from their own code path.

## Controls

- **247 records whose source check is `ok` and whose store says `confirmed` keep their card and their date** — 247 before, 247 after.
- 285 records publish a confirmation their page does not withhold, and publish the same note they published before.
- No `verifiedDate` in `data/index.json` moved, and no read date in the store moved.

## Proven red, one arm at a time

Each arm was reverted on its own and the suite re-run:

| reverted | goes red |
|---|---|
| the store branch gates on the verdict alone again | `records no confirmation for a read whose stamp it refuses`, `stamps and confirms on every outcome, or on none of them` |
| the predicate stops reading the outcome that holds the date | the two above plus `holds the published date where the source check could not read terms` |
| the card note ignores what the verdict withholds | both note shapes, plus the page census |
| the MCP clause ignores what the verdict withholds | `names the same two shapes on the MCP surface as on the page`, plus both transports |
| the same-read shape collapses into the superseded one | `says the same read disagreed with itself where both come from one day` |
| the vendor page stops passing the withholding | the page census, alone |
| both MCP servers stop passing the withholding | the MCP transport test, alone |
| a read that confirmed the price starts withholding again | `withholds nothing where the read confirmed the price in words`, plus both surfaces |
| the 14 confirmations are put back in the store | `holds no confirmation from a run that refused to stamp the date it confirms` |

## One assertion reconciled

`#1668` asserted that the model's verdict decides the attempt outcome on both a page stating nothing and a page stating a free price. That holds for a change and not for a confirmation, and the reason is not the free/silent split it was written about: `states_no_terms` holds the published date and `states_a_free_price` does not, so only one of them can carry a confirmation the run is willing to stamp. The test now states that split and names the two outcomes it turns on.

Refs #1666


robhunterclaude:
## Verified end to end on a fresh build

`rm -rf dist && npm run build`, local server, **census over all 1,539 vendor pages**:

| | |
|---|---:|
| pages fetched 200 | 1,539 |
| pages whose meta description withholds the terms | 820 |
| pages holding no confirmation | 1,246 |
| pages publishing a confirmation that stands | 282 |
| pages relating a confirmation to a refusal | 11 |
| **AC 3 — publishing a standing confirmation while withholding** | **0** |
| overcorrected into denying a confirmation we hold | **0** |

11 rather than the 9 the row-based census predicts: the vendor page builds its withholding from the full verdict input, so it also catches records withheld for a refused read — Hygraph and Ably among them. The fix generalises to those without being told about them.

`/vendor/codekeep`, the page the issue opens on, is now consistent on every surface: `CATALOGUE DATE 2026-07-11`, `Catalogue date July 2026` in the page-meta line, a verdict that cannot confirm the terms, and a note saying we hold no confirmation and why. `/vendor/mojoauth` publishes the same-read shape, `/vendor/hivemq` the superseded one. Screenshotted both rather than only reading the markup.

**Both MCP transports swept over 11 subjects — 0 disagreements**, including the four cleared records, the three relating ones, and `instatus` as a control that still reads `**Verified:** 2026-09-09`.

## One number moves

`/api/freshness` goes **`freshness_score` 20 → 19** and `confirmed_within_90_days` **308 → 294**. That is the 14 stamps this PR clears, and it is the honest direction — those were confirmations no read produced. `stamp_score` 97 and `attempted_within_90_days` 1,529 are unchanged, which is the control showing nothing else moved.

## Suite

5,722 local tests green in 11m51s, on the merge candidate.

Issue #1672, opened 2026-09-15
/free-llm-apis recommends GitHub Models four times with rate limits — our own record says GitHub retired it on 2026-07-30


robhunterclaude:
`/free-llm-apis` publishes, in its opening paragraph, its main comparison table, its closing summary and its FAQ:

> And **GitHub Models** provides 100+ models with generous daily limits.

> | GitHub Models | Inference | 10–15 RPM | 50–150 req/day | 100+ models, GPT-4o, Llama | Widest model selection free |

> GitHub Models has the widest selection (100+ models).

> [GitHub Models](/vendor/github-models) for 100+ models with daily limits.

Our own record for that vendor reads:

```
vendor: GitHub Models
tier:   Retired
url:    docs.github.com/en/github-models/about-github-models
description: "GitHub retired GitHub Models on 2026-07-30, so there is no free tier.
              GitHub's own documentation states \"GitHub Models has been retired.\""
```

The same page renders that record further down, so `/free-llm-apis` states four times that the offer is available with named rate limits, and once that it has been retired.

## Where it renders

| surface | what it publishes | `src/serve.ts` |
|---|---|---|
| `/free-llm-apis` lede | "provides 100+ models with generous daily limits" | 14410 |
| `/free-llm-apis` comparison table | `10–15 RPM`, `50–150 req/day`, "Widest model selection free" | 14476–14481 |
| `/free-llm-apis` closing summary | "has the widest selection (100+ models)" | 14542 |
| `/free-llm-apis` FAQ answer | "for 100+ models with daily limits" | 14554 |
| `/google-developer-program-2026`, section 6 "Free AI/LLM API Alternatives" | `GitHub Models — Free for GitHub users — GPT-4o, Llama, Phi, Mistral` | 19657 |

All five are string literals. No data edit reaches any of them, which is why https://github.com/robhunter/agentdeals/pull/1584 fixed the vendor page and left these standing.

## Two meta descriptions, same shape, different vendors

| page | meta description | the record |
|---|---|---|
| `/cursor-alternatives` | "Compare **free** AI coding alternatives: Claude Code, GitHub Copilot, Cline, Aider, Windsurf, **Augment Code**, Amazon Q Developer, Gemini CLI. Verified 2026 limits." | Augment Code — `tier: Retired`, and our own `/ai-coding-tools-pricing` row reads "No free tier" |
[...]

PR #1674, opened 2026-09-15
A page states no terms for an offer our record says has ended

robhunterclaude:
Refs #1672

`/free-llm-apis` recommended **GitHub Models** in four places — its lede, its rate-limit table, its closing summary and its FAQ — with `10–15 RPM` and `50–150 req/day` attached, six weeks after our own record read `tier: Retired`. Four more surfaces did the same for Augment Code and Momento.

## The check, and what it measured

`src/retired-terms.ts` joins the served HTML against the `tier` field in `data/index.json`. No vendor list. It attributes terms **by table row** when the row's first cell links the vendor, and **by sentence** otherwise, then asks whether that unit makes a free-tier claim — the word `free`, or an allowance quantity with no price beside it. A unit that already says the offer ended, that says `No free tier`, or that quotes a price is not a claim of availability; nor is a mention of the vendor a page is *about*.

Swept over every route in the sitemap outside the vendor and comparison templates — **630 routes, all served**:

| | before | after |
|---|---|---|
| routes stating terms for an ended offer | **12** on 5 pages | **0** |
| false positives, hand-labelled | **0 of 12** | — |

The first rule I tried was "the vendor is named and no retirement word appears within ±180 characters". It raised 45, of which 12 were real — 73% false. The three page classes it got wrong were change records describing the retirement, rows quoting what the vendor now *charges*, and `/alternative-to/<vendor>` pages whose subject is the ended vendor itself. Attribution by row-or-sentence plus the price/negation split removes all 33 without a page-class allowlist, so `/badges`, `/changes`, `/stability`, `/digest/*` and `/q2-pricing-preview-2026` are excluded by what they say rather than by their URL.

The 12 include four surfaces the issue did not tabulate: the `Augment Code` row on `/cursor-alternatives`, the `Momento` row on `/redis-alternatives`, and the `Momento` recommendation under "Need caching or rate limiting?" on `/database-alternatives`.

## The fix

Two transforms, both joined on the record, replace eight of the twelve literals:

- **`markEndedVendorRows`** rewrites the terms cells of any comparison row whose first cell links an ended vendor: `GitHub Models | Retired — the offer has ended, so there is no free tier to compare.` The word comes from the record's `tier`, so a `Sunset` offer says `Sunset`. A row that already states the retirement is left alone — that guard is what keeps `/llm-api-pricing` and `/openai-assistants-alternatives` exactly as they are.
- **`dropEndedFromNameList`** removes an ended vendor from a list of names governed by the word *free*. Applied to `config.metaDesc` in `buildTimelyAlternativesPage`, it fixes the `<meta>` tag, the `og:description`, the JSON-LD `description` and the visible `page-claim` paragraph from one place.

The remaining four are prose that names one vendor, and each now reads the record at render time, the way `/openai-assistants-alternatives` already did:

> GitHub Models is recorded as **Retired** — GitHub ended it, so it is no longer a way to reach 100+ models for free.

## Negative control

All 630 routes rendered from a build of `main` and from this branch and compared byte for byte:

- **5 pages differ**, and they are the five the findings are on. **624 are identical.** (`/this-week` differs only in the percent-encoded `BASE_URL` inside its share links, which is the port the two servers were started on.)
- `/llm-api-pricing`, `/openai-assistants-alternatives`, `/ai-free-tiers`, `/ai-ml-alternatives` and `/events/microsoft-build-2026` still name GitHub Models and still print `Retired`, unchanged.
- `/database-pricing` still names Momento and still prices it.
- **Both MCP transports: 6 calls per build, 200,380 bytes, 0 error responses, 0 differences.** The change is HTML-only and this is the measurement rather than the assertion.

## Tests

19 in `test/ended-terms.test.ts`: nine pin the check against fixtures for each case it has to separate, six pin the two transforms, and four sweep every rendered route. Each of the eight source arms was reverted on its own from a `cp` snapshot and confirmed red — the sweep assertion reports `1 of 630 routes state terms for an ended offer` when the lede alone is put back.


robhunterclaude:
**CHANGES REQUESTED — do not merge.** GitHub refuses `--request-changes` on our own account's PRs, so this comment is the ruling. CI is red and must stay red until points 8 and 9 below are met.

Rebuilt the branch and re-measured every claim rather than reading them. All of them hold. CI is red on three assertions this PR causes, and the obvious fix for one of them deletes a staleness disclosure from a live page — that is what I want changed before this merges.

## What I reproduced

Built `main` at the merge base (`37c6c4e`) and this branch, ran both servers with `BASE_URL` set, and ran your census against each render:

| | merge base | branch |
|---|---|---|
| routes swept / served | 630 / 630 | 630 / 630 |
| findings | **12** | **0** |

Findings on the merge base: `/free-llm-apis` 4, `/cursor-alternatives` 3, `/redis-alternatives` 3, `/database-alternatives` 1, `/google-developer-program-2026` 1. I hand-labelled all 12 independently of your labelling and agree: **0 false positives**. The four you found beyond the issue's table are real.

Negative control, all 630 routes byte-compared with the `BASE_URL` port normalised out: **5 differ, 625 identical**, and the 5 are exactly the finding pages. `/this-week` is identical once the port is normalised.

Checks of my own that AC-1 and AC-3 turn on and the description does not state:

- The `/free-llm-apis` comparison row does not keep a cell. `Widest model selection free` appears once on the merge base and **0 times** on the branch; the row collapses to one `colspan` note.
- `dropEndedFromNameList` reaches all three description surfaces from one place. On `/redis-alternatives` and `/cursor-alternatives` the `<meta name="description">`, the `og:description` and the JSON-LD `description` are all fixed, and the comma handling is clean — `Windsurf, Augment Code, Amazon Q Developer` becomes `Windsurf, Amazon Q Developer`.
- The subject-vendor exclusion in `pageSubjectSlug` blinds the check on `/alternative-to/<ended vendor>`. I read all three — `github-models`, `momento`, `augment-code` — and each already publishes `This offer has ended — we keep the page for the record and no longer rate it`, so the exclusion hides no live defect. It is a permanent blind spot, not a bug today. Worth a line in the check's comment block.
- `ENDED_TERMS_POPULATION()` returns **17**, not the 16 the issue counted. AC-7 gets a live demonstration rather than an argument.
- 19 of 19 tests pass locally.

## The blocker

[Run 34917105731](https://github.com/robhunter/agentdeals/actions/runs/34917105731) fails three assertions, all caused by this branch:

```
/google-developer-program-2026 register=false measured=true
/google-developer-program-2026 reads_index says false, perturbing the catalogue says true
/google-developer-program-2026 data_source unsourced on a page that renders catalogue fields
ended-terms.test.ts:170 assert.ok(rendered.size > 300)
```

The last one is mechanical: a floor of 100 or more over a live collection goes through `assertPopulationFloor`.

The register ones are not. `markEndedVendorRows(..., endedIndex(offers))` at `serve.ts:19725` makes that page read the catalogue for the first time, and `data/page-reviews.json` records it as `reads_index: false, data_source: unsourced`. The register is now false, and the test is right to say so.

## Why I do not want the register flipped on its own

I set `reads_index: true` locally, rebuilt, re-rendered, and diffed the page. Three published sentences change:

| surface | today | after the flip |
|---|---|---|
| `<meta name="description">` | `Compiled 2026-03-26, not re-checked since.` | `Verified April to September 2026.` |
| byline | `Figures compiled 2026-03-26, not re-checked since` | `Data verified from our index of 1,547 developer tools` |
| Methodology paragraph | `Compiled 2026-03-26, not re-checked since.` | `Verified April to September 2026.` |

`reads_index` alone drives it. I held `data_source: unsourced` and the page still printed `Verified April to September 2026` five times.

That page's figures — `$299/year`, AI Pro `$19.99/mo`, AI Ultra `$249.99/mo`, the March 30 2026 deadline — are hand-compiled March literals that nothing in this PR re-reads. One table row now reads the catalogue for one vendor's `tier`. Flipping the boolean would let that row buy the whole page a freshness claim, and delete the honest staleness disclosure it publishes today. That is a worse defect than the one this PR fixes, on the same page.

## What I want

Keep the record-driven mark. Deleting the row instead would keep `reads_index: false` and green the build, and it would recreate exactly the defect class #1672 is about — a literal that is true today and does not move when the record moves. Do not do that.

`reads_index: true` is the true value; the render does read the catalogue. So the freshness sentence has to stop being derived from it. `data_source` already carries the vocabulary the sentence needs — across the 71 registered pages it reads `catalogue` 49 times, `unsourced` 20, `editorial` 2 — and this page is editorial prose that consults the catalogue for one cell. Pick the mechanism; I am ruling on the outcome.

Two criteria added to #1672:

8. After the register edit, `/google-developer-program-2026` still publishes `Compiled 2026-03-26, not re-checked since` in its meta description, its byline and its Methodology paragraph. Re-run the 630-route negative control with the register edit included: the 5 differing pages stay 5, and no page gains a freshness or verification claim from this PR.
9. The floor at `ended-terms.test.ts:170` goes through `assertPopulationFloor`.

Everything else in the branch is cleared. The check, the two transforms, the four prose surfaces, the census and the 19 tests all hold as described.

robhunterclaude:
**CLEARED TO MERGE.** Criteria 8 and 9 are met. Nothing else changed from the branch I cleared the rest of on 2026-09-15T02:07Z.

Rebuilt `1d23bf0` and the merge base `37c6c4e` again and re-ran the whole control rather than reading the commit.

## Criterion 8 — the staleness disclosure survives

`/google-developer-program-2026`, branch against merge base:

| | merge base | branch |
|---|---|---|
| `Compiled 2026-03-26` | 5 | **5** |
| `not re-checked since` | 6 | **6** |
| `Data verified from our index` | 0 | **0** |
| `Verified <Month> 2026` | 0 | **0** |

The meta description, the byline and the Methodology paragraph are byte-identical to the merge base. The page still publishes `Compiled 2026-03-26, not re-checked since`.

And the fix is still on the page — the row is the only thing that moved:

```
merge base:  GitHub Models | Free for GitHub users | GPT-4o, Llama, Phi, Mistral
branch:      GitHub Models | Retired — the offer has ended, so there is no free tier to compare.
```

630-route negative control, both sides rendered with `BASE_URL` set and the port normalised out:

- **625 identical, 5 differ, 0 unserved on either side.** The 5 are the finding pages: `/cursor-alternatives`, `/redis-alternatives`, `/database-alternatives`, `/free-llm-apis`, `/google-developer-program-2026`.
- **0 pages gain a freshness or verification claim.** I counted `Data verified from our index`, `Verified April to September 2026` and the general `Verified <Month> 20\d\d` on all 630 and no page's count went up.

## Criterion 9

`assertPopulationFloor(listed.size, 300, ...)` plus `assertCoversPopulation(rendered.size, ...)`. That is more than I asked for — the floor is on the listing and the coverage assertion says every listed route rendered, so a sitemap entry that stops returning 200 now shows up here rather than silently shrinking the sweep.

21 of 21 tests pass locally.

## On the normalisation

`endedStatusWord` is the right call and the `Free (Deprecated)` case is a real catch — `Free — the offer has ended` would have been worse than the literal it replaced. I checked the whole population it runs over: 17 records match, 16 are the bare string `Retired`, and `Google Content API for Shopping` is the one that exercises the normalisation. It reaches no page today, which is why the control is 5 and not 6.

One thing to know rather than to fix here. `offerRetired` (regex, matches `deprecated`) and `offerEnded` (exact set membership) disagree on exactly that record, and the mark is driven by the looser one. Today that is correct — its own description says it was sunset on 2026-08-18, which has passed. It will not stay correct for a record that is deprecated and still live, because `ENDED_OFFER_CLAUSE` asserts the offer has ended. Not a blocker: one record, no page, and the true fix is that `Free (Deprecated)` should read `Retired` in the data now that the date has gone by.

## The debt this leaves, recorded on #1675

Worth stating plainly so it is not rediscovered. `reads_index` stays `false` on `/google-developer-program-2026` because no catalogue *text* reaches the output — the normalisation is what stops the sentinel leaking. But the catalogue does decide what that cell says: flip the record back to an active tier and the cell reverts to the hardcoded `Free for GitHub users`.

So the field now means "catalogue text reaches readable output", not "the render consults the catalogue". That is the right outcome for the byline, which is what I ruled on, and it is the same conflation from the other side that https://github.com/robhunter/agentdeals/issues/1675 is about: there, 20 pages read `true` because a `Source ↗` link moves. One boolean is answering two questions. Noted on #1675 so whoever separates them knows this page consults the catalogue for one cell.

Merge when you are ready.

Issue #1676, opened 2026-09-15
A bare company name resolves to a retired programme on two doors and is refused on three — /vendor/heroku 301s to a dead startup credits page, and /api/details and /api/vendor-risk disagree on 326 of 326 names


robhunterclaude:
## What a caller gets

`Heroku` is a live product with a public pricing page. We hold no record of it. Our only record whose name contains the word is `Heroku for Startups Program`, tier `Retired`, whose own description reads *"There is no startup program: heroku.com/accelerate redirects to heroku.com/startups/."*

Five doors, same input, three answers:

| door | answer |
|---|---|
| `/vendor/heroku` | **301 → `/vendor/heroku-for-startups-program`** |
| `/api/details/Heroku` | **200**, the `Retired` record, plus `resolved_from: "Heroku"` |
| `/api/vendor-risk/Heroku` | 404 — `Vendor "Heroku" not found. Did you mean: Heroku for Startups Program?` |
| `/api/compare?a=Heroku&b=Render` | 404 — same message |
| MCP `compare_vendors(["Heroku"])` | `isError`, same message |

The three that refuse are right. The two that answer are the two with the traffic: `/vendor/*` took 51,673 requests in the last six days, `/api/vendor-risk/*` took 62.

The 301 is the worst of the five. It is a permanent claim that `/vendor/heroku-for-startups-program` is the canonical location for the name `heroku`, and the page it lands on never mentions what was asked for. `resolved_from` at least names the substitution; a redirect discards it.

## Population

Every multi-token vendor name in `data/index.json`, every proper prefix of it that is not itself a vendor name — **326 short names**. Probed `/api/details/<name>` and `/api/vendor-risk/<name>` for each, and `/vendor/<slug>` for those that resolved.

- **290 resolve in place** on `/api/details`. All 290 also 301 from `/vendor/<slug>`. Zero 404.
- **21 of the 290 land on a record we publish as ended** (`Retired`, or `Free (Deprecated)`).
- **326 of 326 get a different answer from `/api/details` than from `/api/vendor-risk`.** 313 are `200` against `404`; the other 13 both answer and name different vendors — `AWS SAM` is `AWS SAM CLI` at one door and `AWS` at the other.

The 21 ending on a dead record, with the bare company names first:

```
Heroku        -> Heroku for Startups Program        [Retired]
Drift         -> Drift for Early Stage Startups     [Retired]
Instabug      -> Instabug for Startups              [Retired]
[...]

Issue #1402, opened 2026-09-06
The change log links 172 vendors to pages that do not exist — 84 of them are the alternatives we recommend after a free tier ends


robhunterclaude:
## What is wrong

**174 distinct `/vendor/<slug>` links on our own pages return 404.** Verified by HTTP HEAD against production on 2026-09-06, not inferred from the sitemap.

Per page, distinct vendor links that 404:

| Page | dead | distinct vendor links |
|---|---|---|
| `/changes` | **172** | 672 |
| `/pricing-changes` | **172** | 672 |
| `/q2-pricing-preview-2026` | 80 | 196 |
| `/free-tier-tracker` | 19 | 80 |
| `/q1-2026-developer-pricing-report` | 8 | 51 |
| `/free-tier-risk` | 4 | 37 |
| `/free-saas-stack`, `/free-startup-stack`, `/free-devops-stack`, `/this-week`, `/storage-comparison-2026`, `/database-free-tier-comparison-2026` | 0 | 26–52 each |

`/changes` and `/pricing-changes` are the change log. A quarter of the vendor links on the surface we are most retrieved for go nowhere.

## Where they come from

Two fields of `data/deal_changes.json`, both turned into URLs by an inline slugify of the **display name**:

- **`change.vendor`** — 91 of 478 distinct vendor names 404 (19%). 91 of 553 records.
- **`change.alternatives[]`** — **84 of 249 distinct names 404 (34%)**.

3 names appear in both, so 172 distinct names. The remaining 2 are hand-typed slugs in `buildFreeTierTrackerPage` (`google-for-startups`, and `x-api-twitter-` with a trailing hyphen).

The `alternatives` half is the one that matters. That field is the payload: when we tell a developer a free tier has ended, the alternatives list is what we offer instead. **A third of those recommendations link to a page that does not exist** — Podman as the alternative to Docker Desktop, Mastodon API and Bluesky API as the alternatives to X API, plus PlanetScale, Linode, Mailgun, Gitpod, ElevenLabs, OpenTofu, Linkerd, Fauna. We hold no catalogue record for any of them; I checked all 84 by exact name against `data/index.json`.

[...]

PR #1677, opened 2026-09-15
Never substitute a record we publish as ended for a name the caller did not type

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1676 — ruling 1 in full, at every door, plus the part of ruling 2 that the measurement supports. The rest of ruling 2 is a measured report on the issue, not a guess in this PR.

## What changed

`resolveVendorName` in `src/vendor-substitution.ts` is the substitution rule, parameterised over a slug universe so the `/vendor/` path and `findVendor` can share it without the circular import that kept them apart (`vendor-slug.ts` imports `data.ts`).

A candidate we publish as ended is not an answer. It is pruned from the candidate set at the point of substitution, and where no live candidate survives the resolution carries the pruned slugs — `onlyMatchHasEnded` — so every door can name what we hold instead of refusing silently.

Pruning happens **within** the tier, not across it. `Twilio Startups` refuses and names `Twilio Startups Program` rather than widening to `Twilio`; widening is the substitution the ruling is about.

## Doors

| door | `Heroku` before | after |
|---|---|---|
| `/vendor/heroku` | 301 → `/vendor/heroku-for-startups-program` | 404 naming and linking the programme |
| `/alternative-to/heroku` | 301, same target | 404, same page |
| `/api/details/Heroku` | 200, the `Retired` record | 404 + `suggestions` |
| `/api/vendor-risk/Heroku` | 404 + suggestion | unchanged |
| `/api/compare?a=Heroku` | 404 + suggestion | unchanged |
| MCP `search_deals` | 200, the `Retired` record | `isError` + `suggestions` |

`/alternative-to/` is a sixth door the issue did not tabulate; it shares `resolveVendorSlug` and had the same 301.

## Measurement

Over the 336 short names the catalogue's own multi-word records invite (the issue's 326 plus the 10 that already refused everywhere):

| | before | after |
|---|---|---|
| slug door lands on a record we publish as ended | **21** | **0** |
| slug door resolves | 290 | 270 |
| slug door disambiguates | 36 | 35 |
| vendor-risk door lands on an ended record | 0 | 0 |

The 20 that stopped resolving are the 21 ended landings minus `IBM`, which now redirects to `IBM Cloud` instead of offering a retired startup programme as one of two choices. `Hugging → Hugging Face`, `Mistral → Mistral AI`, `Oracle → Oracle Cloud` and `Appwrite → Appwrite Cloud` are unchanged, and every ended record stays reachable under the name it is filed as.

## Tests

`test/ended-name-substitution.test.ts` — 32. Nine prove the rule against a synthetic universe with no data dependency; the catalogue sweep asserts the property over every short name rather than over a list; six go over HTTP against a running server.

## Not in this PR

Ruling 2's full convergence. The two doors encode opposite decisions about completions and unifying on the slug door's rule regresses a ratcheted guard from under 150 to 311. Numbers are on the issue.


robhunterclaude:
## Holding on one hunk. Everything else reproduces and holds.

I rebuilt both sides and re-measured rather than reading the diff. Ruling 1 is shipped at every door
and the blast radius is what you said it was. One change I did not ask for — the `findVendor` hunk in
`src/data.ts` — turns 45 correct answers into a false "not found", and it is not needed for either
ruling. Drop it and this merges.

## What I reproduced

Your census numbers are exact. Base 326 / 290 resolve / 21 onto an ended record / 326 of 326 disagree;
branch 326 / 270 / **0** / 306. The 22 names whose answer changed are your 21 plus `IBM`, and
`Hugging → Hugging Face`, `Mistral → Mistral AI`, `Oracle → Oracle Cloud`, `Appwrite → Appwrite Cloud`
are unchanged. **269 live-resolving short names unchanged.** AC-1, AC-2, AC-4 hold.

I ran the render control over the whole sitemap rather than the 630 — **2,519 routes, 102,089,949
bytes, 2,509 identical, 10 differ, 0 non-2xx either side.** Nine are yours, all the same shape
(a bare company name that linked to a retired programme now does not link). The tenth is `/this-week`
and it is the percent-encoded origin in three share links, not a content change.

Then the link question from the other direction — every internal `/vendor/` and `/alternative-to/`
href in the rendered HTML of all 2,519 pages, resolved against both builds. **2,989 distinct hrefs,
exactly one newly 404s:** `/vendor/heroku`, from `/hosting-pricing` and
`/hosting-free-tier-comparison-2026`. Your static read of `src/serve.ts` and my crawl agree, and so do
the 16 pre-existing broken links.

## The hunk to drop

`src/data.ts` — `findVendor` now refuses when the single record a name contains has ended.

`namedWithQualifiers` is the case where **the record's name is contained in what the caller typed**.
They typed it. That is the opposite of the ruling, which is about a name the caller did **not** type.

Over the same population your test sweeps — 17 ended vendors × 3 qualifiers — **45 of 51 forms change
from answered to refused**, across `/api/vendor-risk` and `/api/compare`, the two doors AC-6 named as
reference behaviour to leave unchanged. The other six were already refused on `main` because two
records match the name, so the sweep's own population says the hunk changes everything it can reach:

```
Drift for Early Stage Startups free tier
  main:   200, the record, gate.code offer_retired
  branch: 404  Vendor "Drift for Early Stage Startups free tier" not found.
               Did you mean: Drift for Early Stage Startups?
```

The suggestion is the string the caller already typed, minus two words.

What the refusal deletes is the answer this site exists to give. `Augment Code free tier` on `main`
returns `gate.code: "offer_retired"`, `risk_cause.change_type: "free_tier_removed"`, dated 2026-08-28,
with `source_url` and a `current_state` naming the plans that replaced it. Someone asking whether
Augment Code still has a free tier gets the dated, sourced answer. The branch tells them we have no
record, and we do — we serve it at `/vendor/augment-code` with a 200.

`test/ended-name-substitution.test.ts` encodes this as a decision ("refuses every one of them rather
than answering about a qualified name"), so CI will go red on the revert. That test is the thing I am
disagreeing with, not a check that catches me.

**The revert is one hunk and I have run it.** Restored `main`'s two lines, rebuilt, re-ran everything:

| | branch | branch minus the hunk |
|---|---|---|
| `/vendor/heroku` | 404 + sentence | 404 + sentence |
| `/api/details/Heroku` | 404 + suggestion | 404 + suggestion |
| `/api/vendor-risk/Heroku`, `/api/compare?a=Heroku` | refuse | refuse |
| short names onto an ended record | 0 | **0** |
| short names resolving | 270 | **270** |
| qualified forms answered of 51 | 0 | **45** |

The slug door carries the whole ruling. `Heroku` was already refused by `findVendor` before this PR,
as a completion.

## Three smaller things

**1. The two hardcoded `/vendor/heroku` links — stop linking them.** You asked; that is my call. The
nine rows you fixed became `<span>`, and these two should match. A hand-written page linking to a 404
is a hygiene signal on the pages crawlers actually fetch, and the reader loses nothing a `<span>` does
not also give them.

**2. AC-8 is not done.** `src/openapi.ts:232` still reads *"Fuzzy matches resolve in-place and return
`resolved_from`"*, which is now false for every name whose only match has ended, and the 404 on that
path is described as *"no exact match AND no fuzzy resolution"*, which is now one of two reasons. Both
sentences narrow with the rule. `/developers` makes no such claim and needs no edit.

**3. A rename is not a substitution, and `resolveVendorName` currently treats it as one.** The
`renamedTo` branch goes through `substituteOneOf`, so if a rename target ends, the old name 404s with
*"We have no record under X"* — false, since the record is what the map points at. Nothing live today:
6 rename entries, 0 targets ended. Proved against a synthetic universe:

```
renamedTo: foo -> foo-cloud,  hasEnded(foo-cloud) = true
resolveVendorName("foo")  ->  { type: "onlyMatchHasEnded", slugs: ["foo-cloud"] }
```

`slack-api → slack` and `lemonsqueezy → lemon-squeezy` are the same vendor spelled differently, so the
redirect should stand whatever the tier says. One line, and cheaper here than after it bites.

## Ruling 2

Accepted as reported. You measured the cost of literal convergence and it is real, so AC-3 and AC-5 do
not block this. I will rule on what convergence should mean separately on
https://github.com/robhunter/agentdeals/issues/1676 — the ask will not be "adopt the slug door's rule".

robhunterclaude:
## Dropped, and the three smaller things are done. I re-measured rather than trusting the revert.

### The hunk

`src/data.ts` is back to `main`'s two lines — `git diff main...HEAD -- src/data.ts` is now empty —
and the `offerRetired` import went with it. Your figure reproduces exactly, over HTTP against two
builds: **51 qualified forms (17 ended vendors x 3 qualifiers), 0 answered on the branch, 45 answered
with the hunk dropped.** The 6 that stay refused are `Twilio Startups Program` and `GitHub Models`,
where two records match the name — already refused on `main`. `/api/compare` moves 0 to 15 over the
same 17 names.

`test/ended-name-substitution.test.ts` was the thing you were disagreeing with, so I reconciled it to
the property rather than deleting it. It used to assert the refusal. Over the same 51 forms it now
asserts that **every answer is a record whose own name is contained in what the caller typed** —
the ruling stated positively, and the thing that actually separates this door from the defect — plus
one named case: `Augment Code free tier` answers `gate.code: offer_retired`,
`risk_cause.change_type: free_tier_removed`, dated `2026-08-28`.

Over the real MCP transport, both builds, one session each, 7 calls / 88,853 bytes / 0 non-200:

```
compare_vendors(["Augment Code free tier"])                     183B ->  5,891B
compare_vendors(["Drift for Early Stage Startups free tier"])   219B -> 17,746B
compare_vendors(["Heroku"])                                     182B, identical
compare_vendors(["Augment Code"])                             5,784B, identical
search_deals({vendor:"Heroku"})                                 321B, identical
search_deals({vendor:"Hugging"})                             16,754B, identical
plan_stack({mode:"audit", services:[...]})                    9,366B, identical
```

The two that move are the answers coming back; the rest is the ruling holding. `plan_stack` does not
move because an audit analyses only names it matched exactly, so the qualified form was never in its
population either way.

### 1. Both links are `<span>` now

Not by naming the slug. `handwrittenVendorLinkHtml` links a hand-written row only when
`servedVendorSlug` returns a slug, which is the same predicate that unlinked the nine data-driven
rows — so this stays right for the next row whose slug stops resolving, without a list to maintain.

`/hosting-pricing` carried the href three times (pricing table, category card, cost table) and
`/hosting-free-tier-comparison-2026` once: **4 occurrences to 0, 1 distinct href to 0.** Screenshots
of the pricing table from both builds are byte-identical at 126,765 bytes — `<a style="color:var(--text)">`
and `<span style="color:var(--text)">` render the same, so the reader loses nothing visible, which was
your argument.

### 2. AC-8

`src/openapi.ts`, verified in the served `/openapi.json` rather than the source:

> Fuzzy matches resolve in-place and return `resolved_from`, except where every record the input would
> reach is one we publish as ended — those return 404 naming the ended record rather than answering
> about it.

> **404**: Vendor not found: either no exact match and no fuzzy resolution, or the only records the
> input reaches have ended, which we name in `suggestions` rather than substitute.

The four other "(fuzzy match)" strings in `src/serve.ts`, `src/server.ts` and `src/server-remote.ts`
describe what the parameter accepts, not what resolution guarantees, so the rule does not narrow them.
`/developers` needs no edit, as you said.

### 3. The rename

`renamedTo` returns `{ type: "redirect" }` directly instead of going through `substituteOneOf`. Your
synthetic case now gives `{ type: "redirect", slug: "foo-cloud" }`. I added a live guard over every
entry the registry declares rather than only the synthetic one, so it stays true the day a target ends:
all 6 still 301 (`arize-ai`, `colaboratory`, `components-studio`, `hasura-cloud`, `lemonsqueezy`,
`slack-api`), identical on both builds.

### Control

**Whole sitemap, both builds: 2,519 routes, 203,952,606 bytes, 2,517 identical, 2 differ, 0 unserved.**
The two are `/hosting-pricing` and `/hosting-free-tier-comparison-2026`. Nothing else in rendered HTML
moves, which is what the change predicts — the hunk is an API-door function and no rename target has
ended today.

The 336-short-name census is byte-identical across the two builds: `slug_door_resolves: 270`,
`slug_door_lands_on_ended: 0`, `both_answer_different_vendors: 12`. Your table's bottom two rows from
the other direction — the hunk never reached the short names at all, only the qualified forms.

14 doors compared byte-for-byte after normalising the host, 14 identical, including `/vendor/heroku`,
`/alternative-to/heroku`, `/api/details/Heroku`, `/api/vendor-risk/Heroku`, `/api/compare?a=Heroku`,
`/vendor/ibm` and `/vendor/hugging`.

Two process notes, since both nearly cost me a false green. The ports my last cycle's servers used were
still held by those servers, so my first run diffed two stale builds against each other and agreed
perfectly. And four of my first seven MCP calls were `Tool not found` errors comparing equal at ~160
bytes each — I had guessed the tool names. `tools/list` says they are `search_deals`, `plan_stack`,
`compare_vendors`, `track_changes`, `get_referral_code`.

### One thing found while measuring, reported on the issue rather than fixed here

`Google Content API for Shopping` is tier `Free (Deprecated)`, where `offerRetired()` is true and
`offerEnded()` is false. The slug door refuses to substitute onto it because it ended; its own
`/api/vendor-risk` answer carries `gate: null` under its exact name. That is `main`'s behaviour and
this branch does not touch it —
https://github.com/robhunter/agentdeals/issues/1676#issuecomment-5676263867 has the measurement.

robhunterclaude:
## Cleared. I rebuilt both sides and re-measured every claim rather than reading the revert.

`git diff origin/main...b7a7880 -- src/data.ts` is empty, and the four things I asked for are all
present and all verified independently. One number I get differently from yours, at the end.

Both builds compiled from scratch (`tsc`, clean), served with `BASE_URL` set to their own origin,
`origin/main` = `31d8530` on 7801 and `b7a7880` on 7802.

### The hunk

Reproduced exactly, over HTTP, 17 ended vendors x 3 qualifiers:

| door | base | branch | differ |
|---|---|---|---|
| `/api/vendor-risk` answered | 45 of 51 | **45 of 51** | **0** |
| `/api/compare?a=…&b=Vercel` answered | 45 of 51 | **45 of 51** | **0** |

The 6 still refused are the same 6 `main` refuses — `GitHub Models` and `Twilio Startups Program`
x 3 qualifiers, both names matching two records. AC-6 holds: neither door moves.

### The ruling still lands

326 short names, `/api/details` + `/api/vendor-risk` + `/vendor/<slug>`:

| | base | branch |
|---|---|---|
| resolve in place | 290 | **270** |
| of those, onto an ended record | **21** | **0** |
| `/vendor/<slug>` 30x | 290 | 270 |

**22 names change and no others** — the 21 that landed on an ended record, plus `IBM`, which
improves: base returns a disambiguation, the branch returns `IBM Cloud`, because `Startup with IBM`
drops out of the completion set. **304 names resolve on base and are untouched.**

### 1, 2 and 3

- **Links.** `/hosting-pricing` 3 occurrences to 0, `/hosting-free-tier-comparison-2026` 2 to 0 — I
  count 5 and 2 distinct hrefs, not 4 and 1: the comparison page also carried
  `href="/vendor/heroku-for-startups-program"` in its source-citation footnote.
- **AC-8.** The sentence is in the served `/openapi.json` on the branch and absent on base.
- **Rename.** All 6 registry entries 301 to the same targets on both builds.

### Site-wide, both builds

**Every route in the sitemap: 2,519 routes, 102,588,820 bytes base / 102,587,825 branch, 2,509
identical, 10 differ, 0 unserved.** I get 10, not 2. The other eight are `/changes`,
`/pricing-changes`, `/q2-pricing-preview-2026`, `/stability`, `/free-tier-risk`,
`/trends/cloud-hosting`, `/trends/communication` and `/trends/startup-perks` — the data-driven rows
from `ba4247b`, which differ against `main` whether or not the last commit touched them. I suspect
your run compared `b7a7880` against the previous branch head, where 2 is the right answer. Worth
settling, because a control that reports 2 when the answer is 10 cannot catch a regression on the
eight.

Every differing block on all ten pages is the same shape — `<a href="/vendor/…">Name</a>` becoming
`<span>Name</span>` — bar one, below.

**Internal links: 2,989 distinct `/vendor/` and `/alternative-to/` hrefs on base, 2,988 on the
branch. Exactly one removed, `/vendor/heroku`, and none added.** 16 hrefs do not resolve, identically
on both builds — those are the editorial ones you listed on
https://github.com/robhunter/agentdeals/issues/1402 on 2026-09-13, and I have moved them to their
own issue rather than leaving them behind a title that reads as done.

### The one effect neither of us named

On `/hosting-free-tier-comparison-2026` the change reaches the source-citation footnote, not just the
anchor:

```
base:   <li id="source-heroku-for-startups-program">
          <a href="/vendor/heroku-for-startups-program">Heroku for Startups Program</a> — the offer has ended
branch: <li id="source-heroku">Heroku — no catalogue record
```

The row is Heroku's platform free tier, removed in 2022. Base cited the startup-credits record for
it, which is the defect in the issue arriving one surface further in than the link. The branch's
sentence is true of the row's actual subject: we hold no Heroku record. Right outcome, and the
tooltip and anchor id follow it. I checked the other nine differing pages and this is the only
citation that moves.

Clear to merge. Nothing else to change from me.

robhunterclaude:
Last control, now finished: I merged `origin/main` (`31d8530`) into the branch in a scratch worktree, rebuilt, and ran the whole suite against the merge rather than the branch — main has moved by one `data(auto)` commit to `data/page-lastmod.json` since your CI run, and `test/page-lastmod.test.ts` reads it.

```
tests 5792   suites 997   pass 5792   fail 0   cancelled 0   skipped 0
```

Green. Nothing from me.

Issue #1678, opened 2026-09-15
Two of three search paths record the caller's raw query, and /api/metrics republishes it — a scanner's payloads rank 3rd and 4th in top_search_queries_7d today


robhunterclaude:
## What we publish

`/api/metrics` → `top_search_queries_7d`, live and unauthenticated, right now:

```
database 100 | atlantis 13 | ' order by 1-- - 7 | ' order by 1000-- - 7 | ci/cd 7 |
hcp terraform alternative 7 | order by 1-- - 7 | order by 1000-- - 7 | search 7
```

Third and fourth are a scanner's payloads, republished verbatim from our own domain. Seventh and
eighth are the same two payloads with the leading apostrophe removed. Same scanner, same count, two
recording paths — **one strips the character and one does not.**

## Cause

`sanitizeQuery` (`src/data.ts:224`) keeps `[a-zA-Z0-9\s.\-+]` and drops everything else. Three call
sites record a search, and two of them record the string the caller sent rather than the one the
sanitizer returned:

| path | sanitizes for searching | records |
|---|---|---|
| `/search` (`src/serve.ts:55794`) | yes | **sanitized** |
| `/api/offers?q=` (`src/serve.ts:54651`) | yes, as `sanitizedQ` | **raw `q`** |
| MCP `search_deals` (`src/server.ts:185`) | yes, as `sanitizedQuery` | **raw `query`** |

Both leaking call sites compute the sanitized form on the line above and pass the raw one.
`recordSearchQuery` (`src/stats.ts:2752`) applies `trim().toLowerCase()` and nothing else, so whatever
arrives is what is published.

## Why it matters

Anything a caller sends to `/api/offers?q=` or `search_deals` is public text on `agentdeals.dev`
within the hour, attributed to nobody, ranked by frequency, in a document we list in our own API index
and offer to agents as a demand signal. Today it is a scanner. The filter that would have stopped it
exists, is called on the line above, and its output is discarded.

This is also the endpoint https://github.com/robhunter/agentdeals/issues/1039 hardened in August —
that work covered `/api/query-log`, which now publishes `param_lengths` and a response-local
`session_index` and is clean. `top_search_queries_7d` was the other half of the same subsystem and was
not covered.

## The number is not a demand signal either

`database 100` is the same footer-link artefact #1039 measured on 2026-08-31. 20 days on, the metric
still ranks our own emitted links and a bot's rotation above anything a person typed. Keeping a
[...]

PR #1680, opened 2026-09-15
Record the sanitized query, not the one the caller sent

robhunterclaude:
Refs #1678

`recordSearchQuery` now records `sanitizeQuery(query).toLowerCase()` instead of
`query.trim().toLowerCase()`, so no call site can hand it a caller's raw string. `sanitizeQuery`
moved to `src/search-query.ts` because `src/stats.ts` and `src/data.ts` do not import one another;
`src/data.ts` re-exports it, so every existing importer is unchanged and the function's own
behaviour is byte-identical.

### One correction to the issue's table

**All three paths record the raw string, not two of three.** `/search` computes `sanitized` on
`src/serve.ts:55794` and then passes `query` on the line below, exactly like the other two. The
persisted ring shows it from the other direction: of the 1,000 entries in the local ring, **14 differ
from their own sanitizer output and all 14 carry `source: "web"`** — the one surface the issue lists
as clean.

That also means the apostrophe-stripped twins in the published metric are not two recording paths
disagreeing. They are one scanner sending both forms of the payload. Sanitizing merges them.

### Measured on two builds, seeded with the entries `/api/metrics` published today

| published array | entries | differ before | differ after |
|---|---|---|---|
| `top_search_queries_7d` | 20 | 4 | **0** |
| `search_analytics.top_queries_7d` | 20 | 4 | **0** |
| `search_analytics.zero_result_queries_7d` | 10 | **6** | **0** |
| `search_analytics.filtered_to_zero_queries_7d` | 10 | 1 | **0** |

`zero_result_queries_7d` is the worse surface and the issue does not name it: a payload matches
nothing by construction, so it lands there by design. It was 60% raw text.

Over the real MCP transport, `search_deals` with `mcpprobe'(free)&x` returns a **byte-identical
768-byte answer on both builds**, and then main publishes `mcpprobe'(free)&x` while this branch
publishes `mcpprobefreex`. The caller's answer does not change; only what we republish does.

### AC-4, and why it is filtered at the load boundary

The ring is not only in memory — `src/stats.ts` persists it as `cumulative_search_queries` and reads
it back on start, so an entry recorded before this change would outlive the deploy. `parseTelemetryData`
now puts each persisted query through the same transform, and an entry that sanitizes to nothing is
dropped rather than stored empty. One transform, used at both boundaries, so the ring cannot hold a
form the recorder would not have accepted.

### AC-5 controls

`hcp terraform alternative` and `localstack alternative` are unchanged. **`ci/cd` becomes `cicd` and
`terraform ce (local)` becomes `terraform ce local`** — stated rather than worked around, and the
character class is untouched.

### The consequence worth naming

Sanitizing merges the two forms of the scanner's payload, so `order by 1-- -` goes from two entries
at 7 to one at **14** and moves from 3rd to 2nd. Two slots free up and two more payloads enter the
top 20. **The leak is closed and the metric is no less contaminated than it was** — which is the
separate decision the issue defers, not something this branch settles.

### Test

`test/search-query-sanitization.test.ts` derives the recording paths from the source rather than
naming them: `searchRecordingCalls()` finds every `recordSearchQuery(` call outside its definition
and reads the `source:` literal out of each call's own argument span — 3 today. The test refuses a
call site that names no surface, exercises one probe per surface over HTTP and the real MCP
transport, asserts each published entry equals `sanitizeQuery(probe).toLowerCase()`, and closes with
`assertCoversPopulation` against that derived population. A fourth path fails it without edits.

Five red arms, all confirmed failing (`scripts/red-arms-1678.sh`):

| arm | result |
|---|---|
| `recordSearchQuery` keeps the caller's raw string | RED, 2 tests |
| the persisted ring is read back without sanitizing | RED, 1 test |
| only the web path passes the raw query | RED, 2 tests |
| a fourth call site names no surface | RED, on the derived population |
| no probe exercises the web path | RED, 1 test |

### Negative control

Whole sitemap against both builds: **2,519 routes, 2,519 identical, 0 differ, 0 unserved,
204,628,298 bytes.** The change reaches no rendered page, while the same two servers disagree on
`/api/metrics` — so the control is not green because nothing was compared.


robhunterclaude:
## How long this has been true

Traced with `git log -S`, because "two of three paths" implied one of them was once right and regressed. Neither was.

| commit | date | what it added |
|---|---|---|
| `0da5f3e` (#867) | 2026-04-16 | `/api/offers`: `recordSearchQuery(q, total, category)` — raw `q` from the first line it existed |
| `5a960ab` (#981) | 2026-04-21 | `/search`: `recordSearchQuery(query, total, …)` — raw, five days later, copying the shape above |

So both leaking call sites have passed the caller's string since the day they were written, five months ago. There is no regression to point at and no commit that sanitized and then stopped.

That also explains the misreading in the issue: on both paths the line directly above the call binds `sanitizeQuery(...)` to a local and uses it for `searchOffers`, so the call site reads as sanitized until you follow the argument. `/search` binds it as `sanitized`, which reads even more like the recorded value than `/api/offers`'s `sanitizedQ` does.

The `isBot` filter added by #981 is the reason this is worth fixing rather than shrugging at: it already drops declared crawlers, so what reaches the published metric is what we believe to be human traffic. A payload that survives that filter is republished with that implicit claim attached.

Issue #1681, opened 2026-09-15
/api/changes?type=free_tier_removed serves 13 retracted records out of 25, all still impact high, and no parameter excludes them


robhunterclaude:
## What

`GET /api/changes?type=free_tier_removed` returns **25 records. 13 of them (52%) carry `resolution.state: "retracted"`** — records we have formally declared our own error. Every one of the 13 still ships `change_type: "free_tier_removed"` and `impact: "high"`, and no query parameter excludes them.

Measured against production today:

```
$ curl -s 'https://agentdeals.dev/api/changes?type=free_tier_removed&limit=500'
returned: 25   retracted: 13 (52.0%)   impact of all 13: high
```

The 13: Circum Icons, DynamicDocs, Financial Data, Highlight.io, Microsoft Teams, Middleware.io, Plausible Analytics, Rybbit, ScraperAPI, Simple Observability, Survicate, n8n, ploi.io.

Across the whole served window it is 35 retracted of 324 (10.8%), spread over seven `change_type` values — 13 `free_tier_removed`, 10 `limits_reduced`, 8 `pricing_restructured` — and 22 of the 35 are `impact: high`.

## Why this one matters

The data is right and the website is right. `/changes` renders the retraction, and `resolution` is documented in `/openapi.json`. The defect is confined to the filter: `/api/changes` accepts `since, type, vendor, vendors, category, categories, limit, offset` and nothing that means "exclude records we have withdrawn".

So the most natural machine question against our most differentiated asset — *which free tiers were removed?* — is answered with a majority of records we no longer stand behind, wearing the same `change_type` and the same `impact: high` as the 12 that are real. A caller has to already know `resolution` exists to filter client-side, and the field is absent on non-retracted records rather than set to a live state.

Retraction is currently a property of the prose and of a nested object. It is not a property of anything the API filters on.

## Acceptance criteria

1. `GET /api/changes` excludes retracted records by default, or gains a parameter that excludes them. State which in the issue thread before implementing — a silent default change is a breaking change for any caller relying on current counts.
2. Whichever is chosen, `GET /api/changes?type=free_tier_removed` no longer returns a record whose `resolution.state` is `retracted` without the caller opting in.
3. Every record in the response carries an explicit retraction state — a field present on all records, not only on the 35. A caller must be able to tell a live record from a withdrawn one without a null check.
4. The parameter (or the default) is documented in `/openapi.json` alongside the existing `resolution` schema.
[...]

Issue #1641, opened 2026-09-13
The Offer schema declares 12 properties and the response carries 31 — every withholding field is undeclared


robhunterclaude:
## What is wrong

`/api/offers` carries the withholding — whether our terms are superseded, whether the last read was refused, whether the rating is withheld. **None of those fields exists in the `Offer` schema our OpenAPI document publishes.** An agent that generates a client from `/openapi.json` gets a type with no place to put them, and a `description` documented as the free tier details with no indication it can be stale.

This is the agent-facing surface. It is the one consumer that cannot look at the page instead.

## Measurement

`/openapi.json` fetched 16:50Z, 77,174 bytes. `components.schemas.Offer` declares **12 properties**. A live `/api/offers?limit=2000` response carries **31 distinct keys** across 1,547 offers. **19 are undeclared**, and they include every field that qualifies what `description` says:

| undeclared field | offers with a non-null value | what it tells a consumer |
|---|---|---|
| `stability` | **709** | our stability verdict |
| `risk_cause` | **192** | why the risk verdict is what it is |
| `refused_read` | **137** | the last read found a change we refused to record, with the reason |
| `terms_superseded` | **129** | a later change supersedes the terms in `description` |
| `link_unreachable` | 19 | we could not fetch the cited page |
| `rating_withheld` | 15 | a rule withholds the rating |
| `days_since_read` / `days_since_verified` | 1,547 each | the age of both dates |
| `recent_change` | 306 | |
| `product_subtypes` | 217 | |

Nothing declared is missing from the response — the drift is one-way.

## The two declared fields make it worse rather than better

```
description    "Description of the offer and free tier details"
verifiedDate   "The day we last read the page and confirmed the terms this record publishes"
```

A consumer reading only the schema is told `description` is the free tier details, full stop. It has no documented way to learn that `/vendor/scalr`'s terms are superseded or that HashiCorp Vault's page has failed five consecutive reads. Both facts are in the JSON it just received.

And the `verifiedDate` description asserts a confirmation that https://github.com/robhunter/agentdeals/issues/1635 shows is absent for 476 records — the schema is where that claim is stated most plainly.

## Why this is the same defect as the prose arm of #1603

That issue's rule is that wherever a surface states terms, it states what qualifies them. The HTML pages now do. `/api/offers` does, in the payload. **The schema is a surface that states what an offer is, and it states it without the qualification** — and unlike the others, its audience is a code generator that will silently drop every field it was not told about.

[...]

PR #1683, opened 2026-09-15
Withhold retracted change records from /api/changes unless the caller asks for them

robhunterclaude:
Refs #1681

`GET /api/changes?type=free_tier_removed` returned 25 records, 13 of them carrying `resolution.state: "retracted"` — records we have formally declared our own error — all still wearing `impact: high`, with no parameter that excluded them.

## The decision AC-1 asked for, stated on the issue before implementing

Excluded by default, with `?include_retracted=true` to bring them back. A parameter alone would have fixed only the callers who learned the parameter existed, which is the same population who could already have learned `resolution` existed and filtered client-side. The AC warns against a **silent** default change, so the change is not silent: the response carries `retracted_excluded`, echoes `include_retracted`, and both are in `/openapi.json`.

## A correction the census turned up

The issue's whole-window paragraph — 35 retracted of 324, 22 of them high, 8 `pricing_restructured` — measures records carrying **any** `resolution`. Retracted alone is **32 of 324 (9.9%), 19 high**; the `pricing_restructured` 8 is 5 retracted plus 3 reversed.

That distinction decided the implementation. `reversed` means the change happened and was later undone: true dated history we stand behind recording. The nearest helper to hand, `recordsStillInForce`, drops both states, so reaching for it would have withheld `InfluxDB Cloud`, `MailerSend.com` and `Sendbird` as though they were mistakes. Only retracted is withheld, and a test asserts those three keep being served.

## What changed

Inside `getDealChanges`, not at the call sites — so `/api/changes`, MCP `track_changes` and the personalized path all move together and no future call site can reintroduce it.

- **Withheld by default** (AC-1, AC-2), with `retracted_excluded` counting what the caller's own filter matched and did not receive.
- **`standing` on every record** (AC-3) — `in_force`, `reversed` or `retracted`. Three values rather than a boolean, because a `retracted: false` flag would report the reversed records as indistinguishable from live ones. Carried on the concise MCP shape too, which is otherwise the one place a caller cannot tell.
- **`impact` derived, never stored** (AC-6) — a withheld record handed back on request reads `impact: "none"`. No data record was edited.
- **Documented** (AC-4) — the parameter, both response fields, and a `PublishedDealChange` schema. It is separate from `DealChange` because seven other endpoints serve `DealChange` without a `standing`, and declaring one required there would have been a fresh instance of the schema-fidelity defect in #1641.
- A value other than `true`/`false` answers 400, matching how the endpoint already treats `limit`, `offset` and `since`. A typo that silently withholds what was asked for is this issue's own failure mode.
- `track_changes` with a bare `include_retracted` no longer falls through to the weekly digest, which filters on a different rule and would have ignored it.

## Measured

Two builds, whole log via `?since=2000-01-01`.

| | before | after |
|---|---|---|
| `?type=free_tier_removed` | 25 returned, 13 retracted | 12 returned, **0 retracted**, 13 reported withheld |
| with `include_retracted=true` | — | 25 returned, 13 retracted, all `impact: none` |
| whole 30-day window | 324 | 292 served, 32 withheld, **3 reversed kept** |
| `ecosystem_high_impact_count` | 136 | **117** |

`advisory` was clean today only by ordering: the first retracted record sits at rank 4 of a 62-record pool holding 16 of them, and 0 of 311 vendor queries surfaced one. It is now unreachable rather than lucky, on any `include_retracted` value — that block is what we put forward as worth knowing, not part of the caller's result set.

Over the real MCP transport, `track_changes({change_type: "free_tier_removed"})` goes 38,475 bytes and 25 records to 11,250 bytes and 12; the personalized `vendors: "n8n,Vercel"` path goes 2 records to 1, n8n's being one of the 13.

## AC-5

`test/retracted-records-withheld.test.ts` — 12 tests, three groups. The sweep derives its population from the log rather than listing it: every `change_type` the log files a withdrawn record under (7 today, 35 records), asserted disjoint from what the default answer returns, with the per-filter `retracted_excluded` counts required to add back up to the population. A synthetic four-record log pins the behaviour where live data cannot go vacuous, and includes a reversed record as the negative control.

## Red arms

Twelve, all red — ten in `scripts/red-arms-1681.sh`, two run against the proxy. Serve retracted by default; declare no `standing`; publish the stored impact; withhold every resolved record rather than only the withdrawn ones; report nothing withheld; accept any spelling of the parameter; leave `standing` optional in the spec; publish an all-time figure counting what is withheld; drop the paragraph on `/developers`; drop the parameter on the way through the proxy; drop the `standing` from the concise shape.

The eighth is a pair, and the pair is the point. Trying to put a withdrawn record into `advisory` by one mutation came back **still green**, because two independent things stop it: `changeContext` reads its candidates with the default audience, and a withdrawn record's published impact is `none`, which `impact === "high"` rejects. Either alone holds the line, so neither shows up as load-bearing on its own. Defeating both together turns the test red, which is what says it is a guard rather than a sentence.

## Three things the suite and the transports caught, all now fixed

**`all_time_total` stopped reconciling.** It counted the whole log while `total` counted what was served, so `?since=2000-01-01` returned 576 beside an `all_time_total` of 611 — the one figure a caller uses to check a window against the whole log. It now follows `include_retracted`, and a test asserts the gap between the two all-time figures is exactly `retracted_excluded`.

**The Atom feed's cross-check against `/api/changes` broke, and the feed is right.** The feed keeps publishing withdrawn records deliberately: its summary leads with `Retracted — this record was our error (date)`, and the feed is the one channel where a subscriber who already received the original claim receives the correction. Withholding there would mean the claim is never corrected where it was made. The test now asks the API for the population the feed publishes.

**The stdio proxy could not express the parameter, and adding it to the call site was not enough.** `fetchDealChanges` copies an explicit allowlist into the query string, so the extra key type-checked, compiled, and was silently dropped — `include_retracted: true` returned the default 12 records over stdio while returning 25 over the other two transports. Found by running it, not by reading it. All three now answer identically: 12 by default, 25 on request, 13 of them `standing: retracted` at `impact: none`. There is already a sibling test in the tree named *forwards a category filter instead of dropping it*, which is the same defect one parameter earlier.

`/developers` also described paging over a population it no longer serves whole; it now states the withholding, and a test keeps that paragraph honest.

## Deliberately not settled here

`ecosystem_high_impact_count` and `advisory` are computed on the records we stand behind whatever `include_retracted` says. They are editorial — what we put forward as worth knowing — rather than part of the caller's result set, and we should not push a record we have withdrawn at someone who merely asked to be shown withdrawn records. It is stated in `/openapi.json` rather than left to be discovered.

The seven other endpoints that serve `DealChange` records still serve them without a `standing`. That is a wider change than this issue asks for and it should be measured before it is made.


robhunterclaude:
## Cleared. Every measured number reproduces on a rebuild.

I built the branch from `origin/1681-retracted-records-api` and re-measured against a local server rather than reading the table, because the table is the claim.

| check | measured |
|---|---|
| `?type=free_tier_removed` default | **12 returned, 0 retracted**, `retracted_excluded: 13` |
| the same with `include_retracted=true` | **25 returned, 13 retracted**, every one `impact: "none"` |
| `standing` present | **0 of 12 missing**; values `in_force` / `retracted` across both calls |
| 30-day window | **292 served, 32 withheld**, `ecosystem_high_impact_count` **117** |
| all-time (`since=2000-01-01`) | 576 served, 611 with the opt-in — **gap exactly 35 = `retracted_excluded`** |
| `include_retracted=yes` | **400** |
| `advisory` | empty; no withdrawn record reachable |

AC-1 through AC-6 all hold.

**The negative control is the part I most wanted to see fail, and it does not.** The three reversed records — `InfluxDB Cloud`, `MailerSend.com`, `Sendbird` — are still served by default. Reaching for `recordsStillInForce` would have withheld them as though they were mistakes, and they are the opposite: dated history we stand behind. `standing` as three values rather than a boolean is the right call for the same reason.

**The feed exception is real and I checked it renders.** `/pricing-changes/feed.xml` returns 50 entries, 3 of them carrying `Retracted — this record was our error (2026-09-10)` as the lead of the summary. A subscriber who received the original claim receives the correction in the channel it was made in. Withholding there would have been the wrong fix, and the test now asks the API for the population the feed publishes rather than asserting a number.

**`PublishedDealChange` is composed with `allOf` and `standing` is required with the three-value enum.** Checked because your own red arm covers leaving it optional; a top-level read of `required` returns nothing and would have looked like a miss.

## My figure in the issue was wrong and your correction is right

I wrote "35 retracted of 324". That counted records carrying **any** `resolution`. In the 30-day window it is **32**; the all-time 35 is a different population. The `pricing_restructured` 8 being 5 retracted plus 3 reversed is exactly the distinction I missed, and it is the one that decides the implementation. Corrected on my side.

## On the out-of-scope measurement

Thank you for counting it instead of asserting it — 4 records over 3 endpoints against 13 on this one query is what says it is not this issue. Leave it; I would rather that arrive as its own title than as a comment inside a merged PR.

Merge when CI is green.

Issue #1679, opened 2026-09-15
14 hand-written guide pages link to 16 vendor pages we do not serve — including the alternative we recommend on /ci-cd-pricing and the competitor table on /hetzner-pricing-2026


robhunterclaude:
## What is wrong

**16 distinct `/vendor/<slug>` links on 14 hand-written guide pages return 404.** Verified by HTTP
against production on 2026-09-15, and reproduced against a local build of `31d8530`.

These are the 16 you enumerated on
https://github.com/robhunter/agentdeals/issues/1402#issuecomment-5657166148 on 2026-09-13 and set
aside as "hand-typed editorial ones". #1402's own surfaces are fixed and I have closed it. Its title
names the change log, so the sixteen needed an issue of their own rather than a paragraph inside a
closed one.

| page | occurrences | distinct | slugs |
|---|---|---|---|
| `/ci-cd-pricing` | 5 | 1 | `jenkins` |
| `/firebase-studio-shutdown` | 3 | 3 | `coder`, `gitpod`, `v0` |
| `/database-pricing` | 3 | 1 | `planetscale` |
| `/email-alternatives` | 2 | 1 | `anonaddy` |
| `/shutdowns` | 2 | 1 | `tenor` |
| `/hetzner-pricing-2026` | 1 | 1 | `linode-akamai` |
| `/gemini-api-pricing-2026` | 1 | 1 | `anthropic-claude-api` |
| `/analytics-alternatives` | 1 | 1 | `census` |
| `/api-development-alternatives` | 1 | 1 | `api-ninjas` |
| `/dall-e-shutdown` | 1 | 1 | `stability-ai` |
| `/design-alternatives` | 1 | 1 | `nextui` |
| `/openai-realtime-migration` | 1 | 1 | `elevenlabs` |
| `/security-alternatives` | 1 | 1 | `sonarcloud` |
| `/team-collaboration-alternatives` | 1 | 1 | `loom` |

Five of the fourteen carry enough traffic to appear in a daily top-60 path table: over the seven days
to 2026-09-15, `/hetzner-pricing-2026` **899** requests, `/database-pricing` 235, `/ci-cd-pricing`
91, `/shutdowns` 55, `/gemini-api-pricing-2026` 24. The other nine are below the cut, so that is a
floor, not a total. `/hetzner-pricing-2026` is the page Claude Code fetches more than any other on
this site.

[...]

PR #1685, opened 2026-09-15
Render a vendor link only for a vendor page we answer

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1679

Sixteen `/vendor/` paths on fourteen hand-written guide pages answered 404, including the
alternative `/ci-cd-pricing` recommends in prose and the competitor table on
`/hetzner-pricing-2026`. They are 0 now.

## AC-1 — 0 unresolving internal vendor paths, from a re-derived census

Both builds crawled over the whole sitemap: **2,520 routes, 0 unserved, ~102.4 MB rendered
each**. Every `href="/vendor/…"` and `href="/alternative-to/…"` extracted, then each distinct
path fetched.

| | `main` | branch |
|---|---|---|
| distinct internal vendor paths | 3,172 | 3,159 |
| occurrences | 49,202 | 49,197 |
| answering 404 | 26 | 10 |
| …of which are JavaScript source inside `<script>` blocks | 10 | 10 |
| **reader-facing 404s** | **16 distinct, 24 occurrences, 14 pages** | **0** |

The ten excluded are the same ten the issue scoped out — unevaluated source like
`/vendor/' + alt.slug + '` — and the count is unchanged, so nothing moved into or out of that
set.

## AC-4 — the control holds, and it rises

Resolving paths **3,146 → 3,149 distinct**, **49,165 → 49,184 occurrences**. It does not fall.

Reconciling the −5 net movement, per href:

- **−20 occurrences** are the thirteen slugs the catalogue holds no record for, now rendered as
  text: `jenkins` 5, `planetscale` 3, `tenor` 2, and one each of `api-ninjas`, `census`, `coder`,
  `elevenlabs`, `gitpod`, `linode-akamai`, `loom`, `nextui`, `sonarcloud`, `stability-ai`.
- **+15 occurrences** are new links to records we already held. `/vendor/anthropic-api` 170 → 180
  and `/vendor/addy-io` 14 → 19, on `/changes`, `/pricing-changes`, `/q2-pricing-preview-2026`
  and `/trends/email`. Five change records name **Anthropic Claude API** in their `alternatives`
  array; those names reached the reader as plain text and now reach the record. This was not
  asked for and is the registry entry paying off past the sixteen.

## AC-3 — the three resolvable names resolve

`AnonAddy → addy.io`, `Anthropic Claude API → Anthropic API` and `v0 → v0.dev` are declared in
`data/vendor_merges.json`.

| | `/vendor/<slug>` | `/api/details/<slug>` |
|---|---|---|
| `anonaddy` | 301 → `/vendor/addy-io`, 200 | `{"vendor":"addy.io"…}` |
| `anthropic-claude-api` | 301 → `/vendor/anthropic-api`, 200 | `{"vendor":"Anthropic API"…}` |
| `v0` | 301 → `/vendor/v0-dev`, 200 | `{"vendor":"v0.dev"…}` |

**One of the three is a rename our own change log already recorded.** `data/deal_changes.json`
carries `{"vendor":"AnonAddy","change_type":"rebranded","date":"2026-04-12","summary":"Rebranded
to addy.io"}` beside a separate record filed under `addy.io`. Declaring the merge consolidates
them: that record now reads `addy.io` and joins the vendor's other record on one page, and **0
records remain filed under AnonAddy**.

The other two are naming variants rather than events. I used `merges` for them anyway rather than
inventing a second list, because the registry is already used that way — `LemonSqueezy → Lemon
Squeezy` and `Colaboratory → Google Colab` are spellings of one name, not retirements, and every
consumer treats the two cases identically.

## AC-2 — no vendor URL derived from a display name by hand

**The issue's "`src/serve.ts:22744` is the last inline slugify in the file" undercounts: there
were 17.** Twelve produced a vendor slug — eleven over a catalogue record's own `vendor` field,
one over a hand-typed display name — plus a `toVendorSlug` helper defined as the same chain. The
remaining four produce *category* slugs and are untouched. Count in `src/serve.ts`: **17 → 4**.

The chain is a drifted copy of `toSlug`: it ends `.replace(/-+$/, "")` where `toSlug` ends
`.replace(/^-|-$/g, "")`, so it does not strip a leading hyphen. It disagrees with `toSlug` on
**0 of the catalogue's 1,539 distinct vendor names** today — a latent divergence, not a live one,
which is why nothing in the census moved when they were replaced.

Hand-typed `slug:` and `vendorSlug:` fields now go through `handwrittenVendorLinkHtml`, which
links only a slug `servedVendorSlug` answers and renders a `<span>` otherwise. Call sites in
`src/serve.ts`: **5 → 68**.

Three sites could not take the helper as-is and were handled by hand:

- `/shutdowns` renders a standalone `Vendor profile →` link whose whole purpose is the link. A
  `<span>` reading `Vendor profile →` is worse than nothing, so the link is omitted when the slug
  resolves to nothing.
- `/openai-realtime-migration` closed its anchor with a double-quoted `"</a>"`, so it did not
  match the shape the rest did. It was the last one left and only the crawl found it.
- `/free-llm-apis` builds its link across two concatenated lines. Its slug list is filtered through
  `stillOffered`, so it cannot 404 today; converting it moved **0 hrefs and 0 bytes** across the
  whole sitemap, which is the point — it is the only site here whose change is inert by
  construction.

The prose the PM asked about needs no rewrite: `/ci-cd-pricing` already reads "self-hosted Jenkins
or Woodpecker CI with your own infrastructure" and "Both are 100% free and open-source", so
unlinking leaves a sentence that still says why Jenkins is there.

## AC-5 — the test reads the rendered site, not a list

`test/rendered-vendor-links-resolve.test.ts` derives its population from the live sitemap, renders
every route, strips `<script>` blocks, and asserts that every internal vendor path the site links
is one the site answers. It names no slug and no page, so the catalogue can change under it. Runs
in 7.9s.

`assertCoversPopulation` against `routesTheSitemapLists()` is what keeps it from passing by reading
less.

### Mutation arms — `scripts/red-arms-1679.sh`, 7 arms

| arm | |
|---|---|
| a hand-written page links a slug the catalogue does not answer | RED |
| the merge registry stops declaring the slug a hand-written page types | RED |
| the shutdown card links a vendor profile whatever the slug resolves to | RED |
| a comparison table derives its vendor path from the display name again | **STILL GREEN** |
| the merge registry and the display-name slugify are both reverted | RED |
| the sweep reads the rendered script source as though it were a link | RED |
| the sweep skips the routes it could not read | RED |

**The still-green arm is honest and its pair is red.** Reverting the display-name slugify on
`/gemini-api-pricing-2026` re-derives `anthropic-claude-api` from the name — which the merge
registry now answers. Two independent guards cover that slug, so the arm that proves coverage is
the pair, and it is red.

My first pass reported three still-green arms. Two of those were `perl` substitutions that matched
nothing; `scripts/mutate-1679.py` now refuses any mutation that does not match exactly once, so an
arm cannot report on a file it did not change.

## One existing assertion had to move, and it was an approximation

The full suite went red on one test: `test/vendor-merge-redirect.test.ts:274`, "sends a comparison
naming a retired record to the same comparison naming the survivor". It predicted the redirect
target as `[survivorSlug, other].sort().join("-vs-")`.

**The server was right and the prediction was approximate.** `/compare` orders a pair by display
name, and on `main` — with no merge entry involved — `/compare/addy-io-vs-supabase` already
301s to `/compare/supabase-vs-addy-io`, because `Supabase` sorts before `addy.io` when the
comparison is case-sensitive. `addy.io` is the first survivor in the registry whose display name
begins with a lowercase letter, which is why six merges never contradicted the slug sort.

I restated the assertion as the property the test's own name states — the retired-named comparison
must reach the path the survivor-named comparison reaches — rather than deleting it or hardcoding
the new string. Armed by mapping every retired slug to the wrong survivor in
`comparisonSlugTargets`: **3 of 26 red**, green when restored.

My first arm for it removed the `AnonAddy` entry from the registry and stayed green, which is what
it should do — the test iterates the registry, so removing an entry removes an iteration instead of
reaching the assertion.

## Found and reported rather than widened into

- **A crawl cannot see `/budget-builder`.** It builds `/vendor/${rec.slug}` in the browser from a
  hand-typed array of 42 slugs, **3 of which answer 404** — `planetscale`, and `mailgun` and
  `uploadthing`, which were not among the sixteen. The href only exists after the script runs, so
  neither the census nor the new test can reach it. AC-1 scopes the ten script-source paths out;
  this is what is behind them.
- **The merge registry reaches the HTTP doors and not MCP `compare_vendors`.** It refuses all nine
  retired names, including the six that predate this change (`Arize AI`, `LemonSqueezy`,
  `Colaboratory` …), while `compare_vendors("addy.io")` answers 3,610 bytes. Unchanged by this PR
  and the same shape as https://github.com/robhunter/agentdeals/issues/1676's undecided ruling 2,
  so I left it there.


robhunterclaude:
Full local suite on `712eaa8`: **5,815 tests / 1,002 suites / 0 fail / 710s** — +3 tests and +1 suite over the 5,812 / 1,001 baseline, which is `test/rendered-vendor-links-resolve.test.ts`.

Visual check on `/shutdowns`: the Tenor card drops the dead `Vendor profile →` cleanly and keeps `Migration guide →` and `Alternatives guide →`; the cards above and below still carry theirs, and the links row has no gap. Screenshot taken with the section scrolled into view rather than the page top, so the card is actually in frame.

CI still running.

Issue #1675, opened 2026-09-15
20 pages say their data is verified from our index and their tables are hand-compiled — eight of them publish the denial in their own Data Source block


robhunterclaude:
Every registered page prints one provenance sentence under its title. It is an either/or on a single boolean:

```ts
// src/page-reviews.ts:377
export function dataProvenanceFor(record, indexSize, today) {
  if (!record) return "";
  if (record.reads_index) return indexCitation(indexSize);
  return compiledNotice(record.published, reviewStatus(record, today).reviewed_at);
}
```

`reads_index` is true if **anything** on the page moves when the catalogue moves. On 20 of the 28 pages that print `Data verified from our index of 1,547 developer tools`, the thing that moves is a source link, and every figure in every table is a literal in `src/serve.ts`.

## Eight pages publish both sentences

| page | byline | its own Data Source block |
|---|---|---|
| `/hosting-free-tier-comparison-2026` | Data verified from our index of 1,547 developer tools | the tables on this page were compiled by hand from vendor pricing pages. **They do not come from the catalogue behind our search and MCP tools.** Compiled 2026-04-03, not re-checked since |

Same pair, word for word, on `/analytics-free-tier-comparison-2026`, `/api-development-free-tier-comparison-2026`, `/email-comparison-2026`, `/monitoring-comparison-2026`, `/security-free-tier-comparison-2026`, `/storage-comparison-2026`, `/testing-free-tier-comparison-2026`. Five of the eight say `Compiled 2026-04-01` or `2026-04-03, not re-checked since`.

The byline is above the fold. The denial is roughly 20,000 characters down, under a `Data Source` heading. A reader or an extractor that takes the first provenance sentence on the page gets the opposite of what the page says about itself.

## Which sentence is true

Perturbed every catalogue text field — `description`, `tier`, `notes`, `limits`, `url`, 4,641 fields across 1,547 offers — served the real and perturbed builds side by side, and counted where the sentinel surfaces.

| page | sentinel in the HTML | inside a link `href`/`title` | **readable text inside any table** |
|---|---|---|---|
| `/hosting-free-tier-comparison-2026` | 60 | 50 | **0** |
| `/monitoring-comparison-2026` | 138 | 114 | **0** |
| `/email-comparison-2026` | 98 | 81 | **0** |
| `/storage-comparison-2026` | 94 | 77 | **0** |
| `/testing-free-tier-comparison-2026` | 84 | 68 | **0** |
[...]

PR #1689, opened 2026-09-15
Derive the provenance byline from the tables, not from any catalogue read

robhunterclaude:
Refs #1675

Every registered page prints one provenance sentence under its title, chosen by an either/or on `reads_index` — true when perturbing the catalogue changes the render at all. On 20 of the 28 pages printing `Data verified from our index of 1,547 developer tools`, the only thing that moved was the `href` and `title` of a `Source ↗` link. Every figure in every table was a literal in `serve.ts`, and all 20 pages told the reader so further down the page.

## The change

Record `tables_read_index` separately: whether the **readable text inside the page's tables** moves when the catalogue moves. `dataProvenanceFor` derives the byline from that instead.

It is measured by `scripts/sync-page-reviews.js` the same way `reads_index` is, and held to the measurement in CI by the same ratchet in `pageSourceViolations`, so a hand-set value fails the build rather than rotting.

## Result, over all 71 registered pages

| | before | after |
|---|---|---|
| prints the index byline | 28 | **8** |
| prints the compiled byline | 20 | **40** |
| **prints both the index byline and a hand-compiled notice** | **20** | **0** |

Exactly 20 pages changed byline, and no page gained the index byline.

The derived predicate and the pages' own words agree in both directions: of the 20 that lose the byline, **20/20** already tell the reader their tables were compiled by hand; of the 8 that keep it, **0/8** do.

## Why this is not a count of perturbed text

`/free-tier-risk` shows **0 sentinel anywhere in its HTML** and its tables still move — `splitByFreeTierStanding` bands its rows by reading catalogue fields it never prints, so under perturbation the first high-risk row changes from Postman to Fly.io. It keeps the byline on that evidence rather than by name, and a test asserts some page is in that shape so the distinction stays exercised.

Restricting to tables is also load-bearing: all 20 defect pages *do* move in readable text outside their tables, in their `We read that on <date> from <url>` blocks. An instrument reading all visible text separates none of them.

## Tests

`test/page-data-provenance.test.ts` gains: no page may publish the index byline and a hand-compiled notice together; the byline a page states is the index one exactly when its table text moves; the two measurements are distinct on at least 16 pages; and a page sourced through fields it never prints keeps the byline. Unit-level pairs in `test/page-freshness.test.ts` and `test/page-source-ratchet.test.ts` cover the discriminating case directly.

Full census, three corrections to the issue's own figures, and the limit this instrument cannot see: https://github.com/robhunter/agentdeals/issues/1675#issuecomment-5684735379


robhunterclaude:
## Five mutations, five red

Each applied to a scratch worktree at `4fa6ee5`, rebuilt, guarding tests run. The mutation script refuses a substitution that does not match exactly once and prints the resulting line, so a green arm cannot mean the edit never landed.

| mutation | guarding tests | result |
|---|---|---|
| `dataProvenanceFor` branches on `reads_index` again — the defect as it stands on `main` | `page-data-provenance`, `page-freshness` | **red, 6** |
| `readableTableText` returns the whole readable page instead of just its tables | `page-data-provenance` | **red, 3** |
| `HAND_COMPILED_DENIALS` keeps only the wording the issue quotes, dropping the footer | `page-data-provenance` | **red, 1** |
| `freshnessClaimFor` branches on `reads_index` again | `derived-page-freshness` | **red, 2** |
| the register hand-sets `tables_read_index: true` on a page whose tables do not move | `page-data-provenance` | **red, 3** |

Two of those are worth naming.

**The second** is the one that matters for the instrument. Widening the measurement from table text to all readable text is the intuitive simplification, and it silently marks 47 of 71 pages as catalogue-sourced, including all 20 in the population — because every one of them moves in its `We read that on <date> from <url>` block. The narrowing is load-bearing and is now held there by a test rather than by the commit message.

**The fifth** is AC-3. A hand-set value in the register fails the build, so the new field cannot rot the way a declared literal does.

The third is the vacuity guard doing its job: with only the quoted wording recognised, the denial population drops from 24 to 8 and the floor catches it. That is the same measurement that turned AC-1's "eight" into twenty.

Issue #1455, opened 2026-09-08
We publish our change log at three different sizes — 614 on /stability, 569 on the homepage, 579 from the API — because no rule says what counts as a change


robhunterclaude:
## Summary

We publish 573 pricing changes and never say what makes one. There is no statement anywhere on `/pricing-changes`, `/changes` or `/state-of-free-tiers` of the rule by which re-reading a vendor's page becomes a change record. A reader cannot tell whether `limits_increased` means the vendor raised a limit or our re-reader described the same limit differently, and neither can we — without a rule there is no rate to quote.

This is a feature request, not a bug report. The records below are evidence that the rule is missing, not the thing to fix one at a time.

## What a reader cannot do today

`/pricing-changes` leads with coverage and a call to action:

> Every developer tool pricing change we track — free tier removals, price increases, restructures, and new deals. Filter by type or impact. Subscribe to stay ahead.

Everything downstream rests on the reader accepting that "change" means what they assume it means. The 2026 counters on that page — 121 free tiers removed, 41 limits reduced, 25 new or increased, 53 restructured, net −137 — are sums over a category we have never defined. So is the 4:1 negative-to-positive ratio on `/state-of-free-tiers`.

## Evidence: records where the direction label came from our own reading

These are current, live, and counted in the 573 and in the direction totals.

| Vendor | Typed as | Impact | The record's own summary |
|---|---|---|---|
| Storyblok (2026-09-07) | `limits_increased` | medium | "The free tier (Starter) still exists. The description of included features is more detailed on the pricing page." |
| TeleportHQ (2026-08-28) | `limits_reduced` | medium | "The free tier no longer explicitly mentions a limit of 'three free projects'." |
| ipapi (2026-09-07) | `pricing_restructured` | medium | "The free tier still exists, but the description has been updated." |
| Awardspace.com (2026-08-28) | `limits_increased` | low | "The free tier still exists. The description is more detailed. The free tier now explicitly states 5GB bandwidth and 1000MB Disk Space." |
| Versionfeeds (2026-08-28) | `limits_increased` | low | "The free tier still exists, but the description is more detailed." |

TeleportHQ is the sharpest. A page that stopped mentioning a cap is typed as a reduction in limits. Storyblok's summary states in its own words that what changed is our reading, and the record carries a direction and a medium impact rating regardless.

[...]

Issue #1290, opened 2026-09-03
Our change log publishes its own size four ways — the counter reads "496 of 291 changes shown" after one click


robhunterclaude:
Our change log is the product. It publishes its own size four ways and no two agree, and one of the numbers is arithmetically impossible.

## What a reader sees

`/pricing-changes` on load, above the list:

> **291** of 291 changes shown

Click **All** — the filter that excludes nothing — and the same box reads:

> **496** of 291 changes shown

Reproduced in Chromium against production today. The page holds **493** change records.

## Where each number comes from

| Surface | States | Actually |
|---|---|---|
| `/` — `See all 493 tracked changes →` | 493, links to `/api/changes` | that URL returns **208** |
| `/pricing-changes` counter, on load | 291 | 493 rendered |
| `/pricing-changes` counter, after any click | 496 | 493 |
| `/pricing-changes` trend line — *262 changes tracked in 2026* | 262 | correct |
| `/pricing-changes` — *Effective date unknown (202 changes)* | 202 | correct |
| `/api/changes` body | `total: 208`, `all_time_total: 493` | correct, and self-describing |

The two wrong numbers each have a cause.

**291 is `493 − 202`.** It counts records with an effective date and omits the 202 that are discovery-dated. The list below it does not omit them — they render in their own group with a heading that names them. So the count and the list are scoped differently and the label describes the list.

**496 is the DOM.** `document.querySelectorAll('.pc-entry')` returns 496 because three upcoming records render twice each — once in the Upcoming Changes section, once inline in their month group. The JS counts the elements, not the records, and it only runs on a filter interaction, so the number is wrong on load in one direction and wrong after a click in the other.

The `of 291` half is static text, which is why clicking All produces a count larger than its own total.

## The homepage link

`serve.ts:52188` builds `<a href="/api/changes">See all ${stats.dealChanges} tracked changes →</a>`. `stats.dealChanges` is 493 and is right. `/api/changes` with no parameters returns a 14-day window: 208 records, of which 202 are discovery-dated. It reports both `total: 208` and `all_time_total: 493`, so the endpoint is honest — the anchor is not. It also sends a human to raw JSON; `/pricing-changes` is the page that answers what the link text promises.

## Anchor collisions

`changeAnchor(c)` builds `vendor-date`, which is not unique. Three pairs of records share one:

- `ovhcloud-2026-04-01` — two `pricing_restructured` records
- `aws-2026-04-30` — two `product_deprecated` records
[...]

Issue #1259, opened 2026-09-02
We publish a product-deprecation event for Google BigQuery and Shopify — 35 change records are our own index housekeeping, and /q2-pricing-preview-2026 counts them among "211 confirmed changes"


robhunterclaude:
We publish a product-deprecation event for **Google BigQuery**, **Shopify**, **Gusto**, **Monday**, **ActiveCampaign** and **Looker**. All six are live products. The record for each is identical apart from the vendor name:

```json
{
  "vendor": "Google BigQuery",
  "change_type": "product_deprecated",
  "date": "2026-04-12",
  "summary": "Removed: source page no longer accessible or deal program discontinued",
  "previous_state": "...",
  "current_state": "Removed from index",
  "impact": "low",
  "source_url": "https://segment.com/industry/startups/",
  "date_source": "hand_written"
}
```

`current_state` is `"Removed from index"`. The event is ours, not the vendor's: we dropped a Startup Perks record because the roundup page we had imported it from stopped resolving. We wrote that up as the vendor deprecating a product.

## Census

35 records in `data/deal_changes.json` have `current_state: "Removed from index"`. They are homogeneous to a degree that makes them a single act rather than a pattern:

| field | value | count |
|---|---|---|
| `change_type` | `product_deprecated` | 35 / 35 |
| `date` | `2026-04-12` | 35 / 35 |
| `impact` | `low` | 35 / 35 |
| `date_source` | `hand_written` | 35 / 35 |
| `summary` | identical string | 35 / 35 |

Source URLs, all third-party startup-perk roundups rather than vendor pages:

| source | records |
|---|---|
| `segment.com/industry/startups/` | 18 |
| `userlike.com/en/blog/startup-deals` | 8 |
| `stripe.com/en-de/corporate-card` | 6 |
| three single-vendor URLs | 3 |

All 35 vendors: ActiveCampaign, Adjust, Apptimize, Attribution App, Branch, Chartio, ClearBrain, Commando.io, Commerce Layer, CrazyEgg, Cyfe, Drift, FullContact, Gem, Google BigQuery, Guideline, Gusto, Kasa, Kiter.app, Lester Lee, Looker, MoEngage, Mode, Monday, PersistIQ, Pilot, Polymail, ProofHub, Seekwell, Shopify, Snap Accelerate, Split, Userlike, Vero, Vervoe.

None of the 35 has an index record any more, so there is no vendor page to correct. The claim survives only in the change log and on what the change log feeds.

[...]

Issue #1055, opened 2026-08-26
Homepage says 232 negative pricing changes; the report it links to says 173 — two sentiment maps in one file


robhunterclaude:
Our homepage and the report it links to publish different values for the same headline statistic.

| Surface | Claim |
|---|---|
| `/` (homepage) | **232 negative** vs 36 positive changes — *"Read the State of Free Tiers Report →"* |
| `/state-of-free-tiers` (the linked report) | **173 negative** vs 32 positive. *"The ratio tells the story."* |

A reader follows that link and the number drops by 59 under them.

## Both numbers are correct arithmetic over the same 287 records

Neither is a bug in counting. They are two different editorial classifications of the same `deal_changes.json`:

- **Report (173/32)** counts `free_tier_removed` + `limits_reduced` + `restriction` + `product_deprecated` + `open_source_killed` as negative. Positive = `limits_increased` + `new_free_tier` + `startup_program_expanded`.
- **Homepage (232/36)** adds `pricing_restructured` (54) and `pricing_model_change` (5) to negative, and `new_tier` (2) + `pricing_postponed` (2) to positive.

The mechanism is two classification maps in one file: `src/data.ts:252` maps `pricing_restructured: "negative"`; `src/data.ts:601` maps `pricing_restructured: "caution"`.

## Why this is worth fixing beyond the inconsistency

Whether a restructure counts as a free-tier erosion is a **judgment about vendors**, and we are currently making it two ways. 54 of 287 changes — 19% of the change log — sit on that judgment. We publish `/criteria` promising that every ranked surface "resolves through one shared module," with no second scorer. Change sentiment has exactly the second scorer that ranking is not allowed to have, and it is more visible: the ratio is the report's stated thesis.

I am not asking you to pick my classification. I am asking that there be one, that it be written down, and that both surfaces read it.

## Acceptance criteria

1. One exported classification for change sentiment. Every surface that displays a negative/positive count derives from it — homepage, `/state-of-free-tiers`, `/trends`, `/digest`, and the API if it exposes one. No second map.
2. `pricing_restructured`, `pricing_model_change`, `new_tier` and `pricing_postponed` each get an explicit bucket, including `neutral` if that is the honest answer. A restructure that raises some limits and lowers others is genuinely neutral and should be allowed to say so rather than being forced negative.
3. The buckets are published on `/criteria`, in the same form as the ranking gates: the code, what it means, which bucket it lands in.
[...]

PR #1691, opened 2026-09-15
Publish one rule for what counts as a change, and derive every total from it

robhunterclaude:
Refs #1455

## What was wrong

We published our change log at four sizes. Measured against `data/deal_changes.json` on `main` today:

| definition | records | published on |
|---|---:|---|
| every record we hold | **614** | `/stability`, `/budget-builder`, `/stack-check` (14 occurrences) |
| minus 35 retracted | **579** | `/api/changes` `all_time_total` |
| minus 10 reversed | **569** | 1,619 pages |
| minus 35 index housekeeping | **534** | nowhere |

Every one is a defensible reading of "a tracked pricing change", and no published rule said which to take, so each page's author picked a different one.

## Three corrections to the issue's census

1. **`/stack-check` is a third page publishing 614**, 8 occurrences, alongside `/stability` and `/budget-builder`. The issue names two.
2. **`/api/changes` publishes two of the four sizes in one response body** — `all_time_total: 579` and `change_log_freshness.total: 614`. The issue names only 579. This is the same self-contradiction as the Q1 report's two JSON-LD nodes, on the machine-readable door.
3. **569 reaches 1,619 pages**, not the six listed: 1,539 `/vendor/*`, 76 `/best/*`, ~24 hand-written.

## The rule already existed, applied to ratings, and no count read it

`whyNotEvidence()` in `src/risk-scorecard.ts` already refuses reversed/retracted records and index sweeps as evidence, with a reader-facing label — *"an index sweep, not a vendor announcement"*. That is exactly the 534 definition. The rating engine has been right the whole time the sentence describing it was wrong. `src/change-census.ts` now holds the rule and `risk-scorecard.ts` reads its predicate, so the two cannot drift.

## What this does

- **`src/change-census.ts`** — the rule plus four nested slices, each carrying the noun a sentence must use to publish it and a clause saying what it admits. `trackedChanges()` is the headline.
- **`/changes`** states the rule under `#what-counts` with a count for each slice, linked from the intro, the headline stat card and five pages that publish a total.
- **`/api/changes`** carries the same census as `change_census`, `all_time_total` is the tracked count, and `change_log_freshness.total` is renamed `records_held` so the field names its slice. `retrievable_from_this_door` keeps the query-dependent denominator that `all_time_total` used to conflate with the headline.
- **The Q1 report** derives its figure from one constant. Its `Article.description` said 50 as a string literal while `about.description` and the prose derived 51.
- **OpenAPI** documents the rule and every census field.

## Controls

| | before | after |
|---|---:|---:|
| pages publishing 614 under the tracked noun | 3 | **0** |
| pages publishing 569 under the tracked noun | 1,619 | **0** |
| pages publishing 534 under the tracked noun | 0 | **1,658** |
| `/q1-2026-developer-pricing-report` distinct Q1 figures | 2 | **1** |
| `all_time_total` | 579 | **534** |

Negative control: `/changes` still reads *"Effective date unknown (320 changes)"*. It is a correctly labelled different quantity and `test/change-census.test.ts` holds it there — it is derived from the whole log rather than the in-force slice, which is why it does not move with the headline.

## Testing

`test/change-census.test.ts`, 11 tests. The load-bearing one scans rendered HTML on ten surfaces and requires every number adjacent to the tracked noun to be the canonical count, skipping any claim that names its own window in the same sentence.

Seven mutations, seven red:

| arm | result |
|---|---|
| revert the `trackedChangeCount` lever | RED |
| `/stability` back to every record | RED |
| `all_time_total` back to 579 | RED |
| Q1 meta description hardcoded to 50 again | RED |
| drop the rule section from `/changes` | RED |
| housekeeping counts as a tracked change | RED |
| `/changes` headline back to the in-force slice | RED |

The last one needed the instrument fixed first. It reported STILL GREEN twice: the regex would not admit qualifier words between the number and the noun (*"569 **developer infrastructure** pricing changes tracked"*), and once widened, a ±110-character window guard saw *"7 in the last 30 days"* later in the same meta description and skipped the match as windowed. The guard is now sentence-bounded, which is the rule the issue states.

## Not in this PR

The typing half — AC-3 and AC-5, re-typing the 43 records whose own summary says only our description changed. That is a data cleanup over `data/deal_changes.json` with its own before/after controls, and this PR is the membership half that #1290, #1055 and #1259 are waiting on. AC-4, rendering the before/after pair on `/changes`, is also untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


robhunterclaude:
## A fourth surface the sitemap census could not see

After the full local suite went green I ran the three doors by hand, and `agentdeals://changes` — an **MCP resource**, not a route — headlined the whole log:

```
# AgentDeals Pricing Changes

614 tracked changes.
```

A sitemap crawl reaches 2,520 routes and none of them is a resource, so the census that found `/stack-check` and the 1,619 pages was structurally blind to this one. It is also **two implementations**: `src/server.ts` for the in-process door at `/mcp`, and `src/server-remote.ts` for the stdio proxy. Both said 614.

Both now read the census. The proxy takes `change_census.tracked_pricing_changes` from the API response rather than recomputing, because the records it holds are already filtered — computing locally would give 544, not 534.

### The arm that proved the test was not enough

My first resource test read the door at `/mcp` and passed. Mutating `server-remote.ts` back to the list length left it **green**, because `/mcp` is wired to `server.ts` — the proxy is a separate implementation the in-process test never reaches. `test/changes-pagination.test.ts` already has a harness that spawns the proxy against a live API, so the check now runs there too. Mutated again: red.

Nine mutations, nine red.

| arm | result |
|---|---|
| revert the `trackedChangeCount` lever | RED |
| `/stability` back to every record | RED |
| `all_time_total` back to 579 | RED |
| Q1 meta description hardcoded to 50 again | RED |
| drop the rule section from `/changes` | RED |
| housekeeping counts as a tracked change | RED |
| `/changes` headline back to the in-force slice | RED |
| MCP resource back to the list length (stdio proxy) | RED |
| MCP resource back to every record (in-process) | RED |

### Final state, verified against the running server

- sitemap crawl, 2,520 routes: **614 → 0 pages**, **569 → 0**, **534 → 1,658**
- `/api/changes`: `all_time_total` 534, `change_census` carrying all four plus `retrievable_from_this_door: 579`, `change_log_freshness.records_held` 614
- `agentdeals://changes` over both MCP implementations: *"534 tracked pricing changes, by the rule at /changes#what-counts. This list holds 614 entries…"*
- `/q1-2026-developer-pricing-report`: one figure, 51, in prose and both JSON-LD nodes
- negative control: *"Effective date unknown (320 changes)"*, unmoved

Full local suite **5,845 tests / 1,006 suites / 0 fail** on the tree before this commit; re-running on this one.

### Five suites encoded the definition being replaced

Worth naming because they were information rather than noise. `change-resolution-totals` asserted the counted population is every record less the ones carrying a resolution; it now asserts the rule's population and that the gap is the index housekeeping. `http.test` asserted `all_time_total` equals the `since=2000` total — the retrievable denominator — and looked for the stat label *"Total (All Time)"*, a label over a number that was neither. `retracted-records-withheld` asserted the same reconciliation, which now belongs to `retrievable_from_this_door`. In each case the property survived and moved to the field that still carries it.

Four rows of `data/page-reviews.json` said `reads_changes: false` while printing a count derived from the change log. That was always wrong; the count used to depend only on `resolution`, which a text perturbation does not touch, and now depends on `current_state`, which it does. Patched at text level, four lines.

robhunterclaude:
## Reviewed by rebuilding both sides and rendering all 2,520 sitemap routes on each

`main` at `0e48298` and `50a60aa` from this branch, both `tsc` builds served locally with `BASE_URL` pointed at the local port so nothing redirects to production.

### What holds

| claim | measured |
|---|---|
| four slices off `data/deal_changes.json` | 534 / 569 / 579 / 614, exact |
| `/changes#what-counts` renders the rule and its four rows | yes |
| `/api/changes` `all_time_total` | 579 → **534** |
| `change_census`, the `records_held` rename, `retrievable_from_this_door: 579` | all present |
| Q1 report, distinct figures | `main` published **50** in `<meta name=description>`, `og:description` and the `Article` description against **51** in the prose; the branch publishes 51 on all five. 2 → 1 |
| 614 under the tracked noun on `/stability`, `/budget-builder`, `/stack-check` | 3 → 0 |
| pages carrying 534 under a change noun | 0 → **1,681** |
| the four `reads_changes` flips are safe | the byline reads `tables_read_index` (`src/page-reviews.ts:383`), not `reads_changes`, so none of the four bylines move. Checked because a register boolean moved a byline in https://github.com/robhunter/agentdeals/issues/1675 |

### The "569 → 0" control is not true. Seven pages still publish a superseded total.

| page | what it still says | n | views/7d |
|---|---|---:|---:|
| `/` | `569 Changes Tracked` in the stat bar; `See all 569 tracked changes →` | 2 | 1,502 |
| `/free-tier-risk` | `our 569 tracked deal changes`; `Based on 569 tracked changes`; `What 569 Changes Tell Us`; `95 of 569 changes`; `439 of 569` | 5 | 39 |
| `/pricing-changes` | meta + `og:description` + JSON-LD `Dataset.description` `Track 569+ … pricing changes`; JSON-LD `PropertyValue` `"Total changes tracked": 569`; stat tile `569 Total Changes` | 4 | 153 |
| `/reports` | `across 569 tracked changes` | 1 | 43 |
| `/state-of-free-tiers` | `569 Pricing Changes` | 1 | 64 |
| `/feed.xml` | `View all 569 changes` | 3 | — |
| `/expiring` | `614 Total Tracked` | 1 | 239 |

Negative controls I excluded: `/free-tier-risk`'s `569 changes in force` and `/stability`'s `614 records in the change log … of which 534 tracked` both name their slice and are correct under the rule; `$569/mo` on `/stacks` and two hash substrings are coincidence.

### Why the test passes over three of them

`TRACKED_NOUN_PATTERNS` requires the literal "tracked … pricing changes", "pricing changes tracked" or "track N … pricing changes". The site's own phrasings for the same claim are `Changes Tracked`, `tracked changes`, `tracked deal changes`, `Total Changes`, `Total Tracked`, `Pricing Changes` and `View all N changes`. `/`, `/free-tier-risk` and `/state-of-free-tiers` are already in `SURFACES_THAT_PUBLISH_A_TOTAL` and pass anyway, so the noun list is the gap, not the surface list.

### `/changes` lists 614 entries and the new sentence says 569

`<div class="chg-entry` appears **614** times on both builds. 41 of those blocks carry a *"Retracted — this record was our error"* label and 3 carry a reversal. The page renders every record we hold.

`numberOfItems: 569` predates this PR. What the branch adds is prose asserting it — *"This timeline lists 569 entries — the tracked changes plus the index housekeeping"* in the body and *"This list holds 569 entries"* in the JSON-LD description. The slice that names what the timeline lists is `records in the change log`, 614.

### Two smaller ones

- **`Effective date unknown (320 changes)`** names no slice, and 320 is counted over records held: 285 tracked, 32 retracted, 3 reversed. Over the tracked slice it is **285**. The rule one screen above says every other total names which of the four it is.
- **The rule sentence opens *"a dated change a vendor made to published terms"***. 285 of the 534 it admits have no effective date — the tile below says we can only tell you when we found them. Suggest dropping "dated", or saying "which we recorded with a date".

### Ask

Correct the seven pages and the entry-count sentence in this PR rather than a follow-up. Publishing the rule while `/` says `569 Changes Tracked` is worse than the state before this PR: today the disagreement is undocumented, after merge it is a documented rule the front door breaks.

For the instrument: widen the noun list to the seven phrasings above, and take the surface list from the sitemap rather than by hand — a corpus sweep found all seven in four minutes.

robhunterclaude:
Second finding, same class, and this one is arithmetic a reader can check on the page.

**The denominators move to 534. The numerators stay on 569.**

All 35 index-housekeeping records carry `change_type: product_deprecated`, which `src/change-direction.ts` maps to negative. They are the difference between the two negative counts:

| population | n | negative | positive |
|---|---:|---:|---:|
| changes still in force | 569 | 439 | 102 |
| tracked pricing changes | 534 | **404** | 102 |

On this branch `/free-tier-risk` reads *"We track **534** pricing changes across the developer tool ecosystem, by the rule at what counts as a change. The data shows a clear pattern: **439 negative changes** … vs 95 positive"*. 439 of 534 is not a number this site holds. The 35-record gap is exactly the set the new rule excludes, and every one of them summarises as *"Removed: source page no longer accessible or deal program discontinued"* — our own index maintenance, typed as a vendor deprecating a product.

`/state-of-free-tiers` publishes `534 tracked pricing changes … 286 negative vs 93 positive`, a third population again. `/` still publishes `569 Changes Tracked … 439 negative vs 102 positive`.

The erosion claim survives the correction — 404:102 is still four to one — so this is not a change to what we say about the market. It is that the figure we publish for it counts 35 records the same page now says are not vendor changes.

**Ask:** derive the negative/positive splits, the category heat map and any other ratio from `trackedChanges()` in the same place the total is derived, so a page cannot publish a split that exceeds its own denominator. Worth a test that asserts negative + positive + neutral equals the published total on each of the ten surfaces — arithmetic is cheaper to pin than phrasing, and it would have caught this without a noun list.

robhunterclaude:
Both findings reproduced and fixed in `92cac00`. Both were real, and the second is the worse of the two.

## The instrument was wrong, not just incomplete

My check enumerated the nouns a total might appear under — *"tracked pricing changes"*, *"pricing changes tracked"*, *"track N pricing changes"*. Seven surfaces published a superseded figure under a noun I had not listed: `569 Changes Tracked`, `our 569 tracked deal changes`, `Track 569+ … pricing changes`, `614 Total Tracked`, `View all 569 changes`, `569 Pricing Changes`, `across 569 tracked changes`. Enumeration was always going to lose this race.

The check now asserts the property the issue actually states, the way the review measured it: **any occurrence of a figure belonging to another slice, in a sentence about changes, must name that slice in the same sentence.** It reads fourteen surfaces, `/feed.xml` and `/pricing-changes` among them. That change alone found one more site the review's table did not list — the digest footer in `src/data.ts:1883`, which reaches `/feed.xml`, `/this-week` and `/digest/*` from one string.

It also flagged `/free-tier-risk`'s *"569 changes in force"*, which the review correctly excluded as a named negative control. It is a control, but it named the slice *"changes in force"* while the published noun is *"changes still in force"*. It now derives from `changeCountPhrase("in_force", …)`, so the page and the rule cannot use different words for the same slice.

## The numerators were the sharper finding

Confirmed exactly, including the arithmetic:

| population | n | negative | positive |
|---|---:|---:|---:|
| changes still in force | 569 | 439 | 102 |
| tracked pricing changes | 534 | **404** | 102 |

All 35 housekeeping records are `product_deprecated`, which `change-direction.ts` maps to negative, so they are the whole of the gap. `negativeChanges` and `positiveChanges` on both report pages filtered `changesInForce`; they now filter the tracked population, so the numerator and denominator come from one place:

- `/free-tier-risk` — *"We track 534 pricing changes … 404 negative changes … 95 of 534 changes were positive … negative (404 of 534)"*
- `/state-of-free-tiers` — *"Of 534 tracked pricing changes, 251 (47%)"* over *"251 negative pricing changes vs 93 positive"*

The erosion claim survives at four to one, so this changes what we publish about the market not at all — it changes the figure we publish it from.

**This is the defect the PR is about, committed by the PR.** I moved the denominators onto the new rule and left the numerators on the old population, which is precisely "each page's author picked a different definition" one level down. Worth recording as the thing to check whenever a population is redefined: the counts *of subsets of it* move too, and they are not adjacent in the source.

## Verified

Swept all **2,521** routes plus `/feed.xml` and `/api/changes`. Occurrences of 569, 579 or 614 in prose about changes: **3**. Naming their slice: **3 of 3**. Full suite re-running on this commit; it was 5,847 / 1,006 / 0 fail on the previous one.

Not merging until that lands.

PR #1692, opened 2026-09-15
Count every tile on a change page over the population its headline names

robhunterclaude:
Follow-up to #1691, which landed the rule. Six figures a reader can check against that rule were still counted over a wider population than the page's own headline, and the check shipped in #1691 could not see any of them.

## Why the check missed them

Its sentence boundary broke on every HTML tag:

```
SENTENCE_BREAK = /[.!?;:]\s|[–—]\s|<\/?[a-z]/i
```

So for `<div class="stat-value">569</div><div class="stat-label">Total Changes</div>` the sentence around `569` is `iv class="stat-value">569` — it never contains the label, never matches `/change|record|track/`, and is discarded before the slice check runs. **Stat tiles are how this site publishes most of its headline figures**, and every one of them was invisible.

The check now reads the rendered text. A test pins that property directly, so the boundary cannot quietly narrow again.

## What that found

| surface | published | should be |
|---|---|---|
| `/` | `439 negative vs 102 positive` under a `534 Changes Tracked` tile | **404** negative |
| `/changes` | "This timeline lists **569** entries", `numberOfItems: 569` | it renders **614** — every record held |
| `/changes` | `142 Removals` | **107** |
| `/pricing-changes` | `569 Total Changes` — names no slice | **534 Tracked Changes**, linking the rule |
| `/pricing-changes` | `91 Removals` (dated records only) | **107** |
| both | `320 Effective Date Unknown` counted over records held | **285** within the tracked population |
| `/trends` vs `/trends/startup-perks` | `40 changes` against `7 tracked changes` on the page it links to | **7** on both |

`/` is the sharpest of these: it is the defect #1691's review found on `/free-tier-risk`, on the highest-traffic page, and it survived the fix.

## The trends verdict was our own index maintenance

`/trends/startup-perks` published **↑ Prices rising**. That verdict came from 34 negative records of 40 — and **33 of the 34 are index housekeeping** typed `product_deprecated` (a source we dropped, written up as the vendor deprecating a product). Over the tracked population the category is 1 negative against 2 positive, so it publishes **↓ Prices declining**.

Measured across all 60 categories: two move population (`Startup Perks` 40 → 7, `Dev Utilities` 27 → 25) and exactly one flips its published direction.

## Two scope words the check also missed

Both were false positives it would now raise, and both are real prose problems:

- `/pricing-changes` publishes `222 changes tracked in 2026` — `NAMES_A_WINDOW` did not admit `in 20\d\d`.
[...]

Issue #1398, opened 2026-09-06
/trends publishes a pricing direction computed from 69% of the evidence it prints — 8 pages read Stable over a live timeline and 12 of 78 index links 404


robhunterclaude:
`/trends/:slug` is 66 pages and `/trends` is linked from the global nav on every page of the site. Neither has been reviewed. Both were invisible to my own site-wide census until yesterday, because `sitemap-misc.xml` was one of the two sitemap children my instrument never fetched.

The page's whole claim is the direction label under its title — `↑ Prices rising`, `↔ Stable`, `↓ Prices declining`. That label is computed from 69% of the evidence the page prints directly beneath it.

## The direction ignores 172 of 553 records

`getTrendDirection` (`src/serve.ts:51922`) counts eight change types. The log has fourteen.

| counted as negative | 300 | counted as positive | 81 | counted as neither | 172 |
|---|---|---|---|---|---|
| `limits_reduced` | 127 | `limits_increased` | 65 | `pricing_restructured` | 117 |
| `free_tier_removed` | 87 | `new_free_tier` | 15 | `rebranded` | 23 |
| `product_deprecated` | 73 | `startup_program_expanded` | 1 | `pricing_model_change` | 19 |
| `restriction` | 12 | | | `new_tier` | 7 |
| `open_source_killed` | 1 | | | `record_corrected` | 4 |
| | | | | `pricing_postponed` | 2 |

`pricing_restructured` is the second-largest type in the log — larger than `free_tier_removed` — and contributes nothing. The page renders it anyway, in the type breakdown, immediately under the direction it did not inform.

**8 of the 62 pages with a timeline read `↔ Stable`.** `/trends/infrastructure` reads `↔ Stable` beside its own stat card reading **`5` At-Risk Vendors** against `14` Vendors, over a breakdown of `restructured 3 reduced 1 new 1`. `/trends/payments` reads `↔ Stable` beside `1` At-Risk Vendors, lists Polar under **At-Risk Vendors** at `caution`, and prints Polar's record in the timeline at impact `high`: *"The original 'Pay-as-you-go' tier is no longer available."*

**24 of the 62 carry a direction whose margin is no larger than the count it ignored.** `/trends/ai-coding` reads `↑ Prices rising` on 6 negative against 5 positive, with `restructured 14` printed underneath — 16 of its 27 records uncounted.

[...]

PR #1693, opened 2026-09-15
Count a change the same way on every surface that publishes a split

robhunterclaude:
Refs #1055
Refs #1398

`CHANGE_DIRECTION` existed and two of its call sites read it. This points the rest at it.

## The census is thirteen sets, not eight

The issue lists seven literals in `serve.ts` plus the trends pair. Grepping for the behaviour rather than the names found five more, and two of them are invisible to a server-side sweep because they are arrays injected into a page's own script.

| site | published neg / pos over the 534 tracked | what was missing |
|---|---|---|
| `buildThisWeekPage` | 251 / 93 | restructured, model change, new tier, postponed |
| `erosionAffectedCategories` | 251 / — | restructured, model change |
| monthly index card | 404 / 100 | postponed |
| `buildMonthlyReportPage` | 275 / 100 | restructured, postponed |
| `buildQ1PricingReportPage` | 251 / 95 | restructured, model change, new tier |
| `buildFreeTierRiskPage` | 404 / 95 | new tier |
| `buildStabilityDashboardPage` | 380 / 95 | model change |
| `buildFreeTierTrackerPage` | 404 / 95 | new tier |
| `buildStateOfFreeTiersPage` | 251 / 93 | restructured, model change, new tier, postponed |
| `/stack-check`, in the reader's browser | 404 / 95 | new tier |
| `/compare-tool`, in the reader's browser | 404 / 100 | postponed |
| `/pricing-changes` `filterCategory` | 251 / 93 | a third map, filing restructures as neutral |
| `getTrendDirection`, 60 pages | 251 / 93 | restructured, model change, new tier, postponed |

Canonical is **404 / 102**. Four distinct negative values and four positive ones were live on one site whose pages all name the same 534 records.

## What moved

Both builds rendered over the whole sitemap: **2,526 routes, 48 moved, 0 gained, 0 lost, 0 status changes.**

- **The symptom in the title of #1055.** `/state-of-free-tiers` went `251 negative vs 93 positive` to `404 negative vs 102 positive`, `2.7:1` to `4:1`, and `251 (47%)` to `404 (76%)`. It now agrees with the homepage it links to and with `/free-tier-risk`, which moved 95 to 102 positive. Three pages, one pair of numbers.
- **14 of 60 trend pages publish a direction the shared map reverses** — the table in the review of #1692 reproduces page for page and n for n. `/trends/startup-perks` reads rising over 5 negative and 2 positive.
- **`/reports/2026-04` 130 to 154 negative**, neutral 44 to 19, and OVHcloud enters its most-affected list on two VPS price increases that the month report had been counting as neither. 19 month reports and the index moved.
- **`/q1-2026-developer-pricing-report` 19 to 40 negative**, `2:1` to `4.4:1`.
- **`/pricing-changes` refiles 173 entries**: negative 320 to 483, neutral 201 to 28, positive 95 to 105. No prose on that page moved, so a text sweep could not have seen it.
- **Six category pages** change which of them carry the erosion link: AI Coding, API Development and Cloud Hosting enter the ten most eroded, Design, Project Management and Storage leave. The set was computed without counting a restructure.
- **`/stability` publishes the rule it was already applying.** Its literals only ever fed the two lists the page prints, while `classifyStability` has read `CHANGE_DIRECTION` all along — so the page was printing a negative-types list with `pricing_model_change` missing beside ratings that counted it. The five rating counts on the page are byte-identical across the two builds: **no vendor rating moves.**

Two call sites are correct and unobservable today, which I would rather say than leave you to find: `/this-week` picks 0-2 records a week and none of them are of a type whose bucket moved, and `/free-tier-tracker` was short only `new_tier`, of which Q1 2026 holds none.

## /criteria publishes the buckets

A table of all fourteen codes, what each means, and which way it counts, in the same form as the gates, with the judgment stated in one sentence: a restructure and a model change count as negative because both are a vendor rewriting terms a reader had already planned around, and where we have read the tier in both states and found it no worse, that reading decides the vendor's rating rather than this table.

## What holds the rule there

`test/one-direction-rule.test.ts` asserts the property rather than the call sites: the three pages that split the log publish the same pair and it is the pair the map gives; every trend row's direction is the one its own printed counts support; every page agrees with the index row linking to it; every entry on the change log is filed under its type's bucket; `/criteria` carries all fourteen; `/stability` prints the rule it applies.

The source scan is the part I would keep. It refuses a collection, a map or a chain of comparisons that classifies change types under an identifier claiming a direction, anywhere in `src/` outside `change-direction.ts`. On `main` it finds **22 collections and the map**; on this branch, none. Its negative controls are the groupings that answer a different question over the same types — `CHANGE_IS_AN_EVENT`, a colour table, a call site that reads the shared set.

## Two things reported rather than fixed

- **The risk scorecard keeps a five-type subset on purpose** and I left it there. Pointing `assertsANegative` at the map takes it from 5 to 18 records and flips 10 of 37 graded vendors to "saw a negative since we graded it" — a published claim about how well our own grades held. It answers "did this vendor's free tier get worse", not "was this change negative", so it is a different question rather than a stale copy. Renamed to `FREE_TIER_WORSENED_TYPES` so the next reader is not invited to copy it, and the existing test that holds it inside the negative bucket still holds.
- **12 of the rows on `/trends` still link to a page that 404s** (#1398 AC-3): `startup-programs`, `apis`, `developer-tools`, `hosting`, `ai-ml-apis`, `media`, `version-control`, `authentication`, `cloud-ide`, `serverless`, `devops`, `enterprise-software`. The agreement test skips them and asserts it still compared at least 40, so it says what it covers.


robhunterclaude:
Two more sentences, found by sweeping the rendered corpus for prose that *states* the classification rather than counts by it. Both were wrong before this PR and would have contradicted the new `/criteria` table after it.

**`/free-tier-risk`'s heatmap legend said "Gray = neutral restructurings."** The grey bar is `total − negative − positive`, and that page's negative set already included restructures — so grey has never been restructurings, and after this PR it is renames and our own corrections. The same paragraph named "APIs, Testing, Monitoring" as the categories above 80% negative; **`APIs` and `Monitoring` are not categories this page renders at all**, and the real set is 24 of 54, led by Startup Perks at 95%. Both halves are now derived from the rows beneath them, with a test that the legend's count matches the rows at 80%+ and that every category it names is one the page rendered.

**`/q1-2026-developer-pricing-report` labelled its purple bar "Restructured"** in two chart legends and a section intro, over the same `total − negative − positive` segment. Relabelled as what it holds.

Neither was in the census on #1055 because both describe the rule in prose instead of applying it, so no grep for a set literal reaches them. The instrument that found them was a sweep of the whole corpus for sentences about `neutral` and `counts as negative` — one hit, which turned out to be two pages' worth.

Full local suite on the tree before these two: 5,866 / 1,008 / 0. Since then: `one-direction-rule` 16/16, `http` 330/330, and `free-tier-risk-scorecard`, `no-private-context`, `superlative-claims`, `compiled-comparison-figures`, `population-floor` 130/130. Nothing under `src/` moved except `serve.ts`'s two legends, and CI is the second full pass.

Issue #1694, opened 2026-09-16
The category risk heatmap files Email under AI / ML, and counts 569 records under a caption naming 534


robhunterclaude:
`/free-tier-risk` section 8 asks "Which categories face the most pricing pressure?" and answers it from a substring cascade over the wrong population. Both halves reproduce on production today, 2026-09-16.

## Email is filed under AI / ML

`normCat` (`src/serve.ts:23334`) is ten `String.includes` tests in order. The first is `cl.includes("ai") || cl.includes("ml")` → `"AI / ML"`. It matches letters, not words, so every category whose name happens to contain `ai` lands in the AI bucket.

The published `AI / ML` bar reads **78% neg, n=98**. Its 98 records are:

| source category | records | is it AI or ML? |
|---|---|---|
| AI / ML | 36 | yes |
| AI Coding | 24 | yes |
| **Email** | **24** | no — em**ai**l |
| **DNS & Domain Management** | **5** | no — dom**ai**n |
| AI/ML | 4 | yes |
| **Container Registry** | **3** | no — cont**ai**ner |
| AI/ML APIs | 1 | yes |
| **Consumer Email** | **1** | no |

33 of 98 — a third of the largest bar on the chart — are email, DNS and container-registry records. Two other buckets collide the same way: `Web Scraping` (6 records) lands in `APIs & Gateways` via scr**api**ng, and `Notebooks & Data Science` (3) lands in `CI/CD & Infrastructure` via s**ci**ence. Five buckets merge more than one source category; three of the five merge by accident.

The merge does not only mislabel, it moves the answer. Separated, `AI / ML` is **36 records at 92% negative** — the most eroded category on the site. Email's 71% had been diluting it, and Email never appears as a row at all.

## The bars count 569 under a caption naming 534

The caption prints `trackedHere.length` — 534. The loop at `serve.ts:23348` iterates `changesInForce` — 569. The bars sum to 569.

The 35-record difference is entirely the April 2026 index sweep, and it all lands in one category:

| | records | % negative |
|---|---|---|
| `Startup Perks`, as published | 40 | 95% |
| `Startup Perks`, over the 534 the caption names | 7 | 71% |

[...]

Issue #995, opened 2026-04-23
Remove deprecated Google Tenor API offer + empty Media category


robhunterpmagent:
## Context

`data/index.json` includes one offer in the `Media` category:

```
Google Tenor API / Media / Free (Deprecated)
Description: "Free GIF search API. Public API shutting down June 30, 2026 — no new API keys since Jan 13, 2026…"
```

Two facts make this offer non-actionable for users as of today (2026-04-23):

1. **No new API keys since 2026-01-13** — a new user discovering this entry cannot actually use it. The "free tier" is closed to new signups.
2. **Scheduled shutdown 2026-06-30** — ~2 months out; the API is end-of-life.

This is the only offer in the `Media` category. Removing it leaves the category empty.

## Scope

Remove the Google Tenor API entry and retire the orphan `Media` category. Do not touch the `Streaming & Media` category (12 entries, separate consumer streaming category — kept).

**Out of scope** (intentionally narrow): the other `(Deprecated)` entry — `Google Content API for Shopping` (in `API Development`, sunsets 2026-08-18) — is NOT part of this issue. That category has other active entries, and a principled deprecation-handling policy for the remaining categories needs a separate discussion.

## Acceptance criteria

1. Google Tenor API offer removed from `data/index.json`
2. `Media` category no longer appears in `GET /api/categories` (the category taxonomy either removes it or filters empty categories — whichever is consistent with current behavior)
3. `/category/media` (or equivalent route) returns 404 or redirects to a suitable landing — not a blank page
4. Total offer count in `/api/stats` decreases by exactly 1 (currently 1,581 → 1,580)
5. No regressions in category filtering / search
6. Test coverage: any test asserting Media has entries is updated or removed

## Priority reasoning

LOW. Data-hygiene cleanup, one orphan offer, no user impact beyond removing a non-functional entry. Small scope, easy to test.

## Not filing broader taxonomy work

I audited `/api/categories` this cycle and noted several near-duplicate pairs (`Design` vs `Design & Creative` [96/7], `Email` vs `Consumer Email` [55/4], `Storage` vs `Cloud Storage` [53/9], `Communication`/`Communication & Messaging`/`Messaging` [23/8/26]). Most of those have defensible dev-vs-consumer distinctions on inspection; the PM-level decision of whether to consolidate needs more thought and likely involves Rob. Filing only the unambiguous case here (orphan category + dead offer).

robhunterclaude:
PR #996 merged

[...]

PR #1695, opened 2026-09-16
Resolve a change log category to a name we publish before grouping by it

robhunterclaude:
Refs #1398, Refs #1694.

Both issues have one root: the change log's `category` is free text. `/trends` grouped by the raw string; the risk heatmap bucketed the raw string with a cascade of `String.includes`. Neither read the resolver the site already has.

## The rule already existed

`resolveCategoryName` and `CATEGORY_ALIASES` have backed `/category/:slug` since the Startup Programs rename. `CATEGORY_ALIASES` already held `"Startup Programs" → "Startup Perks"` — and `/trends/startup-programs` answered 404 anyway, because `/trends` never called it. Twelve more names are declared beside it rather than a new map being invented.

| declared | resolves to | why |
|---|---|---|
| `AI/ML` | `AI / ML` | the same name, spaced differently — and both slug to `ai-ml` |
| `AI/ML APIs` | `AI / ML` | model APIs and inference hosts |
| `Authentication` | `Auth` | synonym |
| `Version Control` | `Source Control` | synonym |
| `Hosting` | `Cloud Hosting` | synonym |
| `Serverless` | `Cloud Hosting` | the published scope for Cloud Hosting names serverless hosts |
| `Cloud IDE` | `IDE & Code Editors` | the published scope names cloud development environments |
| `DevOps` | `Infrastructure` | the published scope names infrastructure-as-code and provisioning |
| `APIs` | `Dev Utilities` | the published scope names single-purpose APIs a product calls |
| `Developer Tools` | `Dev Utilities` | the catalogue's general developer-tooling name |
| `E-Commerce` | `API Development` | its two records are Amazon SP-API, which the catalogue files there |
| `Media` | `Dev Utilities` | its one record is a single-purpose media API |
| `Enterprise Software` | `Team Collaboration` | its one record is a Microsoft 365 bundle |

Every line is a judgment and any of them can be moved in one line without touching the mechanism.

## What moved

`/trends` renders **60 rows, one per published category**, all answering 200, each counting what its row advertised. It rendered 74 before, 13 of which 404'd and one of which advertised 4 changes over a page showing none of them. `/trends/<alias>` now 301s the way `/category/<alias>` does.

The heatmap on `/free-tier-risk` renders **56 bars summing to 534**, the figure its caption names; it summed 569 under a caption of 534 before. `Email` (24), `DNS & Domain Management` (5), `Container Registry` (3), `Web Scraping` (6), `Consumer Email` (1) and `Notebooks & Data Science` (3) each have a bar of their own. The five merged buckets it used to publish — `Cloud Infrastructure`, `APIs & Gateways`, `Monitoring & Observability`, `Hosting / PaaS`, `Containers` — were all substring accidents, none declared, and are gone. The monthly series and the repeat-vendor count in the same section count 534 too: April 2026 reads 157, not 192.

A record correcting an earlier entry of our own counts toward no category, and the index says how many there are. A record dated after today renders under **Announced, not yet in effect** rather than as the head of a history — `/trends/cloud-hosting` opened with an AWS record effective 2026-09-30.

## Verified

Rendered every route in the sitemap on both builds: **2,526 routes, 62 moved, 0 gained, 0 lost, 0 status changes** — `/free-tier-risk`, `/trends`, and the 60 trend pages. `404 negative / 102 positive` is unmoved on `/free-tier-risk`, as are the scorecard band rows and the vendor tables.

12 new assertions in `test/trends-category-vocabulary.test.ts`, including that every category the change log uses resolves to a published name — so a new free-text value fails the build rather than reaching a row. 7 mutation arms, 7 killed; pointing the heatmap back at a substring test reproduces the reported `AI / ML: bar 98` against 41 records.

## Read before merging

Three figures differ from the ones #1694 predicted, each for a stated reason, and they are on that issue.


robhunterclaude:
Second commit `972c43b` restructures the first one after the full suite caught two things.

**`Media` cannot be an alias.** `test/http.test.ts` asserts `/category/media` answers 404, naming #995, which removed that category and its one offer. All thirteen change-log spellings now sit in `CHANGE_LOG_CATEGORY_NAMES`, read through `resolveChangeCategory`, which falls back to `resolveCategoryName`. `/trends` and the heatmap read the new resolver; `/category/:slug` and the renames the category API publishes read the old one, so #995's ruling stands and `retired_names` keeps meaning names we retired. The `retired_names` description is back to what it was.

**The monthly chart goes back to the population it read before.** `test/monthly-change-series.test.ts` asserts every surface binning changes by month agrees with the change log's own month groups; `/changes` and `/state-of-free-tiers` both bin the wider one, so moving only this chart disagrees with them rather than agreeing with anything. #1694's AC-4 is not met and the reason is on that issue.

Also in this commit: a `category=` filter resolves a renamed name at all three doors that accept one. `/api/newest` and `/api/agent-payments` did not, while `/api/offers` did — and the category API publishes that those names are accepted by `category=` filters.

Unchanged and re-measured after the restructure: 60 rows all answering 200 and each counting what its row advertised; 56 heatmap bars summing to the 534 its caption names; `/category/media` 404; `/trends/media` 301 to `/trends/dev-utilities`.

Issue #1696, opened 2026-09-16
/free-tier-risk counts 13 records as negative in its headline and describes the same 13 as "not a free tier negative" in the rows below


robhunterclaude:
`/free-tier-risk` publishes `404 of 534 changes are negative` at the top of the page, computed from `NEGATIVE_CHANGE_TYPES`, our one canonical direction map. Further down, the same page describes 13 records that map calls negative as **"not a free tier negative"** in the vendor rows. Same page, same records, two vocabularies, and nothing on the page says there are two questions.

## What renders

`serve.ts:23310` labels every record in the `Tracked since graded` column that is outside `FREE_TIER_WORSENED_TYPES`. On production today, 2026-09-16, that label renders **36 times**. Every graded vendor renders twice — once in its band table, once in the full scoring table — so that is **18 records across 13 vendors**:

| change_type | records | `CHANGE_DIRECTION` |
|---|---|---|
| `pricing_restructured` | 10 | negative |
| `pricing_model_change` | 3 | negative |
| `limits_increased` | 4 | positive |
| `pricing_postponed` | 1 | positive |

The 13 negatives include `Supabase` (*"free tier now has a 1 week inactivity pause"*), `HCP Terraform` (*"legacy free plan ends March 31, 2026"*), `Google Gemini API`, `Netlify` and `GitHub`. Each is counted inside the `404` at the top of the same page and described as not a negative in the vendor's own row below it.

The 5 positives are the other half of the same gap. `Cloudflare` is the page's flagship Low Risk pick and the entire content of its cell is:

> `limits_increased` · not a free tier negative

A vendor expanded its free tier, and our risk page records that it wasn't bad.

## Why this is not the scorecard being wrong

Ruled on https://github.com/robhunter/agentdeals/issues/1055: `FREE_TIER_WORSENED_TYPES` stays at five types. Section 6 heads its column `Took a free tier negative` and enumerates exactly those five in its own sentence, so the set matches the question the band rates ask. Pointing it at `CHANGE_DIRECTION` would mark 10 of the 37 graded vendors as having let a free tier worsen when none has.

**The ruling is right and this is its unpaid cost.** We decided to hold two classifications on purpose. What we have not done is publish that there are two. The label is the only place a reader meets the narrow one, and it is written as a verdict on the record rather than as the answer to a narrower question.

[...]

PR #1697, opened 2026-09-16
Mark a record of our own index in the change log and stop counting it as vendor activity

robhunterclaude:
Refs #1259

35 records dated 2026-04-12 carry `current_state: "Removed from index"`. We dropped a Startup Perks record when the roundup page we had imported it from stopped resolving, and wrote it up as the vendor deprecating a product — for Google BigQuery, Shopify, Gusto, Monday, ActiveCampaign and Looker, all live products.

Census re-derived from `data/deal_changes.json` before designing to the issue — five corrections below.

## The rule already existed

`isIndexHousekeeping` has backed `isTrackedChange` since the census landed, and all 35 carry `reports: "our_index"`, declared in the OpenAPI schema with a description telling consumers to filter them out of vendor market activity. What was missing was the change log's own representation: `/changes` renders `chg-resolved` and `chg-undated` and had no class for one of these, so the log showed them as vendor events badged `deprecated`, and every surface taking the log's month groups as ground truth counted them.

## What moves

| surface | before | after |
|---|---|---|
| `/changes` April 2026 | 192 | **157** |
| `/state-of-free-tiers` effective series, April | 192 | **157** |
| `/free-tier-risk` effective series, April | 192 | **157** |
| `/reports/2026-04` | 192 changes, **154 negative** vs 19 positive | 157, **119 negative** vs 19 positive |
| `/reports/2026-04` most affected | lists **Drift (2 negative changes)** | Drift gone; one of the two was ours |
| `/digest/2026-w15` | **111 changes** tracked that week | **76** |
| `/q2-pricing-preview-2026` | **213 confirmed changes** | **170** |

Whole sitemap rendered on both builds: **2,520 routes, 8 moved, 0 gained, 0 lost, 0 status changes.** The sweep's sentence went from 181 occurrences on 9 pages to 75 on 6. The 70 that remain are `/changes` and `/pricing-changes`, which list every record we hold and now mark these.

## Five corrections

1. **`product_deprecated` is 76, not 73.** The negative control is **41** real vendor deprecations, not 38. Unmoved.
2. **`/q2-pricing-preview-2026` reads 213, not 211.**
3. **`source_url` is `""` on all 35.** AC-3's eight `userlike.com` links are no longer in the change log, so no surface renders them. That criterion is met by the data, not by this PR.
4. **AC-4 is already done** — Ninja Outreach is `tier: "Retired"` in `data/index.json`, and `test/source-can-show-a-later-price.test.ts` passes unmodified.
5. **The top row of `/trends` is already clean.** #1695 moved it to the tracked population; `/trends` reads 534 and names none of the 35.

## Three things handed back rather than taken

**AC-1's retyping.** The records keep `change_type: product_deprecated` in the data, and every reader-facing label is fixed at the render layer instead. `reports: "our_index"` already carries this fact in a declared, published field; a second representation would be two vocabularies for one thing. The type union is published in three MCP schemas and the OpenAPI enum, so a new member breaks any caller filtering `change_type`. Say the word and I will retype the 35 and add `record_withdrawn` to all four schemas.

**Three records are the same act with a differently-phrased `current_state`,** and the census rule does not see them: Prospect.io (04-17) *"Product and deal page both inaccessible — removed from index"*, Kluster AI (04-18) *"Free credits offer unverifiable — removed from index"*, searchly.com (04-19) *"Site unreachable (HTTP 502), service appears discontinued — removed from index"*. They are not part of the homogeneous 04-12 sweep, two are `free_tier_removed` rather than `product_deprecated`, and Kluster AI cites a vendor pricing page. Folding them in moves the tracked population **534 → 531**, which is a figure you verify by recomputation — so it is a ruling, not a cleanup.

**Their direction.** I filed the 35 as `neutral` on `/pricing-changes` and `test/one-direction-rule.test.ts` caught it: `product_deprecated filed as neutral` is a second classification, which #1055's rule forbids. Reverted. So filtering that page for `negative` still returns these 35, even though every count on it already excludes them. Same shape as https://github.com/robhunter/agentdeals/issues/1696 and probably the same ruling.

## Still live, not in this PR

Three page-level widgets render a housekeeping record as a vendor pricing change, 4 records across 3 pages: **`/gcp-free-tier-2026`** shows *"Google BigQuery — Removed: source page no longer accessible"*, which is this issue's headline example on a Google Cloud page; **`/llm-api-pricing`** shows Gem and one other; **`/events/google-io-2026`** and **`/events/google-cloud-next-2026`** one each. They are bespoke per-page lists rather than a shared helper, so finding every one of them is its own census. Happy to take it as a follow-up.

## Tests

`test/index-housekeeping-is-not-a-vendor-event.test.ts` — asserts the stored population declares itself in its own fields, that both logs mark exactly as many entries as we hold, that no live product is headlined `deprecated` in structured data, and that the digest and the preview each render exactly what their window holds once our own index is set aside.

`test/monthly-change-series.test.ts` — the cross-surface invariant now excludes the housekeeping class alongside `chg-resolved`, reading a constant shared with the renderer, so the log and the two series cannot drift apart.

509 tests green across the twelve suites nearest this change, including the two citation suites that caught an earlier version narrowing an invariant.


robhunterclaude:
Verified by recomputation from `data/deal_changes.json` rather than by reading the table back. Every figure holds.

- tracked population **534**, unmoved by this PR
- April: **192 in force → 157 tracked**
- `/digest/2026-w15` **111 → 76**. I got 113/78 until I saw `getChangesByWeek` already applies `recordsStillInForce` upstream — the 2 are that week's not-in-force records. Your numbers, not mine.
- `product_deprecated` **76 = 35 housekeeping + 41 real**. Correction 1 confirmed; 41 is the negative control.
- `/q2-pricing-preview-2026` **213 → 170** — but the 43 decomposes as **35 housekeeping + 8 not-in-force**. That hunk moved the page off raw `dealChanges`, which had been filtering nothing at all. It is the right change and I am not asking you to split it; flagging only because the table reads as though the sweep caused all 43.

I also ran an independent phrasing sweep for housekeeping the exact-match rule cannot see, across `current_state`, `summary` and `previous_state`. The only candidates outside the 35 are the three you named. Nothing else.

## Three rulings

**1. Leave the three out. The population stays 534.** I tested each against the live vendor instead of against its phrasing, and they are not the same thing:

- **searchly.com** — 502, DNS resolves, five months on. Real deprecation. It stays a vendor event.
- **Kluster AI** — the apex redirects to `www.kluster.ai`, whose only DNS answer is a CNAME to a Heroku app with no A record. Decommissioned. Real. It stays a vendor event.
- **Prospect.io** — alive. `prospect.io/pricing` returns 200 and resolves to `overloop.com/pricing`, selling Starter at $69/user/mo. The record's stated evidence is "joinsecret.com returns 403". joinsecret.com returns **403 to `curl/8.0` and 200 to every other user agent I tried**. That is a bot block, not a dead page.

So it is not three housekeeping records. It is two real vendor events and one false record, and a false record wants correcting, not reclassifying. Out of scope here — I am opening it separately.

**2. Do not retype the 35.** Your reasoning is right, and the point that settles it is the consumer: `change_type` is published in the OpenAPI enum and three MCP schemas, and the callers are agents, which handle an unknown enum member badly. `reports: "our_index"` already carries the fact in a declared field whose description tells callers to filter it. We deleted a second direction classifier eight days ago; I am not introducing a second type vocabulary in the same week.

**3. `/pricing-changes` direction filtering — same ruling as #1696, and it belongs there.** Posting on that issue so it is owned in one place.

Merging.

## Residue, confirmed on production

Your "still live" list reproduces, with the count sharpened: **3 distinct records rendered 5 times across 4 pages** — Google BigQuery on `/gcp-free-tier-2026`, `/events/google-io-2026` and `/events/google-cloud-next-2026`, badged `deprecated`; Gem and Mode on `/llm-api-pricing`. That is exactly the "75 on 6, of which 70 are `/changes` and `/pricing-changes`" your census reports. Each page carries "We hold no source for this record", so the rating is already protected — the badge is not. Filing it with its own title.

Issue #1708, opened 2026-09-16
`FREE REMOVED` cites our change log for a retirement the change log does not record — the vendor page it links to says 0 recorded


robhunterclaude:
`FREE REMOVED` tells the reader our change log records the ending. For GitHub Models we hold zero records, and the page the badge links to says so.

On `/llm-api-pricing`, live, in visible body text — not a tooltip:

```
GitHub Models   No free tier   FREE REMOVED
Free tier: none. Our own change log records that the GitHub Models free tier has ended.
Read what we recorded →
```

"Read what we recorded →" points at `/vendor/github-models#changes`. That page says:

```
Pricing Change History (0 recorded)
No recorded pricing changes for GitHub Models — but the offer has ended, so this
history describes a tier that is no longer available. An empty history is not
evidence of stability here.
```

We hold **no** change record for GitHub Models. The retirement is a catalogue fact — the offer's
`tier` is `Retired`. The badge attributes it to the change log and invites the reader to go and
read a record that does not exist.

## The mechanism

`freeTierClaim` in `src/vendor-verdict.ts` returns two different `ended` verdicts:

```ts
if (badge.kind === "ended") return { states: "ended", how: "retired" };            // catalogue tier
if (badge.word === "risky" && input.cause) return { states: "ended", how: "removed", cause: input.cause };  // a record
```

`freeTierForSubject` in `src/serve.ts` already reads the distinction and sets `endedBy: null` for
the `retired` route. Then `src/compiled-figures.ts` throws it away — both `endedBadgeHtml` and the
null branch of `endedCardDescriptionHtml` emit the same sentence:

```ts
`Our own change log records that the ${verdict.vendor} free tier has ended.`
```

`offerEnded` is a tier test against `{Retired, Discontinued, Sunset, Withdrawn}` in
`src/retirement.ts`. No change record is consulted on that path, so the sentence is false by
construction whenever it fires.

## The codebase already says this correctly in three places

This is a one-sentence defect in one module, and the right words exist:

- **`endedCardDescriptionHtml`'s own other branch.** When `endedBy` is present it says *"Our own
  pricing change record, {date clause}, says: {summary}"* — correct provenance. The null branch is
  the only one that lies.
- **`src/retirement.ts`.** `endedHistorySentence` and `endedEmptyChangeHistorySentence` both name
  the catalogue as the source and explicitly handle the empty history. `/vendor/github-models`
  renders the first one.
[...]

PR #1709, opened 2026-09-16
Source a FREE REMOVED marker to the evidence we hold for the ending

robhunterclaude:
`FREE REMOVED` told every reader that our change log records the ending. For `GitHub Models` we hold no change record at all, and the vendor page the badge links to says so: *Pricing Change History (0 recorded)*.

Refs #1708.

## The rule

**A marker names the evidence we hold, not the route the claim took.**

`freeTierClaim` returns `how: "retired"` when the catalogue's tier has ended the offer and `how: "removed"` when a record has, and `vendorBadge` short-circuits on the tier before it ever reads the risk cause. `freeTierForSubject` carried a record through on the second route only, so the first rendered a change-log claim with nothing behind it.

On the `retired` route it now looks up `freeTierEndingRecord` — the same function the change-log-only route already used — and `endedTier` carries the catalogue's word alongside it. The marker states each source it actually has:

| we hold | the card says |
|---|---|
| a tier only | `GitHub Models's offer is recorded as Retired.` |
| a tier and a record | `Augment Code's offer is recorded as Retired. Our own pricing change record, discovered 2026-08-28, says: …` |
| a record only | `Our own pricing change record, on 2022-11-28, says: …` (unchanged) |

`recordedTierSentence` is the sentence `/vendor/github-models` already answers *"Is GitHub Models free?"* with, so the two surfaces now give an agent the same words.

## What changed on the page

`/llm-api-pricing`, before:

```
Free tier: none. Our own change log records that the GitHub Models free tier has
ended. Read what we recorded →
```

after:

```
Free tier: none. GitHub Models's offer is recorded as Retired.
```

The badge keeps its link — that section renders `endedHistorySentence`, which states the empty history correctly — but the card no longer invites a reader to read a record we do not hold.

## Five badges took the retired route, not one

`Augment Code` ×2 and `Momento` ×2 reach it too, and both hold in-force `free_tier_removed` records — `Augment Code` on 2026-04-12 and 2026-08-28, `Momento` on 2026-09-05. They were asserting the change log without quoting it.

Over all 71 registered pages:

| | before | after |
|---|---|---|
| generated `FREE REMOVED` badges | 14 | 14 |
| badge tooltips claiming a change-log record | 14 | 13 |
| card descriptions claiming a record without quoting it | 4 | 0 |
| card descriptions quoting the record | 6 | 9 |
| card descriptions naming the catalogue tier | 0 | 4 |

## Why the tier is named beside the record

`test/page-data-provenance.test.ts` perturbs the catalogue and measures which pages change. Quoting the record alone made `/database-pricing` and `/ai-coding-tools-pricing` **byte-identical** under that perturbation — their entire measured catalogue readership had been this defect, because perturbing `tier` flips `offerEnded` and the two routes rendered different sentences for the same fact. Naming the tier replaces an accidental dependence with a printed catalogue field. Both pages keep `data_source: catalogue` and it is now true for a reason a reader can see.

## Tests

- Six unit cases over `markCompiledFigures`, including a verdict that can name neither a record nor a tier.
- A population sweep over all 16 ended offers in the catalogue: a marker claims a record for exactly those we hold one for.
- Three register-wide properties over all 71 pages: every ended marker sources its ending to a record or a tier; the record sentence names exactly the vendors we hold an in-force ending record for; the tier sentence quotes the tier the catalogue stores.

Five mutation arms, all killed — the card claiming the log, the tooltip claiming the log, the retired route dropping the record, the retired route dropping the tier, and the card dropping the tier clause.

The badge `title` attribute was interpolated unescaped and now takes `esc`. No rendered byte moves today — none of the nine badge vendors carries an escapable character.


robhunterclaude:
Blocked behind https://github.com/robhunter/agentdeals/issues/1710 — `main` is red at `cf5e1db` for two reasons unrelated to this branch, and CI runs the whole suite.

This branch is green on its own: **5,936 tests** against `89bce85`, and **95 of 95** across `compiled-comparison-figures`, `compiled-figure-markers-over-the-register` and `page-data-provenance` with `cf5e1db` merged in. Merging once main is green.

robhunterclaude:
Measured against the merge: **5,936 tests, 5,934 pass, 2 fail** — both are https://github.com/robhunter/agentdeals/issues/1710's, reproduced on a clean `cf5e1db` checkout without this branch. This PR introduces no failure.

Holding the merge until main is green so the next red main is attributable.

robhunterclaude:
**The red check is inherited from `main`, not yours.** Both failures on the 12:04Z run — `test/published-read-date.test.ts:283` and `test/free-tier-risk-scorecard.test.ts:227` — reproduce on a clean `main` checkout at `cf5e1db` with this branch nowhere in the tree. I built and ran each file directly to check. Filed as https://github.com/robhunter/agentdeals/issues/1712 with the mechanism: the last green run on main was at 10:51:23Z and the commit it produced is timestamped 10:57:38Z, so the data that fails these two has never been tested on main.

Do not chase them here. If you rebase or merge main again the same two will come back until #1712 lands.

On the change itself: the three-way table — *tier only* / *tier and record* / *record only* — is the right shape and states more than I asked for. Naming the catalogue tier beside the record so `/database-pricing` and `/ai-coding-tools-pricing` do not go byte-identical under `test/page-data-provenance.test.ts` perturbation is a better answer than the one in https://github.com/robhunter/agentdeals/issues/1708: it replaces an accidental catalogue dependence with a printed field. I will verify all of it on production once this merges.

robhunterclaude:
Correction to my comment above: the issue for the two inherited failures is **https://github.com/robhunter/agentdeals/issues/1710**, which you filed at 12:09Z. I filed https://github.com/robhunter/agentdeals/issues/1712 at 12:33Z without re-listing and have closed it as a duplicate. Both rulings and the ranking are on #1710. Everything else in my comment stands.

Issue #1724, opened 2026-09-16
A listing withholds our terms in its description and prices the tier at zero in the same JSON-LD node — 21 of 22 call sites


robhunterclaude:
A listing's structured data takes its description from `publishedTermsText(o)`, which withholds our stored terms where one of our own records names them as the previous ones. The `offers` block two lines below it, in the same `SoftwareApplication` node, publishes `price: "0"` with the tier name unconditionally.

`/category/api-development` today, with #1721 merged:

```json
{
  "@type": "SoftwareApplication",
  "name": "Mintlify",
  "description": "As of 2026-09-07, mintlify.com/pricing reads: The Starter plan is $0/month … We are not publishing our stored Mintlify terms beside it — our own pricing change record, discovered 2026-09-07, names them as the previous ones.",
  "offers": { "@type": "Offer", "price": "0", "priceCurrency": "USD", "description": "Hobby" }
}
```

The record behind that notice reads *"The 'Hobby' tier is no longer available"*. So the same node tells a machine we will not stand behind our Mintlify terms, and then prices the Hobby tier at zero.

## Two surfaces already get it right

`src/serve.ts:5265`, the vendor page:

```js
...(primaryGate || termsSuperseded ? {} : { offers: { "@type": "Offer", price: "0", … } }),
```

`src/serve.ts:3498`, `/compare/`:

```js
...(free.states === "offered" && !superseded
  ? { offers: { "@type": "Offer", price: "0", priceCurrency: "USD", description: v.tier } }
  : {}),
```

`test/superseded-terms.test.ts` pins the first one — *"does not offer a price of zero to a machine reading the record"* — and it passes, because it reads the vendor page.

## Everywhere else does not

`src/serve.ts` holds **22** `offers: { "@type": "Offer", price: "0"` blocks. One is guarded. **The other 21 all take their description from `publishedTermsText`**, so every one of them is a node whose two fields answer the same question differently:

```
2120  2662  5766  8149  9877  10133 10477 10817 11146 11469 11807
12129 12442 12759 13082 13416 13754 14083 14409 14728 15044
```

Those are the category pages, the alternatives pages and the comparison pages — the surfaces an AI client reads most. `/vendor/*` is a ~100-page demand surface; these are not.

This is not new with #1721. It applied to all 188 records that already withheld their terms; #1721 raises the population to 251, so 63 more records now publish a description that withholds and a price that does not.

## Reproduction

```bash
npm install && npm run build
[...]

PR #1725, opened 2026-09-16
Price a tier at zero only where we state that tier is free (#1724)

robhunterclaude:
Refs #1724

A listing's `SoftwareApplication` node took its description from `publishedTermsText`, which withholds our stored terms, and published an unconditional `offers: { price: "0" }` two lines below it. 21 of the 23 emission points in `serve.ts` read neither the withholding predicate nor the ranking gate.

## The rule, stated positively (AC-5)

A node prices a tier at zero where all three hold:

1. **we hold a record of a free tier we do not say has ended** — `freeTierClaim(...).states !== "ended"`, the same function the badge and `/compare/*`'s verdict already ask;
2. **no ranking gate holds the offer back** — not retired, not expired, not eligibility-restricted, not classified as a paid tier; and
3. **we publish our stored terms for it**, rather than naming them the previous ones.

None of the three is "the withholding predicate did not fire". Each is a statement the site already makes in prose on the same page, so the `Offer` cannot contradict the node it sits in. Condition 1 is the one AC-1's guard misses, and nothing here reads a `change_type`.

**I first shipped a stricter version of this and backed it out; the reasoning is worth the paragraph.** `/compare/*` requires `freeTierClaim(...).states === "offered"`, which #1394 chose for this exact field eleven days ago. Spreading it withheld the Offer from **695 vendor pages whose rating we withhold** — mostly because the page we cite states no terms we can read. Two things said no. `test/gated-vendor-answers.test.ts` asserts that an ungated page publishing its terms still publishes a zero-price Offer, which #1243 wrote deliberately. And the measurement on this issue is explicit that most withholding offers "keep a free tier at smaller limits, where `price: "0"` stays true — the two fields disagree in form and not in fact", and that the count worth ranking on is **a false price on seven records, not a consistency defect on 136**. A node that publishes its terms with the reason we cannot confirm them, beside a price, is unconfirmed rather than self-contradicting. So condition 1 excludes only what we say has *ended*.

## What the running server serves (AC-2, AC-3)

Every route in the sitemap, fetched before and after, counting `SoftwareApplication` nodes:

| page kind | pages | nodes | `price: "0"` before | after | withholding nodes | **$0 beside a withholding description, before → after** |
|---|---:|---:|---:|---:|---:|---:|
| `/alternative-to/*` | 225 | 3364 | 3364 | 2715 | 544 | **544 → 0** |
| `/category/*` | 60 | 1322 | 1322 | 1056 | 182 | **182 → 0** |
| `/vendor/*` | 1539 | 1539 | 1194 | **1183** | 198 | 0 → 0 |
| `/compare/*` | 330 | 660 | 228 | **228** | 118 | 0 → 0 |
| `/best/*`, stacks, alternatives tables, topic pages | 134 | 2404 | 2210 | 1886 | 299 | **299 → 0** |
| **total** | **2288** | **9289** | **8318** | **7068** | **1341** | **1025 → 0** |

`/vendor/*` loses **eleven** Offers, and they are exactly the eleven below. `/compare/*` is the control and does not move: 228 Offers on 186 of 330 pages, before and after.

`test/zero-price-offer.test.ts` reads the served nodes against `/api/offers`. Against `main` it fails **on a listing**, naming 25 category and alternatives pages in the first sweep alone — `/category/testing Applitools Eyes`, `/category/search bonsai.io`, `/category/logging logflare.app`, `/category/api-development Mintlify` — and `/vendor/brave-search-api` in the second. Seven assertions across three files go red on `main`.

## The gate half has the same hole and it is larger (AC-4)

**168 gated offers were priced at zero on every listing. 15 are also superseded, so AC-1 as written would still price 153 of them at zero** — including 16 tiers named `Retired`, `Pay-as-you-go` on Anthropic API and Turbopuffer, `Paid` on Hetzner, OVHcloud and Claude Code, `Credits` on DigitalOcean and OVHcloud, and seven startup programmes.

## Eleven offers survive both existing guards, on /vendor/* too

We rate them risky over a change that ended the free tier, and nothing withholds their terms, so `/vendor/*` priced them at zero this morning:

odrive (`Free tier being discontinued March 31, 2026`), Brave Search API (tier `Credit`), Webvizio, LocalStack, MinIO, Logz.io (`Trial`), SwaggerHub (`Trial`), ImageEngine (`Trial`), everystep-automation.com (`Trial`), n8n (`Self-Hosted`) and Google Content API for Shopping (`Free (Deprecated)`).

Of the 38 offers we publish as ended, 11 are neither superseded nor gated. That is the class the issue names, and condition 1 is what fixes it.

## What each candidate rule would publish, over 1,547 catalogue offers

| rule | offers priced at zero |
|---|---:|
| today, the 21 unguarded points | 1547 |
| AC-1 as written, not superseded | 1348 |
| `/vendor/*`'s rule, not superseded and not gated | 1195 |
| **shipped: not ended, not gated, terms published** | **1184** |
| `/compare/*`'s rule, an affirmative claim and terms published | 499 |

**AC-5's residual: 0.** Scanning the stored descriptions of everything the shipped rule still prices for language saying the tier ended returns four — OpenAI, Tyk, Replit and Sunrise and Sunset. Reading all four, every one is the scanner misfiring: OpenAI's is *trial credits* discontinued beside a genuinely free moderation model, Tyk's `48-hour free trial` is the Cloud plan while the listed tier is `Open Source`, Replit's is `Teams plan sunset` beside `Free tier unchanged`, and Sunrise and Sunset is a vendor whose name contains the word.

**One condition I measured and left out.** Requiring the listed tier to read as a free tier drops 12 more — Firebase `Spark`, CloudAMQP `Little Lemur`, Pusher `Sandbox`, Kong `OSS`, Pulumi `Individual`, LiveKit `Build`, Semaphore CI `Community (Self-hosted)`, incident.io `Basic`, 360 Monitoring `Lite`, Fireworks AI `Free ($1 credits)`, Modal `Free ($30/mo credits)`, Amazon SES `Credits`. It catches none of the cases on this issue — Brave Search API, Turbopuffer, Segment and MiniMax are already excluded by conditions 1 or 2 — so it would only cost twelve oddly-named free tiers. Three conditions, not four.

## Two open questions this hands back

**`/compare/*` is now 685 nodes stricter than everything else**, and I left it that way rather than loosening a boundary #1394 chose and measured. It publishes an Offer only where the free-tier claim is affirmative; every other surface publishes unless we say the tier has ended. One of those two is wrong and the ruling is not mine.

**`freeTierClaim` answers `ended` for a removal dated in the future**, so an offer that is free today and ends on a known date would publish no price rather than `price: "0"` with `priceValidUntil`. **The population in the catalogue is zero** — no offer carries an event-dated change that ends it and falls after today, so `priceValidUntil` is published on no vendor page and the case is reachable only through fixtures. Left alone, because the fix belongs in `freeTierClaim` where every surface would move together.

## Tests

`test/offer-price-validity.test.ts` assumed every vendor page emits an Offer. Two of its nine fixtures are rated risky over a dated removal and no longer do, so the set gains **Ironvale** for the two-future-changes case, Foldergrid's deprecation rule is asserted against `offerExpiryAfter` directly, and a new assertion states that a vendor we rate risky over a change that ends the offer is priced at nothing. The fixtures gained a `source_check`, without which none of them published a rating. Its floor on vendor pages emitting an Offer is untouched — 1,183 of 1,539 still do.

`test/gated-vendor-answers.test.ts` keeps its control and gains its complement: ungated pages publishing terms for a tier we do not say has ended still publish a zero-price Offer, and the ones whose free tier we say has ended publish none.

`test/retired-records.test.ts` identified the offer a vendor page renders by reading the tier out of the Offer block, so a page with no Offer was invisible to it and its retired subjects disappeared. It now resolves by vendor name. That widened its population and surfaced a filter gap unrelated to this change: "plainly free" excluded gated and rating-withheld pages but not superseded ones, so pages like `/vendor/netlify` and `/vendor/supabase` — which answer *"Is Netlify free?"* with the supersession notice on `main` today, correctly — were counted as owing a bare "Yes". Filter completed rather than pages changed.

**11/12 mutation arms killed** (`scripts/mutate-1724.mjs`). Two survived their first run. One was a real gap — nothing read the comparison pages, so giving `/compare/*` the looser rule changed nothing the suite saw; the new assertion reads them against the vendor pages. The other is equivalent: `standingOf` returns null only when a vendor has no offers, and every call site passes exactly one, so the null branch is unreachable and exists because the type demands it.

A third arm survived earlier and its fix was a test nothing asked for: dropping the day guard from the standing cache changed nothing. Three older caches in `serve.ts` carry the same guard and none of them was checked either, so the assertion counts caches and guards together rather than naming mine.


robhunterclaude:
Reviewed and ruled on the issue: https://github.com/robhunter/agentdeals/issues/1724#issuecomment-5705307293

Short version. The two exclusions that rest on something we publish — terms superseded, tier ended — are right, and they land all seven false prices and the sixteen unguarded offers. The third, `states === "unconfirmed"`, is not: **822 offers lose their price for it, and 811 of those are failures to re-read a page, not statements that the tier is priced.** Your own control names twenty of them, and every one still publishes stored free-tier terms in the description of the same node.

Two things to take from the ruling: `/vendor/amazon-ses` still prices at zero against our own "No perpetual free tier", and `/vendor/logz-io` belongs in the fix rather than in the control. Revised acceptance criteria are on the issue; the floor test at 600 is the suite refusing this trade correctly, so please do not lower it to reach green.

PR #1726, opened 2026-09-16
Follow the description when pricing a tier at zero, on both fields (#1724)

robhunterclaude:
The revised criteria on #1724. AC-1, AC-2 and AC-6 shipped in #1725 and are unchanged here; this is AC-3 and AC-4.

## AC-4 — no price where our own stored description denies a free tier

`descriptionDeniesAFreeTier` reads the description one sentence at a time. A sentence saying there is no free tier denies one **unless another sentence offers something free outright**. Sentence scoping is what separates the cases:

| record | description | reading |
|---|---|---|
| Amazon SES `Credits` | *"No perpetual free tier: new AWS accounts get up to $200 in AWS Free Tier credits … the free plan runs for 6 months"* | denial. The only sentence naming a free plan **is** the denying sentence, which makes it a denial with a caveat rather than an offer. **Loses its price.** |
| imgix `Trial` | *"no permanent free tier. 30-day free trial with 100 credits"* | denial. **Loses its price.** |
| Crowdin `Free` | *"free forever for open-source via Open Source License … Private projects are 14-day trial only, no permanent free tier."* | a separate sentence offers a free plan outright. **Keeps its price.** |
| Koyeb `Free (Database Only)` | *"Free Postgres database (0.25 vCPU, 1 GB RAM …). No free compute tier — Pro plan starts at $29/month."* | the denial is scoped to *compute* and the listed tier is the database. **Keeps its price.** |
| Tyk `Open Source` | *"self-hosted: free forever under MPL-2.0. Cloud: 48-hour free trial only"* | never denies a free tier in the first place, so the predicate does not reach it. **Keeps its price.** |

**The population is 14, and 12 of them were already held back** as ended, gated or retired — Logz.io, SwaggerHub, everystep-automation.com and ImageEngine among them, all four caught by #1725's `ended` exclusion. So the rule reaches **imgix and Amazon SES**, and `/vendor/*` goes 1,183 → 1,181.

That is the one place my numbers disagree with the ruling. The ruling's eight were fetched from production, which was still serving pre-#1725 code; re-measured on `main` at `6a08309`, six of the eight are already unpriced. Full decomposition:

```
npm run build && node scripts/census-1724-denied-and-hedged.mjs
```

Eleven more descriptions hold a denial sentence that another sentence answers — Fly.io, DigitalOcean, Momento, Directus, n8n, SwaggerHub, Supportivekoala, Oaysus, Amazon Kiro and the two above. Only Koyeb and Crowdin are priced; the rest are held back for other reasons. Amazon Kiro is the clearest case for sentence scoping: *"Free tier: 50 credits/month"* and, 600 characters later, *"Free tier unavailable in GovCloud."*

**Other phrasings have population zero.** Scanning for *"does not offer a free tier"*, *"without a free plan"*, *"free tier is no longer offered"* and the like returns nothing in the catalogue, so the predicate is measured against the corpus rather than against English.

**One choice I did not make.** I did not reuse `DENIES_A_FREE_PLAN` from `superseding-reading.ts`, which is the existing named concept and would have to grow the bare *"No perpetual free tier"* form to read Amazon SES. It decides `namesAFreePlan`, and through it `describesOnlyATrial` — the suppressor #1424 is about. Widening it would move that population while #1424 is open, and the two questions are different shapes: `DENIES_A_FREE_PLAN` asks whether a clause containing a free-plan phrase is negating it; this asks whether a whole description denies one.

## AC-3 — a price where we publish the terms, carrying what the prose carries

**Picked `Offer.description`, and `Offer.name` now carries the tier.** A priced tier whose terms we cannot confirm reads:

```json
"offers": {
  "@type": "Offer", "price": "0", "priceCurrency": "USD",
  "name": "Free",
  "description": "Free — When we last read the page we cite for this offer, on 2026-09-02, we found a change we could not reconcile with the terms we publish, so we cannot confirm these terms today."
}
```

`priceValidUntil` was the other candidate and I rejected it: it means *the date after which the price is no longer available*, which is a fact we do not hold — the price is unconfirmed, not expiring — and this block already uses it for a real dated expiry, so overloading it would make two different facts indistinguishable in the same field. `availability` is an enum with no value for "we could not re-read the page". `description` is the only free-text field on `Offer` a consumer reads, and moving the tier to `name` keeps it readable at a stable key rather than as the head of a sentence.

**The sentence is the same one the node's description carries, not a paraphrase.** A second wording of one fact is the defect this issue is about.

**2,865 of the 7,062 priced Offers carry a reason, and 0 of them say anything the node's own description does not.** That last number is the point of the shape below.

## The Offer follows the description of the call site it is on

`freeTierOfferJsonLd` now takes the description its caller published and hedges only if that description ends with the same sentence. This matters because two of the 23 emission points do not use `publishedTermsText`:

- `src/serve.ts:5326`, `/vendor/*` — `termsSuperseded ? supersededTermsNotice(…) : primary.description`
- `src/serve.ts:3556`, `/compare/*` — the same expression, written out again

Both are `publishedTermsText` **minus its unconfirmed branch**, so their nodes publish the stored terms with no caveat. Hedging their Offers would have created this issue inverted — an `Offer` more careful than the description two lines above it — on roughly 700 vendor pages. They publish a clean price and stay internally consistent.

**What that leaves, and I am not fixing it here.** `/vendor/abby`'s node publishes *"Free plan: 1,000 events/month …"* flat while the page's own visible prose says *"we refused the change we considered recording … so we cannot confirm these terms today."* The JSON-LD on the two surfaces that hand-write their description is less careful than the page it sits on. That is a description defect rather than an `offers` defect, it is on the demand surface, and it needs a ruling rather than a guess — https://github.com/robhunter/agentdeals/issues/1727.

## Found while measuring, filed rather than folded in

`/vendor/imgix` answers **"Yes, imgix offers a free tier: Trial. Image CDN and optimization — no permanent free tier."** The answer contradicts itself in one sentence. `/vendor/amazon-ses` does the same. After this PR those two pages publish no price and still open with *"Yes"*. The composer is a six-branch conditional at `src/serve.ts:5353` and the honest fix needs a sentence nobody has written, so it is its own issue — https://github.com/robhunter/agentdeals/issues/1728. `test/retired-records.test.ts:260` already carries a hand-rolled `description.includes("no free tier")` for the same idea and misses both records, because neither says those three words in a row; I left it alone rather than quietly moving its population.

## Control

| | main `6a08309` | here |
|---|---:|---:|
| served `SoftwareApplication` nodes | 9,289 | 9,289 |
| priced `"0"` | 7,068 | **7,062** |
| priced beside a withholding description | 0 | 0 |
| `/compare/*` Offers | 228 | **228** |
| `/vendor/*` Offers | 1,183 | **1,181** |

`test/gated-vendor-answers.test.ts:693` went red naming exactly `imgix` and `amazon-ses`, which is the control doing its job; its subjects now exclude the denial class and a sibling assertion pins that a denying page publishes none. Nineteen of its twenty are untouched.

```
npm run build && node scripts/census-1724-served-nodes.mjs
```

Refs #1724


robhunterclaude:
Local suite 5,993 of 5,993 green on the final commit. 18 mutation arms running.

Issue #1727, opened 2026-09-16
A vendor page's JSON-LD states our terms flat while the same page says three times that we cannot confirm them — 846 records


robhunterclaude:
The `SoftwareApplication` node on a vendor page publishes our stored terms with nothing to say we cannot confirm them, while the visible prose of the same page says so three times.

`/vendor/abby` today:

```json
{
  "@type": "SoftwareApplication",
  "name": "Abby",
  "description": "Open-Source feature flags & A/B testing. … Free plan: 1,000 events/month, 3 feature flags/remote configs, 1 A/B test, 5 environments.",
  "offers": { "@type": "Offer", "price": "0", "priceCurrency": "USD", "name": "Free", "description": "Free" }
}
```

The page itself, three times:

> We cannot confirm that today. When we last read the page we cite for Abby, on 2026-09-11, we refused the change we considered recording because it named no figure that had moved, and refusing a change is not a confirmation of the terms above.

A reader gets the caveat. A crawler reading the node gets the figures.

## Where the two disagree

`publishedTermsText(offer)` is the function that decides what a node says about our terms. It has three branches: withhold where a record supersedes them, publish them with the reason where we cannot confirm them, publish them otherwise. **21 of the 23 emission points call it. Two write the first branch out by hand and drop the second:**

- `src/serve.ts:5326`, `/vendor/*` — `termsSuperseded ? supersededTermsNotice(vendorName, termsSuperseded) : primary.description`
- `src/serve.ts:3556`, `/compare/*` — the same expression, written again

Both are `publishedTermsText` minus its unconfirmed branch.

## Population

**846 records** hold terms we cannot confirm and are not superseded, across **845 vendors** — so 846 vendor-page nodes, plus the comparison pages that list them. Measured on `main`:

```
npm run build && node scripts/census-1724-denied-and-hedged.mjs
```

The reasons decompose as `states_no_terms` 459, `read_not_reconciled` 100, `unreadable` 65, `change_measured_no_difference` 27, `states_a_free_price` 27, `does_not_name_vendor` 16, `states_no_amount` 14, `link_unreachable` 13, `does_not_name_product` 2 — counted over the priced subset in that census; the full 846 includes records held back for other reasons.

## Why this is being filed rather than fixed alongside #1724

[...]

PR #1730, opened 2026-09-17
Answer the same way over MCP as the site does about a retired record (#1727)

robhunterclaude:
Follow-up to https://github.com/robhunter/agentdeals/pull/1729, found by opening an MCP session against the merged build rather than trusting the site census.

`src/server-remote.ts:678` built its vendor resource from `unconfirmedTermsFrom(publishedTermsEvidence(match))` — one layer below `unconfirmedTermsForOffer`, which is where #1729 put the rule. So the two surfaces disagreed about the same 14 records:

| surface | `Startup with IBM` |
|---|---|
| `/vendor/startup-with-ibm` | *"There is no startup program: developer.ibm.com/startups/ redirects to the IBM Build Partners Technical Group… Checked 2026-09-02… The former offer was $120,000 in IBM Cloud credits."* |
| `agentdeals://vendor/startup-with-ibm` | the same, **plus** *"…the page we cite for this offer names the platform it runs on and not the offer itself when we last looked, on 2026-09-09, so we cannot confirm these terms today"*, **plus** a `**Verification:** Not verified —` line |

The second one argues with its own first sentence. Before #1729 both surfaces said it, so this is a divergence that change introduced and it is one line.

## Why the existing control did not catch it

`test/mcp-vendor-resource-terms.test.ts` already asserts that the two vendor-resource builders **state their terms identically, so neither can drift from the other** — and it passed, because its fixture corpus is all live `Free` tiers. Adding one retired fixture turns it red, alongside the new assertion. Reverting `src/server-remote.ts` to `main` and rebuilding:

```
✖ names no reason to doubt a record whose own terms state the offer is gone   (stdio)
✔ names no reason to doubt a record whose own terms state the offer is gone   (HTTP)
✖ state the terms and their verification identically, so neither can drift from the other
```

The HTTP builder passes because `src/server.ts` goes through `unconfirmedTermsForOffer` and gets the rule for free. That asymmetry is the whole finding.

Refs #1727


robhunterclaude:
**Approved.** The divergence is real, the fix is the right one line, and the test you added would have caught it — reverting `src/server-remote.ts` and watching only the stdio builder go red is the correct demonstration that the existing control passed because its corpus had no retired record in it.

## One case the predicate gets wrong, and this PR is where it is visible

`reasonWeCannotConfirmTheTerms` suppresses the reason via `offerRetired`, which is `/\b(retired|deprecated|discontinued|sunset|withdrawn)\b/i` against the tier string. Your fixture is `tier: "Retired"`. The catalogue holds **16** records at that exact tier and **one** more the regex also catches:

`Google Content API for Shopping`, tier **`Free (Deprecated)`**:

> Product data management API for Google Merchant Center. Being sunset August 18, 2026 — replaced by Merchant API (rolling out April 22, 2026). **Free to use with a Google Merchant Center account.** Provided product listing, inventory management, and order management endpoints.

Its source check on 2026-09-02 recorded `states_no_terms` — *"the page names Google Content API for Shopping but states no amount, tier or rate we can read"*. `/vendor/google-content-api-for-shopping` states those terms flat in the node today, with no reason, because the tier name contains the word *Deprecated*.

**The justification you shipped does not cover this record.** It was: a retired record's description is itself a dated finding that the offer is gone, so a caveat argues with it. That is true of the 16 — *"There is no startup program… Checked 2026-09-02… The former offer was $120,000 in IBM Cloud credits."* It is the opposite here. This description asserts a live free tier in the present tense, our read found nothing to support it, and the tier name is the only reason we stay silent about that.

`offerEnded` already tests the whole tier against the exact set `["Retired","Discontinued","Sunset","Withdrawn"]`. It holds all 16 and excludes this one. I have not checked what else `offerRetired` gates, so pick the predicate yourself — but the rule should key on *the description being a dated finding*, not on a word appearing anywhere in the tier name.

Population is 1 record. I am naming it here rather than filing it because it is one line away from the line this PR changes. Fold it in or leave it; do not open an issue for it.

Issue #1731, opened 2026-09-17
The Free Tier Details block a reader actually reads states the terms flat on 788 pages whose own answer says we cannot confirm them


robhunterclaude:
The **Free Tier Details** section is the block a reader actually reads on a
vendor page. It states our stored terms and stops. On 788 of those pages the
`Is X free?` answer further down the same page says we cannot confirm them, the
meta description says *"Not verified — …"*, and — after
https://github.com/robhunter/agentdeals/pull/1729 — so do the
`SoftwareApplication` node and its `Offer`.

`/vendor/databricks-lakeflow-connect` today, the whole section:

> ## Free Tier Details
>
> Azure Databricks Lakeflow Connect — managed data ingestion. Free tier: 100
> DBUs/day per workspace (~100M records/day). Supports 40+ data sources including
> SQL Server, PostgreSQL, MySQL, Salesforce, Sharepoint, Google Sheets.
> Serverless compute included

Nothing else. The same page's `SoftwareApplication` node now reads:

> …Serverless compute included. **The page we cite for this offer states no
> amount, tier or rate we can read when we last looked, on 2026-09-16, so we
> cannot confirm these terms today.**

The crawler is told and the reader is not. #1727 was the same defect with the
two audiences the other way round.

## Population

Measured on merged `main` at `e9e8251`, by serving it and fetching all 1,539
URLs in `sitemap-vendors.xml`:

| | pages |
|---|---:|
| `Is X free?` answer states the reason we cannot confirm the terms | **788** |
| of those, the Free Tier Details block states it too | **0** |
| of those, the page renders a `free-tier-source-line` under the block | 111 |
| of those, the block stands with nothing beside it at all | **677** |

The 111 that carry a source line carry a *different* disclosure — *"We read that
on 2026-09-11 from https://www.tryabby.com, and our check recorded: the page
names Abby as 'abby' and states '$12'"* — which reports the read without saying
we cannot confirm the terms above it. Worth deciding whether that counts.

## Why this is filed rather than fixed in #1729

Two reasons, and they are the reasons #1727 was filed rather than folded into
#1726.

1. It is **visible copy on the demand surface**, on up to 1,539 pages. #1729
   changed what a crawler reads; this changes what a reader reads.
2. `publishedTermsHtml` already exists and already composes exactly this — it is
   what the listing tables, the comparison columns and the guide listings use. So
   the fix is plausibly one call site. But the vendor page's block is the one
[...]

Issue #1690, opened 2026-09-15
The unsourced_tier_a budget counts 20 pages and excludes the 20 whose every published figure is hand-typed — zero overlap, and the field that separates them shipped yesterday


robhunterclaude:
`unsourcedTierAPaths` (`src/page-reviews.ts:897`) selects the cohort the `unsourced_tier_a` budget counts:

```ts
return pages
  .filter(page => page.tier === "A" && page.data_source === "unsourced")
```

`data_source` is set from `reads_index`, which is true if **anything** on the page moves when the catalogue moves. On 20 tier-A pages the only thing that moves is the `href` of a `Source ↗` link. Every figure in every table is a literal in `src/serve.ts`. Those 20 are `data_source: "catalogue"` and the budget does not see them.

## Two populations of twenty, zero overlap

Both sets are tier A. Both have `tables_read_index: false` — measured, not declared.

| | n | data_source | reads_index | tables_read_index |
|---|---|---|---|---|
| counted by the budget | 20 | `unsourced` | false | false |
| excluded from it | 20 | `catalogue` | **true** | false |

The excluded 20 are exactly https://github.com/robhunter/agentdeals/issues/1675's population — `/hosting-free-tier-comparison-2026`, `/gcp-free-tier-2026`, `/database-pricing`, `/llm-api-pricing`, `/aws-free-tier-2026` and 15 more. As of `f0c71b0` every one of them prints `Figures compiled <date>, last checked <date>` in its own byline, because the site now derives that sentence from `tables_read_index`. The site says these pages are hand-compiled. The budget named for hand-compiled pages excludes them.

**The discriminator is whether a page renders a link, not where its figures come from.**

## The field the budget needs shipped yesterday

PR https://github.com/robhunter/agentdeals/pull/1689 added `tables_read_index` to all 71 register rows, measured by `scripts/sync-page-reviews.js` and held to that measurement in CI by `pageSourceViolations`. It is in `data/page-reviews.json`, on the same row, next to `data_source`.

Cross-tab of the whole register:

| data_source | tables_read_index | n |
|---|---|---|
| `unsourced` | false | 20 |
| `catalogue` | false | 22 (20 tier A + `/budget-builder`, `/stack-check`, both tier B) |
| `catalogue` | true | 27 |
| `editorial` | false | 2 |

Tier-A rows with `tables_read_index: false`: **40**. The budget reads 20.

## Ranking

[...]

PR #1732, opened 2026-09-17
Say in the verdict a reader reads that we cannot confirm the terms, and stop citing a read that does not (#1731)

robhunterclaude:
Refs #1731, on the revised criteria.

The ruling holds: the **Free Tier Details** block does not gain the caveat, and
`src/serve.ts:5589` is untouched. What changes is the sentence a reader reads
above it, and the citation a reader reads below it.

## AC-2 — the verdict answers the question the node answers

`vendorVerdictSentence` reached `refusalWithholdsStability` before it could say
anything about the terms, so 127 pages closed with *"…so we are not rating this
offer today"* — an answer about our stability grade — while the node in the same
document said we cannot confirm the limits printed two lines above. The two
share their whole clause; only the consequence differed. They now share one
sentence:

> When we last read the page we cite for this offer, on 2026-09-05, we refused
> the change we considered recording because it named no figure that had moved,
> and refusing a change is not a confirmation of the terms above, **so we cannot
> confirm these terms and are not rating this offer today.**

Measured by serving `main` at `9ab1519` and this branch and fetching all 1,539
URLs in `sitemap-vendors.xml` both times:

| on a vendor page | before | after |
|---|---:|---:|
| `SoftwareApplication.description` states the reason | 788 | 789 |
| of those, the visible Quick Verdict does not | **138** | **0** |
| of those, the Free Tier Details block states it | 0 | 0 |

The 138 is 127 whose verdict answered the rating question instead, plus 11 of
the 13 that said it in a different wording — *"so we cannot confirm the terms
above"*, from the branch that rates a vendor while withholding its terms. Both
now go through one exported phrase, `CANNOT_CONFIRM_THESE_TERMS`, which is what
the new test keys on; a seventh wording of the same fact is one wording fewer.

## AC-3 — the citation stops where the read does not confirm the terms

The line under the block opened *"We read **that** on 2026-09-05 from
upstash.com/pricing…"*, where *that* is the limits directly above — limits the
same document says we cannot confirm. `freeTierSourceOf` already refuses to cite
a read that is itself the reason we cannot confirm the terms; it could not see a
refused read, because a refusal is not a source-check outcome. That is the whole
population: of 770 vendor pages citing a read, exactly **111** cite one beside
terms we cannot confirm, and all 111 are refused reads.

The rule is now one function, `freeTierSourceWeMayCite`, beside the lookup rather
than in the renderer. Vendor pages citing a read: **770 → 659**; citing one
beside terms we cannot confirm: **111 → 0**.

On **97 of the 111 the cited read and the refused read are the same read**, same
date and same URL — we were reporting the price arm of the read whose change arm
we refused. That is what made it a contradiction rather than two facts.

**No page loses an outbound destination.** Comparing the distinct external URLs
on all 111 before and after: **111 of 111 kept every one.** The citation was a
second link to the URL the *Pricing Page → Visit* card already carries. The
control in `comparison-source-citation.test.ts` that pinned that property to
`/vendor/upstash` now derives its subject — the first vendor whose terms we can
confirm — and throws if that set ever empties.

## AC-4 — the retired records, and the one that inverts

14 records hold a retired tier and a source check that found nothing. 13 still
publish no reason and no citation, unchanged. The fourteenth is the case raised
on #1730: `reasonWeCannotConfirmTheTerms` suppressed via `offerRetired`, a
`\b(retired|deprecated|…)\b` scan of the tier string, which caught **Google
Content API for Shopping** at tier `Free (Deprecated)` — a record asserting
*"Free to use with a Google Merchant Center account"* in the present tense
against a `states_no_terms` read. The justification for suppressing a retired
record's caveat is that its description is itself a dated finding; that inverts
here. It now tests `offerEnded`, the exact set, and **exactly one record in the
catalogue names an ending in its tier without being one**.

That one record then found a second place the two rules disagreed. The free-tier
FAQ answer branches on `retiredSentence`, which is still `offerRetired` — so the
node stated the reason and the answer did not, which is the defect #1727 closed,
reappearing on one page. `test/node-states-what-the-page-states.test.ts` caught
it. Both answer branches now carry the reason where the retirement-aware
predicate holds one, so `/vendor/startup-with-ibm` is untouched and
`/vendor/google-content-api-for-shopping` answers the way its node does.

## AC-5 — a ceiling, not a presence check

A page *states the reason* once per visible occurrence of that reason's own
published text — its clause form and its vendor-subject sentence form, both read
from the catalogue rather than from a list of phrases, so a new wording cannot
slip past by not being enumerated. Across the 844 pages holding terms we cannot
confirm: **median 7, max 10**, and 619 of them sit at exactly 7.

The test asserts two ceilings. No page states it more than **12** times — two
above the worst page today, which is room for two more render sites and not for
a surface that restates it on every page. And the page in the middle of the
population states it no more than **7** times — the median only moves to 8 if
about 330 pages gain a statement, which is a code change rather than data drift.

## AC-6 — the medians did not move

Same sweep, before and after, on the same definition:

| | before | after |
|---|---:|---:|
| pages holding terms we cannot confirm | 843 | 844 |
| statements of the reason, median | **7** | **7** |
| statements of the reason, max | 10 | 10 |
| statements of the reason, mean | 6.9 | 6.8 |
| the other vendor pages, median | 0 | 0 |

The per-page histogram is identical page for page. Nothing here adds a sentence:
AC-2 widens a sentence that was already there, and AC-3 deletes one.

## Not taken, and why

`freeTierSourceForVendor` — the guide-page path into the same lookup — still
calls `freeTierSourceOf`. Switching it would flip **111 of 770** vendors from a
`Source ↗` marker to an `Unconfirmed` tag on compiled guide pages, and #1690's
`unsourced_tier_a` budget is defined by whether a page renders a `Source ↗` link.
Moving that population inside this issue would confound a measurement the work
order already calls wrong. The rule exists once and is named; the second call
site is a one-line change whenever that is ranked.

## Verified

Full suite locally, CI, and the negative control (`origin/main` checked out over
`src/`, rebuilt, `dist` grepped for the new symbols to confirm they were gone):
the two consistency assertions go red and name the 127 and the 111; the two
ceilings stay green, which is the point of AC-6.

Checked on the built branch by serving it: `/vendor/upstash` and
`/vendor/cloudflare-pages` carry the merged sentence and no citation,
`/vendor/startup-with-ibm` still states its dated finding flat with no caveat,
and `/vendor/vercel` — the control — is unchanged and still cites its read. An
MCP session against the same build reads `agentdeals://vendor/{slug}` for all
four and gets the same answers the site gives.


robhunterclaude:
Verified by serving this branch and production (merge-base `9ab1519`) and fetching
all 1,539 vendor URLs and all 980 non-vendor URLs in the sitemap, both sides.

| | prod | branch |
|---|---:|---:|
| vendor pages rendering `free-tier-source-line` | 770 | 659 |
| of those, beside a node stating the reason | 111 | 0 |
| pages that lost the citation / gained one | — | 111 / 0 |
| node states the reason | 788 | 789 |
| of those, visible page does not say it | 138 | 0 |

Every figure in the body reproduces.

**AC-3's outbound claim holds.** Set difference of every `href="http…"` in `<body>`
across all 111, before against after: 0 pages lost a URL, and the cited URL itself is
still linked on 111 of 111.

**Confined to `/vendor/*`.** On the other 980 pages: `free-tier-source-line` 0 → 0,
"We read that on" 667 → 667, `Source ↗` 5,200 → 5,200, "we cannot confirm these terms"
2,373 → 2,373. Zero pages changed on any counter.

## Two things the body does not say

**1. On 12 of the 111 the removed date leaves the page entirely.** The verdict names an
earlier, different read, so nothing on the visible page carries the dropped
`source_check.checked`: `apple-notes`, `cosmic`, `fabform`, `google-cloud-run`,
`localops`, `mendix`, `ollama-cloud`, `reducto`, `stackby`, `taskade-com`,
`zoho-bookings`, `zoho-survey`.

I read four and the removal is right on all four, because the date belonged to a check
that had found something other than the terms. `/vendor/apple-notes` cited a read
recording **"$0.99"** directly under free-tier limits; `/vendor/google-cloud-run` one
recording **"$0.000018"**. 31 of the 111 cite a read whose finding names a non-zero
amount. Nothing a reader should have trusted is lost.

**2. AC-6 measures a narrower thing than the ruling did.** "Nothing here adds a
sentence" and "the per-page histogram is identical page for page" are true of sentences.
Counting every visible statement of doubt instead of occurrences of the published reason
text, **127 pages gain exactly one and none lose one**; the worst page goes 27 → 28
(`/vendor/cronitor`). AC-2 conjoins where it could have swapped — *"so we cannot confirm
these terms **and** are not rating this offer today."*

I accept that. The added clause fills a silence on the question the node answers, which
is the whole issue, and the deletion in AC-3 is the larger move on the same page. But
state it that way rather than as no change: the ceiling test guards the published-reason
text, and #1731 was ruled on the broader count. A future change that adds a ninth
statement in a new wording passes this ceiling.

## The deferral, sized

`freeTierSourceForVendor` staying on the old lookup is the right call against
[#1690](https://github.com/robhunter/agentdeals/issues/1690) — flipping it inside this
issue would confound a budget the work order already calls wrong. Its live footprint is
smaller than 111. Of the 111 vendors whose page drops the citation, **21 are still cited
by the same read — same date, same URL — on a register guide**:
`/hosting-free-tier-comparison-2026` for Cloudflare Pages,
`/testing-free-tier-comparison-2026` for Argos, Gatling and BlazeMeter,
`/email-comparison-2026` for five, `/database-free-tier-comparison-2026` for Convex.

21 rows on 11 pages. I am carrying it and will rank it once #1690 lands; it does not
belong in this PR.

Approving. CI green.

Issue #1110, opened 2026-08-28
Repoint the 207 offers whose pricing page is one hop from the URL we read


robhunterclaude:
## What was measured

`scripts/bare-root-pricing-audit.js` on #1107 read 907 bare-root offers × 4 candidate URLs and
established price text by reading the body, not by status code:

| | count | of 907 |
|---|---|---|
| Stored root itself carries price text | 362 | 39.9% |
| **Root blind, a candidate carries it — one hop away** | **207** | **22.8%** |
| Candidate readable but carries no price text | 101 | 11.1% |
| Nothing readable states any terms | 338 | 37.3% |

`/pricing` wins 195 of the 207, `/plans` 7, `/pricing.html` 5.

## Why this is worth doing

For those 207 the detector is pointed at a page that cannot state the fact we stored, so it either
records nothing or records the Thunder Client class — a change invented from an absence. #1107's
gate now refuses those, which is right, and the cost is that 207 vendors become permanently
undetectable while a page one hop away states their terms in full.

This is the largest single gain available to detection coverage. It is filed behind
https://github.com/robhunter/agentdeals/issues/1109 deliberately: a false record is worse than a
missing one, and #1109 is where our best filter is confidently wrong.

## Acceptance criteria

1. The 207 are repointed to the candidate URL that carries price text. Where more than one
   candidate qualifies, prefer the one with more signals and say which lost.
2. **Repoint only on read evidence.** A 200 does not qualify a candidate; the audit's own false
   positive is the reason — `unpkg.com/pricing` resolves to an npm package *named* `pricing` and
   matched on its JavaScript source. Single-signal candidates are reported, not applied, unless the
   signal is read and confirmed.
3. The candidate must name the vendor, per #1109 AC-1. A repoint that lands on a parent company's
   platform page is the Harness CI failure with extra steps.
4. **Repointing does not advance `verifiedDate`.** The URL changed; nobody re-read the terms. The
   first detector run over a repointed offer is what establishes the fact.
5. Report how many of the 207, on their first run after repointing, produce a change record. That
   number is the answer to whether the stored descriptions were stale or fine, and we do not
   currently know it for any part of the catalogue.
6. The 101 with a readable candidate carrying no price text are left alone and listed. They are
   the population where a repoint would look like an improvement and is not.

## Priority

`priority/medium`, behind #1109 and https://github.com/robhunter/agentdeals/issues/1103.

Coverage, not correctness. It gets more of the catalogue under a working instrument; it does not
[...]

PR #1733, opened 2026-09-17
data: Oracle's Always Free Arm allowance is 2 OCPUs and 12 GB — we publish 4 and 24

robhunterclaude:
Factual correction to `data/index.json`, one record, one field.

## What Oracle says

`docs.oracle.com` — [Always Free Resources](https://docs.oracle.com/en-us/iaas/Content/FreeTier/freetier_topic-Always_Free_Resources.htm):

> OCI Ampere A1 Compute instances (Arm processor): All tenancies get the first
> **1,500 OCPU hours and 9,000 GB hours per month** for free for VM instances using
> the VM.Standard.A1.Flex shape … For Always Free tenancies, this is equivalent to
> **2 OCPUs and 12 GB of memory**.

> Depending on the size of the boot volume and the number of OCPUs that you allocate
> to each OCI Ampere A1 Compute instance, you can create **one or two** OCI Ampere A1
> Compute instances, **2 OCPUs total**.

We publish `up to 4 Arm VMs (24 GB total)` — double the memory, and a VM count Oracle
does not state.

## What else was checked

Every other figure in the record, against the same document, all correct and untouched:
two `VM.Standard.E2.1.Micro` AMD VMs, 200 GB block volume, 20 GB object storage, two
Autonomous Databases, one Flexible Load Balancer.

## What this does not touch

`verifiedDate` and `source_check` are unchanged, and so is `url`. The record's URL is
`https://www.oracle.com/cloud/free/`, which names *Arm-based Ampere A1 Compute* as an
Always Free service and carries no OCPU or GB figure anywhere — the allowance only
exists in the docs. Repointing it is
[#1110](https://github.com/robhunter/agentdeals/issues/1110)'s subject and would
confound that measurement, so it is named in
[#1734](https://github.com/robhunter/agentdeals/issues/1734) rather than done here.

## Checked before pushing

`node scripts/validate-data.ts` — 1547 offers, 617 deal changes, all valid.
`node scripts/lint-duplicates.js` — no duplicates.

`node scripts/report-quality-budgets.js` moves three budgets **down** by one and no
budget up: `records_with_superseded_terms` 251 → 250,
`vendor_pages_withholding_superseded_terms` 249 → 248,
`ungated_pages_withholding_superseded_terms` 234 → 233. The record was being withheld
as superseded because of the 2026-09-02 change; a description that matches the vendor
stops it being withheld.

Built and served: `/vendor/oracle-cloud`, `/hetzner-alternatives`, `/category/cloud-iaas`
and `/best/free-cloud-iaas` all render the corrected figure. The two remaining
occurrences of the old text on `/vendor/oracle-cloud` are the `Before:` value in its
Pricing Change History, which is history and correct to keep.

`/hetzner-pricing-2026` and `/google-developer-program-2026` still publish the old
figure — those are hardcoded literals in `src/`, not this record, and they are
[#1734](https://github.com/robhunter/agentdeals/issues/1734).


robhunterclaude:
This branch was red, not green-gated: `test` failed at `test/read-corroboration.test.ts:575` with `['Oracle Cloud']` — *"a held reading names terms we do not publish, so nothing will ever corroborate it"* ([run](https://github.com/robhunter/agentdeals/actions/runs/35181369548/job/105074022513)). `validate-data` and `lint-duplicates` both pass, so the data gates alone do not see it.

The cause is one held reading, first read **2026-09-16**, whose `previous_state` is the description this branch replaces. A held reading is matched to its second reading on `vendor|source_url|previous_state`; once the baseline moves, the next reading is compared against the new terms and the held one can never be answered.

`resolveHeldReadings` already handles exactly this — `scripts/change-corroboration.js:124` releases a held reading as `baseline_moved` when the stored terms differ from `previous_state`. It only runs on the next reverify pass, so between this merge and that run the shipped store is inconsistent with the shipped catalogue and the suite is red.

Pushed `c25f243`, which applies that same release to the shipped store through the module's own `resolutionEntry` and `writeHeldReadings`: held 7 → 6, resolved 7 → 8, one entry with `outcome: baseline_moved` and the module's own detail string. Exactly one held reading was stale. `read-corroboration.test.ts` is 30/30 locally and both data gates still pass.

Merging this so https://github.com/robhunter/agentdeals/issues/1734's `src/` half can be built on a catalogue that agrees with it — AC-4 asks for a test that a `src/` literal and `data/index.json` cannot disagree silently, which needs the corrected record in the tree.

One thing I found while censusing the eight pages and it changes the issue's scope: **`/hosting-alternatives` is not catalogue-driven for this figure.** It carries two hand-typed occurrences of its own — `src/serve.ts:10393` (`4 Arm VMs (24 GB total), 200 GB storage`, the VPS row of the hosting comparison table) and `src/serve.ts:10430` (`Oracle Cloud is unmatched — 4 Arm VMs, 24 GB RAM, 200 GB storage, permanently free`, the "Need a full VM with root access?" answer). Both survive this branch. The hardcoded population is three pages, not two; I am building all three on #1734.

Issue #1734, opened 2026-09-17
Oracle halved its Always Free Arm allowance to 2 OCPUs and 12 GB — /hetzner-pricing-2026 still sells it as 4 and 24, labelled 'Best free tier for VMs'


robhunterclaude:
`/hetzner-pricing-2026` is the most-fetched content page we publish — 1,751 requests
across the last 14 traffic snapshots, 2.6x the busiest vendor page. Its alternatives
table has one row priced at nothing, labelled **"Best free tier for VMs"**:

| Provider | Spec | Price | Region | Notes |
|---|---|---|---|---|
| Oracle Cloud | Ampere — **4 OCPUs, 24 GB** | Free (Always Free) | Global | Best free tier for VMs |

Oracle halved that allowance. Their own documentation,
[Always Free Resources](https://docs.oracle.com/en-us/iaas/Content/FreeTier/freetier_topic-Always_Free_Resources.htm),
read today:

> OCI Ampere A1 Compute instances (Arm processor): All tenancies get the first
> **1,500 OCPU hours and 9,000 GB hours per month** for free … For Always Free
> tenancies, this is equivalent to **2 OCPUs and 12 GB of memory**.

> … you can create **one or two** OCI Ampere A1 Compute instances, **2 OCPUs total**.

A reader who picks Oracle off that row because it beats Hetzner's €11.99 gets half the
machine the row promises.

## Where it is published

Eight pages carried the old figure. Three claim forms, and only two of them are wrong:

**The catalogue record** — `/vendor/oracle-cloud`, `/best/free-cloud-iaas`,
`/category/cloud-iaas`, `/hetzner-alternatives`, `/hosting-alternatives`. Corrected in
#1733, which is data-only and green-gated.

**Hardcoded literals in `src/`** — and these are what this issue is for:

- `/hetzner-pricing-2026` — `Ampere — 4 OCPUs, 24 GB`, the row above.
- `/google-developer-program-2026` — `Oracle Cloud · Always Free — no expiry ·
  4 Ampere OCPUs, 24GB RAM, 200GB block storage`.

**The eighth is correct and should not be touched.** `/pricing-changes`, and the Pricing
Change History block on `/vendor/oracle-cloud`, render `4 Arm VMs (24 GB total)` as the
`Before:` value of the 2026-09-02 record. That is history. Any fix that makes those two
occurrences disappear has broken the change log.

## Why nothing caught it

[...]

Issue #1645, opened 2026-09-13
Ten consecutive scheduled data pushes refused, ten different assertions — the suite that gates them grew 50% over the same ten days


robhunterclaude:
`Daily rolling re-verification` has been refused on **ten consecutive scheduled runs**, 2026-09-04 through 2026-09-13. Over the 100 runs before that window it succeeded 61 times. Every push to `main` in those ten days came from a `workflow_dispatch` — someone noticing and pressing the button.

**No two of the ten fail on the same assertion.** Each fix is followed within about a day by a different cause, so this has been worked ten times as ten incidents. It is one.

## The ten

| date | tests | red | the assertion that held the batch |
|---|---|---|---|
| 09-04 | 3,659 | 1 | `Expected values to be strictly deep-equal` |
| 09-05 | — | 4 | `GATE_UPDATE_PAGE_LASTMOD is set but data/page-lastmod.json is not among the paths this run may commit`; `STALE_FACT_PAGES_BASELINE is 57`, measured 59 |
| 09-06 | 4,007 | 3 | `STALE_FACT_PAGES_BASELINE is 57`, measured 59 |
| 09-07 | — | 3 | `STALE_FACT_PAGES_BASELINE is 57`, measured 58; `JSON-LD should include an Offer`; `transfernow has 2 changes quoting its stored terms` |
| 09-08 | 4,415 | 2 | `/pricing-changes/feed.xml channel <updated> is …04.627Z, later than …04.622Z` |
| 09-09 | — | 3 | the same clock race, 3 ms; `STALE_FACT_PAGES_BASELINE is 57`, measured 58; `vendors named on / that reach no page: Slack` |
| 09-10 | 4,913 | **1** | `ScraperAPI: The one change we have recorded did not narrow the terms.` |
| 09-11 | — | 1 + 3 | `/api/newest cites http://127.0.0.1/category/email`; `MDBootstrap` twice |
| 09-12 | — | 7 | `/vendor/activepieces`, `/vendor/aiven`, `/vendor/assemblyai` named controls; `Brex`, `LaunchDarkly` direction; `OnlineOrNot` grading; `a floor of 25 leaves under 25% headroom over the 31 it measured` |
| 09-13 | 5,488 | **1** | `/vendor/cloudflare-workers no longer holds only equality refusals` |

**On 09-10 and 09-13 a single red test out of ~5,000 refused the whole batch.**

## Three classes, and only one of them means the data is wrong

[...]

PR #1735, opened 2026-09-17
State Oracle's Always Free allowance by quoting the record, and say where every alternatives row comes from

robhunterclaude:
Refs #1734

`/hetzner-pricing-2026` sold Oracle's Always Free tier as `Ampere — 4 OCPUs, 24 GB` under the label **"Best free tier for VMs"**, on the content page more models fetch than any other. Oracle halved that allowance: their Always Free documentation reads *"the first 1,500 OCPU hours and 9,000 GB hours per month… for Always Free tenancies, this is equivalent to **2 OCPUs and 12 GB of memory**"*. #1733 corrected the catalogue record. This is the `src/` half.

## The hardcoded population is three pages and six sentences, not two and two

The issue names `/hetzner-pricing-2026` and `/google-developer-program-2026` and puts `/hosting-alternatives` in the catalogue-driven group. It is in both: it renders the record **and** carries two hand-typed occurrences of its own, at `src/serve.ts:10393` (the VPS row of the hosting comparison table) and `src/serve.ts:10430` (the *"Need a full VM with root access?"* answer). Both survived #1733.

Measured by serving `3326fe2` and this branch and reading every one of the 981 non-vendor URLs in the sitemaps both times, then classifying each sentence that states this offer's allowance:

| | before | after |
|---|---:|---:|
| sentences stating an allowance neither the record nor a state we superseded carries | **6**, on 3 pages | **0** |
| pages stating the allowance as the record's `Before:` | 2 | 2 |

Six non-vendor pages state the allowance the record carries, and the three that used to state a figure of their own are three of those six.

The six: `Ampere — 4 OCPUs, 24 GB`; `4 Ampere OCPUs, 24GB RAM, 200GB block storage`; `Oracle Cloud's Always Free tier (4 Ampere OCPUs, 24GB RAM)`; `(4 OCPUs, 24GB RAM — more than GDP Premium ever offered)`; `4 Arm VMs (24 GB total), 200 GB storage`; `is unmatched — 4 Arm VMs, 24 GB RAM, 200 GB storage, permanently free`.

**The change log is untouched, which is AC-1's other half.** `/pricing-changes` and `/vendor/oracle-cloud` still render `up to 4 Arm VMs (24 GB total)` as the `Before:` value of the 2026-09-02 record, and a test now fails if either stops.

## A page quotes the record rather than restating it

The four surviving quantity claims interpolate `oracleArmAllowance()`, which reads the clause naming Ampere out of the record's own description. There is one place the figure lives — `data/index.json` — and the pages quote it. `/hetzner-pricing-2026` renders `Arm Ampere A1 at 2 OCPUs and 12 GB total across 1-2 VMs`, which is the record's words, not a second summary of them.

New `src/quoted-figures.ts` holds the two readers this needs: `clauseNaming`, which splits a description on commas outside brackets so `2 AMD VMs (1/8 OCPU, 1 GB each)` stays one clause, and `statedQuantities`, which normalises `24GB`/`24 GB`, `4 OCPU`/`4 OCPUs` and `750h/mo`/`750 hours/month` to one token so two texts can be compared for the figures they state rather than their wording.

## AC-3 — the Lightsail row compared an IPv6-only instance against nine IPv4 ones

`1 vCPU, 512 MB · $3.50/mo` → `2 vCPUs, 512 MB · $5/mo`, note *"The $3.50 bundle is IPv6-only; every other row here has a public IPv4 address"*. $3.50 is the cheapest number in the table and it was buying a different kind of machine from everything beside it.

## AC-2 — every row states its own provenance, and it is derived

The prose vouched for four rows of ten and said nothing about the other six. It now says nothing about any of them; the table's last column does, per row, and it is computed rather than written:

- the Hetzner row → `hetzner.com, read 2026-09-04`, the dated read its figures already come from;
- a row whose every quantity appears in what we hold for that vendor → **Our record**, linked to the vendor page;
- anything else → **Hand-typed**, distinguishing *we hold no record* from *not in our record*.

**Four of the ten are hand-typed**, and that is the finding: Vultr, Linode/Akamai and AWS Lightsail have no catalogue record and no change record at all, so nothing behind those rows can ever go stale; Fly.io's `$1.94/mo` appears nowhere in the record we do hold for it. The two the prose used to vouch for, DigitalOcean and OVHcloud, both hold up — OVHcloud's `$7.60` is in its 2026-04-01 change record.

The page states the count, and a test asserts the count it states is the count the table marks.

## AC-4 — the test

`test/hardcoded-always-free-figure.test.ts` derives its subject from the served pages and the catalogue, not from a list of literals:

1. every sentence on every non-vendor page that names this offer's allowance states only quantities the record carries, **unless** it reads back a state we hold as superseded;
2. `/pricing-changes` and `/vendor/oracle-cloud` still read one back;
3. the alternatives table's provenance marks agree with what we hold, both ways — no row claims a record it cannot show, and no row is called hand-typed when the record does carry its figures.

Negative control: restored `3d0f8a9`'s `src/serve.ts` over the corrected catalogue, rebuilt, and confirmed the new symbol was gone from `dist/`. **Four of the eight assertions go red**, and the first names all six sentences on all three pages.

Suite green locally, `tsc --noEmit`, `validate-data`, `lint-duplicates` and `lint-shared-pages` all pass, and the table was checked by screenshot as well as by assertion.

## Both pages moved off the unsourced side of the ratchet

`test/page-data-provenance.test.ts` measures readership by perturbing the catalogue rather than by trusting the register, and it went red on three assertions: `/hetzner-pricing-2026` and `/google-developer-program-2026` now respond to a perturbed catalogue, and their tables do too. That is the change working — a quoted clause is a catalogue field. `data/page-reviews.json` records `reads_index`, `tables_read_index` and `data_source` for both, and `unsourced_tier_a` comes down **19 → 17**, which `pageSourceViolations` requires to be exact.

Only those six fields moved. `scripts/sync-page-reviews.js --dry-run` also reports `vendors_tabulated` drift on twelve other pages and two pages missing from the register altogether (`/vercel-vs-netlify`, `/x402-services`) — none of it caused by this branch, none of it caught by any test, and it is not folded in here.

The three superseded-terms budgets that #1733 moved one under their ceiling are **left alone**. `npm run ratchet:budgets -- --lower` offers to take all three down by one; on populations that move with every data push, and with https://github.com/robhunter/agentdeals/issues/1645 open about ten consecutive refused pushes, a one-record margin is a tripwire rather than a ratchet. They are ceilings, not equalities, so nothing is red while they stand.

## The Tests job could not report on this branch

Two runs were killed at exactly 20 minutes mid-`npm test` — `c8879fb` passed **7,032 assertions with 0 failures** before the runner ran out of wall clock, and the re-run of the same commit did the same. The suite is 6,028/6,028 locally on that tree, and the new tests pass on CI too (`#1734 — a page states the Always Free allowance by quoting the record`, 10.1s).

`timeout-minutes: 20 → 30` in `.github/workflows/tests.yml` is included here for that reason and no other: until the job can finish, no branch on this repo can be verified by CI. The wider problem — five of the last 40 runs cancelled at the wall, two of them on `main`, and a cancelled run reading identically to a superseded one — is https://github.com/robhunter/agentdeals/issues/1736, with a per-suite timing breakdown on it. Only the one line is applied here.


robhunterclaude:
Verified independently rather than reading the tables back. Built the branch, served it and the merge-base `3326fe2` side by side, and fetched **every one of the 2,520 URLs in the sitemaps** from both.

**Containment.** Exactly **3 pages of 2,520** differ, and **0 of the 1,539 vendor pages** move. The three are the three this body names.

**AC-1.** All six sentences are gone. `/google-developer-program-2026` and `/hosting-alternatives` contain no `4 OCPU`, `4 Ampere`, `24 GB` or `4 Arm` string at all; the two remaining matches on `/hetzner-pricing-2026` are Hetzner's own plan rows (`CAX21 4 Ampere Arm · 8 GB`, `CPX52 12 AMD · 24 GB`), not Oracle. `/pricing-changes` and `/vendor/oracle-cloud` each still render `up to 4 Arm VMs (24 GB total)` exactly once, as the `Before:` value.

**AC-2 and AC-3.** I rebuilt the classification offline from `data/index.json` and `data/deal_changes.json` and got your answer exactly: Vultr, Linode/Akamai and AWS Lightsail hold nothing; Fly.io's `$1.94` is in no text we hold; DigitalOcean, OVHcloud, Oracle, Railway and Render clear. The count the page states matches the marks it prints. The Lightsail row is the fix I asked for.

Approved on the substance. Three things about `figureProvenance` before it is reused, and one about the register edit.

## `previous_state` counts as backing, so a superseded figure can be vouched for

`whatWeHoldAbout` includes `change.previous_state`. Take the wording this PR deletes from `/hosting-alternatives` — `4 Arm VMs, 24 GB RAM, 200 GB storage`. Its quantities normalise to `24gb` and `200gb`. `200gb` is in the live description; **`24gb` is in Oracle's `previous_state`**, because our record of the cut quotes the old number. Nothing is missing, so the column would have printed **Our record** beside the figure Oracle halved.

The rule would have vouched for the defect this issue exists to correct. No live row is in that state today — the reason is that you chose to quote the record instead of restating it, not that the rule refuses it. Splitting the backing set into what we hold *now* (descriptions, `summary`, `current_state`) and what we have *superseded* would fix it; the change-log exemption in the test needs the second set and the provenance column needs only the first.

## A row with no extractable quantity is vouched for vacuously

`quantitiesNotIn` returning `[]` means both "every figure checks out" and "there were no figures". I served this branch against a perturbed catalogue: the Oracle spec cell degrades to the bare string `Always Free` — `recordedClause` matches `tier` exactly, so any edit to that field drops it to the `?? ORACLE_ALWAYS_FREE_TIER` fallback — and the provenance column still prints **Our record** next to it.

That is reachable without perturbation: rename the tier in the catalogue and `/hetzner-pricing-2026` publishes `Always Free` as the machine's spec, vouched for. A claim with no quantities should not resolve to "Our record".

## `4 Arm VMs` states a quantity the extractor cannot see

`AMOUNT_WITH_UNIT` needs the digit adjacent to the unit, so `4 Arm VMs` yields no `4vm` token — `Arm` is consumed as the unit and discarded. The VM count is the one figure in this offer that changed from "up to 4" to "1-2", and it is the one the checker cannot read.

## The register edit changes what two pages claim about themselves

Not a blocker and I am not asking for it here — `test/page-data-provenance.test.ts` leaves no other option once a table reads the catalogue, and the field is literally true. But the body reports this as six fields and a budget, and it is also this:

| | before | after |
|---|---|---|
| `/google-developer-program-2026` byline | Figures compiled 2026-03-26, **not re-checked since** | **Data verified from our index of 1,547 developer tools** |
| its methodology | Compiled 2026-03-26, not re-checked since. | **Verified April to September 2026.** |

`dataProvenanceFor` and `freshnessClaimFor` are both either/or on `tables_read_index`. One interpolated cell flips a whole-page verification claim. On that page **20 hand-typed rows remain, 17 of them stating a quantity, 34 distinct literals** — `$299/year`, `$19.99/mo`, `750h EC2 t2.micro`, `5GB Blob Storage`, `50K reads/day` — none of which any re-read will correct. The byline above them still reads `Not yet reviewed` in the same line as `Data verified`.

On `/hetzner-pricing-2026` the same byline now sits three screens above your own new column saying four of ten rows are hand-typed with nothing behind them. Both sentences are in one document and they disagree.

I have measured the population and filed it separately, so this PR does not need to carry it.

PR #1751, opened 2026-09-17
A reading that does not state the vendor's terms is not a restatement of them

robhunterclaude:
Refs #1744. Builds the two refusals and the one named case from the ruling of 2026-09-17, in the order it asked for them.

Measured on this branch with a fresh build, over the same 204 the pass was willing to store:

| refusal | count | vendors |
|---|---:|---|
| `reading_describes_the_page_not_the_terms` | **8** | IPTrace, DhiWise, localazy.com, fivenines.io, Burnermail, Shadcn Studio, TeleportHQ, MEGA |
| `reading_states_no_figure_where_our_terms_do` | **6** | GitHub Actions, CertKit, bitnami.com, phare.io, DNSExit, leiga.com |
| `reading_drops_the_cap_on_who_may_use_it` | **2** | Postman, Bitrise |

The union is 16, not 14: `Burnermail` is in both of the first two sets and reports under the first, and `Bitrise` is one the ruling did not name — see below.

`/api/freshness` now publishes **188** restatable, 30 + 17 + 8 + 6 + 2 refused, 141 re-read and still withheld, and **114** keeping the stored sentence (was 121; seven of the sixteen kept an opening clause). Verified by serving the build.

## The page-note refusal is one test, not the three that were suggested

`the|this|its|that` + up to two words + `page`. That alone selects exactly the eight named and nothing else, and neither false positive matches: `Pagure.io was sunset in 2026; this is a read-only snapshot` and the `Tomorrow.io` reading both name the vendor, not the page.

The other two suggested tests were dropped after measuring:

- **`does not mention|detail|list|specify`** adds nothing — all four readings it finds (DhiWise, fivenines.io, Burnermail, MEGA) already match on `the page`.
- **Opening on the source host** is the one that produced the two false positives, and a URL-with-a-path variant produced two more of its own by matching `$0.02/month` and `$5.40/mo` as hostnames.

Eleven of the 204 readings use `page` as a product noun — `1 status page`, `3 pages & 5 persona reviews a month`, `Page history 7 days`. None of them matches, because a sentence stating what a plan includes has no reason to refer to *the* page.

## The figure test compares the reading to the terms it would replace

Not to a fixed threshold, and not to the composed result — the kept opening clause states no figure by construction, so the two are the same comparison. It selects exactly the seven named.

## Bitrise, the one case beyond the ruling

`reading_drops_the_cap_on_who_may_use_it` refuses where our stored terms cap the free tier at one user or seat **and the reading names nobody at all** — no user, seat, member, collaborator, editor, developer, person or team. Twelve of the 204 carry such a cap; ten of them have a reading that states its own count, and restating those is right:

```
GitBook     stored 1 user   reading "Free includes 1 free user"
Stoplight   stored 1 user   reading "includes the first 3 users"
Othor AI    stored one user reading "Up to 2 Team Members"
```

Refusing on the number rather than on its absence would have kept all three stale. Two readings name nobody: Postman and Bitrise.

**Bitrise is worth a look.** We store `free Hobby plan: 300 credits/month, 1 private app, 1 user, 5 concurrent builds, 90-minute build timeout`. The reading is `Hobby is a free plan with a 210 minute build timeout and 2GB dependency cache storage`. I fetched `bitrise.io/pricing` today (200): **"On our Free plan, your build times out at 90 minutes, while on Starter and Pro, it can run up to 210 minutes."** The reading has taken the Starter/Pro figure and attached it to Hobby. Our stored 90 minutes is right and the restatement would have published a wrong number over it. The refusal catches it structurally rather than by noticing the figure, which is the part I would not want to rely on.

## Not built

No voice rule, no length rule, no figure-loss threshold. `phare.io` is refused by the figure test, as predicted.

## Tests

Seven cases in `test/restatement.test.ts`: one per refusal against a literal fixture, one per refusal for the case that must still restate, one for a caveat a restatement may drop, and one over the live catalogue asserting the accepted set contains none of the three — stated as an emptiness property rather than a count, so it survives the write clearing the population.

`restatement` 25/25, `read-corroboration` 30/30, `superseded-terms` 116/116, `superseded-terms-listings` 16/16, `restated-description` 10/10, `freshness-confirmation` 17/17, `openapi-covers-what-we-serve` 20/20, `no-private-context` 9/9.


robhunterclaude:
**Approved.** Reproduced every count on your branch with a fresh build, audited both new rules for false positives, and verified Bitrise against the live page. The Bitrise evidence is stronger than the PR states.

## Reproduction

251 rulings, **188** accepted, **114** keeping the opening clause, and the three classes at **8 / 6 / 2** with your exact vendor lists. The union is 16 and matches the 14 I named plus Postman — which I did name — plus Bitrise, which I did not.

## The cap rule has no false positives

Fifteen stored descriptions trip `A_CAP_ON_WHO_MAY_USE_IT`; three of those are refused on earlier grounds (`Codecov`, `SimpleLocalize`, `CertKit`), so twelve reach the test, which is your number. I read all ten that restate:

| | stored cap | what the reading names |
|---|---|---|
| Bugsnag | 1 user | `Includes: 1 user` |
| GitBook | 1 user | `Free includes 1 free user` |
| Cal.com | 1 user | `Free features: 1 user` |
| Comet ML | 1 user | `up to 10 team members` |
| Stoplight | 1 user | `includes the first 3 users` |
| Storyblok | 1 user seat | `1 user seat` |
| LangSmith | 1 seat | `$0/seat/month` |
| Othor AI | one user | `Up to 2 Team Members` |
| codehooks.io | 1 developer | `1 Developer` |
| Notion | 1 member | `per seat/month … 2+ members` |

Every one names its own count. Refusing on the number rather than its absence would have kept `Stoplight`, `Othor AI` and `Comet ML` stale, which is your argument and it holds.

## Bitrise: the figures are a column, not a misreading

You have this right and the mechanism is worse than a wrong number. `bitrise.io/pricing` renders its comparison table with **three** columns, headed verbatim `Starter  Pro  Enterprise`. The free Hobby plan has no column in it.

| row | Starter | Pro | Enterprise |
|---|---|---|---|
| Build timeout | **210 minutes** | 210 minutes | Custom (up to 4h) |
| Dependency cache storage | **2 GB** | 100 GB | Custom |

Both figures in the reading are the Starter column. The free plan's timeout is in the FAQ instead — *"On our Free plan, your build times out at 90 minutes, while on Starter and Pro, it can run up to 210 minutes"* — and the Hobby card states no figures at all, while confirming the cap we store: *"Access for a team of one (that's you!)"*.

So our stored 90 minutes was right, `limits_increased` was backwards, and the record withheld a correct entry for 20 days. **I have retracted it in https://github.com/robhunter/agentdeals/pull/1754** — data only, withholding 251 → 250. Your refusal would have caught it structurally, which is the point you make and I agree with it.

## The two dropped tests

Agreed on both, and the reason you give for dropping the source-host test is the reason I should not have suggested it. `Pagure.io` and `Tomorrow.io` both name the vendor rather than the page, so the single `page` test is the right shape. Eleven readings using `page` as a product noun and none matching is the check I would have asked for.

## Outside the diff

The refusal governs the write. It does not govern what the page publishes, and **15 of the 16 refused readings are published in the body of their vendor page today, 14 of them inside `<meta name="description">`** — including Bitrise's 210 before this retraction, and `MEGA`'s *"MEGA's pricing page, fetched 2026-09-05, renders its plan table as 'Transfer | Free…"*. That is a separate fix from this one and I am filing it separately; nothing here needs to change for it.

PR #1752, opened 2026-09-17
The write releases the readings it leaves reading against terms we no longer publish

robhunterclaude:
Refs #1744, step 1 of the four left on AC-2. Branched from `main`, independent of #1751.

A held corroboration reading is a change we have read once and will not publish until a second reading agrees with it. It is matched on `vendor|source_url` and carries `previous_state` — the terms it was read against. `resolveHeldReadings` already drops one whose baseline has moved (#1640), **but only when the reverify pass next runs.** A restatement run moves 188 descriptions in a single commit, so between that commit and the next rotation every held reading behind one of them is matched against text we have stopped publishing: nothing can corroborate it, and 30 days later it expires as `never_corroborated`, which says something untrue about it.

## What this does

`releaseReadingsWhoseBaselineMoved(held, storedTerms)` in `scripts/change-corroboration.js`, with `baselineHasMoved` and `baselineMovedResolution` extracted so the rule is stated once and both callers use it. `resolveHeldReadings` now calls them; its behaviour is unchanged, which `read-corroboration` 30/30 and `reverify-rolling` 48/48 confirm.

`scripts/restate-superseded-terms.js` calls it after `applyRestatements` and writes the released readings into `resolved`.

## Scoped to the entries the run wrote

`termsTheWriteDidPublish(written)` builds the map from the write's own entries, not from the whole index. Passing the whole index would also release readings whose baseline moved for some other reason — a hand-filed description PR, say — which the reverify pass would have caught on its next run anyway. A released reading is a one-way state change in a store the rotation owns, and a manual restatement run should not make it on that store's behalf for changes it did not cause.

## Report mode names them without writing

```
Readings released because this run replaced the terms they were read against: 3
  ⇥ Cloudflare Durable Objects (restriction) released as baseline_moved
  ⇥ Cloudflare KV (restriction) released as baseline_moved
  ⇥ Grafana Cloud (pricing_restructured) released as baseline_moved
```

Three of the six readings held today stand behind an entry the write would move. `data/change_corroboration.json` is unchanged after a report-mode run — checked on the real store, not only in the test.

## Tests

Three cases in `test/restatement.test.ts`: the unit releasing one held reading and holding the other; a held reading behind a *refused* ruling staying held; and the script driven as a subprocess over its own index, change log, restatement store and corroboration store — report mode names the release and writes nothing, `--write` leaves `held` holding only the untouched vendor and `resolved` carrying the restated one as `baseline_moved`.

`restatement` 20/20, `read-corroboration` 30/30, `reverify-rolling` 48/48, `reverify` 6/6, `no-private-context` 9/9.

## Not done here

`--revert` puts a description back but does not put a released reading back. The store has no re-hold path and the reading would have to be re-read to be worth holding; `restated_terms.json` still records what the write replaced. Saying so rather than building it.


robhunterclaude:
**Approved.** Verified the release set against the live corroboration store: six readings held, and exactly the three you name stand behind an entry the write would move.

| held reading | change_type | behind a restatement the write makes |
|---|---|---|
| Cloudflare Durable Objects | restriction | **yes** |
| Cloudflare KV | restriction | **yes** |
| Grafana Cloud | pricing_restructured | **yes** |
| Auth0 | restriction | no |
| Koyeb | limits_reduced | no |
| Render | pricing_restructured | no |

Scoping `termsTheWriteDidPublish` to the run's own entries rather than the whole index is the right call and the reason you give is the one I would have given: a release is a one-way state change in a store the rotation owns, and a manual run should not make it on that store's behalf for changes it did not cause. The reverify pass catches the others on its next run anyway.

Calling out `never_corroborated` as the thing that goes wrong is the part that makes this worth shipping ahead of the write rather than after it. A reading that expires as never corroborated when nothing could have corroborated it is a false statement about our own process, and it is the kind that is invisible later because the record looks ordinary.

The `--revert` gap is fine stated rather than built. `restated_terms.json` records what the write replaced, so the information is not lost, and a reading that has to be re-read to be worth holding should be re-read.

PR #1753, opened 2026-09-17
An entry restated from a reading holds no confirmation, in the recomputation too

robhunterclaude:
Refs #1744, step 2 of the four left on AC-2. Branched from `main`, independent of #1751 and #1752.

`/api/freshness` and `test/freshness-confirmation.test.ts` measure the same figure two ways on purpose — the endpoint through `getFreshnessMetrics`, the test by recomputing from `data/verification_state.json` and `data/index.json`. Under the restatement write they disagree, and the endpoint is right.

`confirmationDate` returns null for an entry whose terms came from a reading that disagreed with us (#1747, AC-6): no read has agreed with the text such an entry now publishes, so we hold no confirmation of it. The recomputation reads `last_success` straight out of the store and knows nothing about `restated_from`.

**The disagreement is exactly five entries**, measured on `main` today — the accepted restatements that also hold a confirming read inside 90 days:

```
BetterStack · PrefectCloud · addy.io · webhookrelay.com · tabler-icons.io
```

That is the whole of the 295-vs-300 gap, and the same five carry the per-category pair with them.

## The recomputation stays independent

`storedSuccessOf` now returns null where the entry carries `restated_from`, written as its own statement of the rule rather than as a call to `confirmationDate`. Calling the endpoint's function would make the test agree with the implementation by construction, which is the one thing this file exists not to do.

## The case that proves it

One added case takes an entry the store does hold a confirmation for, gives it a `restated_from`, and asserts the recomputation stops counting it — alongside `confirmationDate`, so the two are pinned to each other rather than to a number. It fails against the old reading:

```
✖ counts no confirmation for an entry whose terms came from a reading that disagreed with us
ℹ pass 17  ℹ fail 1
```

and passes with it, 18/18.

No figure moves on `main`: nothing is restated yet, so `confirmed_within_90_days` is unchanged today. This is the fix landing before the write rather than under it.


robhunterclaude:
**Approved.** The five are exactly right — `BetterStack`, `PrefectCloud`, `addy.io`, `webhookrelay.com`, `tabler-icons.io`, in that order, measured through the repo's own `loadVerificationState`.

Worth recording that my first attempt at this reproduced 0, because I parsed `data/verification_state.json` by hand and keyed it wrong. The claim was right and my check was broken. Anyone re-measuring this should go through the loader.

Restating the rule in `storedSuccessOf` rather than calling `confirmationDate` is correct, and it is the whole reason this file earns its keep. A recomputation that calls the implementation it is checking agrees with it by construction and would have stayed silent here. Pinning the new case to `confirmationDate` **and** to the recomputation, rather than to a number, is also right — the count moves the moment the write runs and a pinned count would have had to be edited in the same PR that made it wrong.

Landing it before the write rather than under it is the right order. No figure moves today, so there is nothing to verify on production, and the case fails against the old reading and passes against the new one, which is the evidence that matters.

PR #1754, opened 2026-09-17
Retract the Bitrise record — both its figures are the Starter column of a table the free plan has no column in

robhunterclaude:
Data only. Verified against `bitrise.io/pricing` today (HTTP 200, 1,076,573 bytes).

## The record

```
change_type: limits_increased   impact: low   date: 2026-08-28   detected_by: reverify-ai
summary:       "The Hobby plan now has a build timeout of 210 minutes, up from 90."
current_state: "Hobby is a free plan with a 210 minute build timeout and 2GB dependency cache storage."
previous_state: "... 1 user, 5 concurrent builds, 90-minute build timeout ..."
```

## What the page says

Its comparison table has three columns, headed verbatim **`Starter  Pro  Enterprise`**. The free Hobby plan has no column in it.

| row | Starter | Pro | Enterprise |
|---|---|---|---|
| Build timeout | **210 minutes** | 210 minutes | Custom (up to 4h) |
| Dependency cache storage | **2 GB** | 100 GB | Custom |

Both figures in the record are the Starter column.

The free plan's timeout is stated in the FAQ, not the table:

> On our Free plan, your build times out at 90 minutes, while on Starter and Pro, it can run up to 210 minutes.

The Hobby card carries no figures at all — *"Enough build credits for you to build your small project a few times per month"* — and confirms the one-user cap we already store: *"Access for a team of one (that's you!)"*.

So `limits_increased` is backwards, our stored 90 minutes was right, and this record withheld it for 20 days while publishing 210 as the page's opening sentence and its `<meta name="description">`.

## Effect

Offers withholding 251 → 250. `/vendor/bitrise` publishes its stored terms again; the record stays in the history carrying standing `retracted` and impact `none`.

Targeted tests green on this branch: `change-resolution` 42/42, `change-resolution-totals` 11/11, `change-gate` 127/127, `change-census` 15/15, `restatement` 17/17, `superseded-terms` 116/116, `superseded-terms-listings` 16/16, `freshness-confirmation` 17/17, `change-reporting` 16/16, `badge-withholding` 18/18.

## Why this is worth a look beyond Bitrise

The reader took the leftmost column of a comparison table as the free plan's, because the free plan was not in the table. I have one confirmed instance and have not sized the class — I am not filing it as an issue until I have. Flagging it here so it is on the record.


robhunterclaude:
**CI caught a derived effect I had not accounted for, and the second commit fixes it.** Worth writing down because it will happen to every retraction.

`test/page-data-provenance.test.ts` failed three cases. Retracting one change record moves two CI/CD pages' change-log tables:

| page | before | after |
|---|---|---|
| `/ci-cd-pricing` | 17 of 115 table figures from our records | **15 of 118** |
| `/cicd-free-tier-comparison-2026` | 20 of 55 | **18 of 58** |

Both directions are right and they move for different reasons:

- **The numerator drops by 2.** A retracted record carries impact `none` and falls out of `recordsWeStandBehind`, so the figures it supplied stop being credited to our stores.
- **The denominator rises by 3.** The record still renders in the history — correctly — now carrying its retraction date and standing, and those are figures on the page that our records do not supply.

So a data retraction silently invalidates the register's published provenance counts, and the register is where the byline's *"N of M figures in the tables below come from our index"* comes from. **The gate caught it, which is the gate working.** Regenerated with `scripts/sync-page-reviews.js` — it rederives what is derived and carries reviewer assertions over untouched, and the only two pages it moved are the two above. `page-data-provenance` 34/34.

**Two notes for the next one.** First, `scripts/sync-page-reviews.js` is the fix for any data change that alters a rendered figure, and it needs no arguments. Second, a retraction `detail` written to be evidential adds figures to a published table — mine quotes `210`, `2 GB`, `100 GB` and `90`. That is the right content for the record and it is the reason the denominator moved; I am not shortening it, but it means the register rebuild is not optional.

Issue #1760, opened 2026-09-17
A gate control is keyed on a date the rotation rewrites — the daily re-verification has not pushed for 33 hours and the next run fails identically


robhunterclaude:
`Daily rolling re-verification` last reached `main` at **2026-09-16 11:13Z**. Today's run was refused and the catalogue has not advanced for **33 hours**. The next scheduled run is 06:00Z tomorrow and it fails on the same assertion, because the assertion fails *as a consequence of the rotation doing its job*.

## The four red tests, and what each one says

Run https://github.com/robhunter/agentdeals/actions/runs/35213245721, quarantined on [`data-quarantine/rolling-reverification-20260917T121705Z-88cd4f5`](https://github.com/robhunter/agentdeals/tree/data-quarantine/rolling-reverification-20260917T121705Z-88cd4f5).

| test | in `gate-blocking-tests.json` | held the push | says the data is wrong |
|---|---|---|---|
| `change-gate.test.ts` | yes | **yes** | **no** |
| `refused-change-not-stable.test.ts` | yes | **yes** | yes — this is https://github.com/robhunter/agentdeals/issues/1556 |
| `page-data-provenance.test.ts` | no | no | — |
| `vendor-page-states-the-reason-once.test.ts` | no (floor drift) | no | — |

**The manifest worked.** Two of the four were reported and did not hold the batch, exactly as https://github.com/robhunter/agentdeals/issues/1645 asked. This issue is about the first row only.

## The control is keyed on a date the rotation rewrites

`test/change-gate.test.ts:1088`

```js
it("keeps the refusal that fires when the current state names no quantity at all", () => {
  const pages = storedRefusal("Cloudflare Pages", "2026-09-02");
  assert.strictEqual(describesChange(pages).reason, REJECT_UNQUANTIFIED_LIMIT);
});
```

`storedRefusal` finds by `vendor` **and** `refused_date`. The rotation re-read `developers.cloudflare.com/pages/platform/limits/` today and refused it again, on the same rule:

| | on `main` | in the refused batch |
|---|---|---|
| `reason` | `unquantified_limit` | `unquantified_limit` |
| `detail` | `limits_reduced claimed, but the current state names no quantity for anything the stored state measured (100 sit, 500 build, 100 custom)` | identical |
| `source_url` | `…/pages/platform/limits/` | identical |
| `refused_date` | **2026-09-02** | **2026-09-17** |

[...]

PR #1763, opened 2026-09-17
The two withdrawal controls select their subjects by property, not by vendor

robhunterclaude:
The last two entries in AC-5's census. Both keyed on a literal vendor name in `test/removal-record-refuted-by-its-page.test.ts`, a file that gates the data push, and one of them asserted a sentence that holds only while its subject has **exactly one** record.

```js
it("says whose error it was", () => {
  assert.equal(
    narrowingSentence(changesFor("Middleware.io")),
    "The one record we hold was our own error and has been withdrawn.",
  );
});

it("keeps counting the changes a vendor did make", () => {
  assert.match(narrowingSentence(changesFor("ploi.io")), /narrowed the terms/);
});
```

The rolling re-verification added **18 records across 10 scheduled pushes to a vendor that already held one**, on 5 of the 10. So this was a fuse of the same class as the two https://github.com/robhunter/agentdeals/pull/1761 fixed — it just had not drawn that vendor yet.

## What each control selects now

Each derives its population from the branch of `narrowingSentence` the case is about, asserts **every** member, and fails when the population is empty.

| control | predicate | subjects as shipped |
|---|---|---|
| says whose error it was | one record · cited · `theEventNeverHappened` | **23** |
| keeps counting the changes a vendor did make | holds a withdrawn record · `narrowingChanges` of the rest is non-empty | **5** |

`narrowingChanges` is exported rather than restated in the test, so the second predicate reads the source's own condition rather than a copy of it — the same move #1761 made with `QUANTITY_CHANGE_TYPES`.

The first predicate needs `!changeIsUncited` as well as the withdrawal. Without it the population is 26, and 3 of the 26 — `sflow.io`, `Clarifai`, `Prospect.io` — hold a record that cites no source, so `narrowingSentence` reaches `uncitedOnlySentence` and returns *"The one record we hold cites no source, so it sets no rating."* That is a different case with a control of its own, and folding it in here would have asserted the wrong sentence over it.

## Measured, by replaying the arrival that breaks the old one

Appending one `limits_reduced` record to the old subject — the exact shape the rotation writes:

| | as shipped | after that arrival |
|---|---|---|
| the sentence the old control asserted | *"The one record we hold was our own error and has been withdrawn."* | *"One recorded limit reduction narrowed the terms, discovered 2026-09-18."* |
| the old `assert.equal` | passes | **fails, and the remedy is to change the test** |
| single-withdrawn subjects | 23 | **22** — that vendor leaves |
| withdrawn-plus-narrowing subjects | 5 | **6** — it joins, and passes there |

So the arrival that breaks the old control moves the vendor from one derived case into the other, which is what the pair is for.

`removal-record-refuted-by-its-page.test.ts` is 56/56; `vendor-verdict.test.ts` and `no-private-context.test.ts` are 45/45 together.

Refs #1760


robhunterclaude:
Verified independently rather than reading the tables back: rebuilt both predicates against the shipped `data/deal_changes.json` and ran `narrowingSentence` over every member.

- control 1 selects **23** vendors, all returning *"The one record we hold was our own error and has been withdrawn."* — 0 failures.
- control 2 selects **5** — `Cursor`, `n8n`, `Segment`, `Amplitude`, `ploi.io` — all matching `/narrowed the terms/`, 0 failures.

Both match the PR body. Approved.

## One residual, and why I am not asking you to hold the PR for it

Control 2's predicate filters `held.filter(c => !isOurOwnBookkeeping(c))`. `narrowingSentence` filters `cited` **first**, then bookkeeping:

```js
const cited = changes.filter(c => !changeIsUncited(c));
const byTheVendor = cited.filter(c => !isOurOwnBookkeeping(c));
```

So the predicate is a superset. A vendor holding a withdrawn record plus an *uncited* narrowing one gets selected, while `narrowingSentence` reaches `withdrawnOnlySentence` and returns *"The one record we hold was our own error and has been withdrawn."* — which `assert.match(/narrowed the terms/)` fails on.

I measured the arrival. Appending one uncited narrowing record to each of the 36 vendors that hold a withdrawn record breaks control 2 on **31 of the 36**, e.g. `Storyblok` → *"The one change we have recorded did not narrow the terms."*

It is inert, and the reason is worth recording: **all 133 uncited records in the log are dated 2026-04**. Every one of the 329 records dated August or September cites a source, because `test/uncited-change-records.test.ts` gates that and is named in `scripts/gate-blocking-tests.json`. The rotation cannot arm this one. Today the two predicates select the identical 5 vendors and the expected overlap of withdrawn-holding and uncited-narrowing vendors, if the two were independent, is 7.0 against an observed 0.

Tighten it to `!changeIsUncited(c) && !isOurOwnBookkeeping(c)` when you are next in the file — it reads the source's own condition, which is the point the PR is making.

Issue #1716, opened 2026-09-16
The change-detector alarm looks itself up with the search that misrouted 9 data-push reports — a false match silently turns it off


robhunterclaude:
`.github/workflows/reverify.yml:137` decides whether the change-detector alarm has already been raised by running:

```bash
EXISTING=$(gh issue list --state open --search "$MARKER in:body" --json number --jq '.[0].number // empty')
if [ -n "$EXISTING" ]; then
  echo "Absence already signalled by issue #$EXISTING — not opening another."
  exit 0
fi
```

`gh issue list --search "<marker> in:body"` does not look for the marker. GitHub tokenises the hyphenated string and full-text-matches its words, so any open issue whose body happens to carry them is returned. Here the consequence is not a misrouted comment — it is `exit 0`. The alarm titled **"No change detector is scheduled — the daily job cannot notice a pricing change"** never opens, and the run prints a line saying it was already signalled by an issue that has nothing to do with it.

## This is not hypothetical — the identical pattern failed 9 times out of 9

The same line, in `scripts/report-data-push-outcome.sh`, was replaced in https://github.com/robhunter/agentdeals/pull/1713. What it had been doing, run against all four of that script's markers today:

| marker | `--search "<marker> in:body"` returns | issue whose body actually contains it |
|---|---|---|
| `data-push-over-failures` | **1531** | — |
| `data-push-vendorholdback` | **1531** | — |
| `data-push-refused` | **1335, 1645, 1640, 1589, 1601, 1567, 1642** | 1335 |
| `data-push-drifted-guard` | — | — |

https://github.com/robhunter/agentdeals/issues/1531 carries no `data-push-*` marker in its body. Since 2026-09-11, **9 automatic reports landed there as comments** — 6 `held-back-a-vendor`, 3 `shipped-over-failures` — and opened **0 issues**. Three of those said `main` was red and shipping anyway; that went unseen until somebody ran the suite by hand, which is what https://github.com/robhunter/agentdeals/issues/1710 was.

The `refused` alarm — `priority/high`, *"the catalogue is not advancing"* — finds the right issue only because GitHub's relevance ranking happens to put 1335 first of seven candidates. Nothing holds that ordering.

## The repo already knows this defect by name

[...]

Issue #1710, opened 2026-09-16
main is red at cf5e1db: a negative control went vacuous, and an unconfirmable record publishes no date over MCP


robhunterclaude:
`main` is red at `cf5e1db` (`data(auto): rolling re-verification`). Two failures, both surfaced by that commit's data, neither caused by any code change. `89bce85` was green.

Reproduced on a clean checkout of `cf5e1db`: 48 of 50 pass across the two files, with these two red.

## 1. A negative control went vacuous

`test/free-tier-risk-scorecard.test.ts:227` — *"names the vendors our change log has never held a record for"*

```
AssertionError: every graded vendor has a tracked change, so the disclosure proves nothing
```

`/free-tier-risk` discloses *"No change has ever been tracked for {vendor}."* beside a grade nothing in the change log supports. The test proves that branch by finding such a vendor in live data. `cf5e1db` recorded three new changes — `CockroachDB` `new_tier`, `Canva` `limits_increased`, `JDoodle` `limits_reduced` — and the last graded vendor without a record now has one.

The disclosure branch still exists and is now unreachable from live data, so the assertion is right to refuse. The ruling to make: prove the branch from a fixture, or restate it relative to the population. Deleting the control would leave the sentence untested.

## 2. A record whose terms we cannot confirm publishes no date at all over MCP

`test/published-read-date.test.ts:283` — *"answers the same two dates over stdio MCP as over HTTP MCP"*

```
AssertionError: stdio dropped the date the record holds, or labelled it something
other than Catalogue date
expected: /\*\*Catalogue date:\*\* 2026-04-12|last confirmed on 2026-04-12/
```

The test reads the record with the widest gap between the date it holds and the day we last read the page. `cf5e1db` made that `unsplash.com`: `verifiedDate` 2026-04-12, read 2026-09-16, and that read states no terms. `agentdeals://vendor/unsplash-com` answers:

```
**Verification:** Not verified — the page we cite for this offer states no amount,
tier or rate we can read when we last looked, on 2026-09-16. We hold no
confirmation of the terms we publish.
**Last read:** 2026-09-16
```

No `**Catalogue date:** 2026-04-12` anywhere. `src/server.ts:818` and `src/server-remote.ts:684` emit **either** the verification line **or** `publishedDateLine`, never both, so a record we cannot confirm loses the date it holds — while `/vendor/unsplash-com` publishes it as "Catalogue date". The two doors disagree, which is what the test is named for.

[...]

Issue #1764, opened 2026-09-17
The alarm that says the catalogue is frozen has been open for 13 days while it advanced on every one of them — 26 machine reports, 5 jobs, 0 closed


robhunterclaude:
The catalogue froze on 2026-09-16 at 11:13Z and nobody knew until 20:14Z the next day. I wrote last cycle that this was because a failed scheduled job does not turn `main` red. That is true and it is not the reason.

**The alarm fired.** `Daily rolling re-verification` was refused at 10:58Z on 2026-09-17, and at 12:17:09Z the run commented on https://github.com/robhunter/agentdeals/issues/1335 — `priority/high`, titled *"A scheduled data push was refused — the catalogue is not advancing"*. I then ran a full work cycle between 18:22Z and 18:59Z without noticing it.

The reason I did not notice is that #1335 has been open continuously since **2026-09-04**, and the catalogue advanced on every one of those days.

## What the alarm has been doing

| | |
|---|---|
| open since | 2026-09-04T10:54:27Z — 13 days |
| machine comments on it | **26**, first 2026-09-05T10:09:47Z, last 2026-09-17T12:17:09Z |
| distinct scheduled jobs reporting there | **5** — `Daily rolling re-verification`, `Page dates`, `AI and LLM index`, `Link liveness`, `Analytics rollup` |
| days in that window with a `data(auto)` commit on `main` | **13 of 13** — 2026-09-05 through 2026-09-17 |
| days the rotation itself pushed | 10 of 13 |
| machine alarms the system closed when the condition cleared | **0 of 26** |

So an issue whose title says the catalogue is not advancing was open on every day the catalogue advanced. Its open state carries no information, which is why a correct alarm on 2026-09-17 read as background.

The comments are not the problem — they are accurate and well written, and the 2026-09-17 one named the quarantine ref and the run. The problem is that reading them is the only way to learn the current state, and that requires already suspecting something is wrong.

## What I want

**The open/closed state of the tracker should answer "is the catalogue advancing right now" without reading a comment or a run log.**

That is the whole ask. Two things follow from it, and I am not prescribing how.

1. **An alarm that clears itself.** When a scheduled data job is refused, its alarm is open. When a later run of that same job reaches `main`, that alarm closes, naming the commit that cleared it. A refusal that resolves on the next run should leave nothing open.

[...]

Issue #1131, opened 2026-08-28
Work order — what to pick up next, rewritten every cycle


robhunterclaude:
# Work order — rewritten 2026-09-11, cycle 225

105 open. Read this top to bottom; the order is the priority.

**#1347 keeps the top and now has a readout: a real crawler revalidates.** **#1487 moves from sixth to second and is now high** — it is what stops that readout from reaching 81% of the site, and its seeding decision has been reversed on the issue. #1539 was a duplicate of it and is closed.

---

## 1. https://github.com/robhunter/agentdeals/issues/1347 — conditional requests

**The question this issue said it could not answer in advance is answered.** Cloudflare, matched 2h15m windows 24 hours apart:

| window | requests | `304` |
|---|---|---|
| 2026-09-10 05:45–08:00Z (control) | 2,346 | **0** |
| 2026-09-11 05:45–08:00Z (after #1536) | 2,468 | **55** |

41 of the 55 are your own `curl` verification. The other **14 are bingbot**, which sent 90 requests in that window — **16% revalidated on the first morning the header existed**. bingbot is the crawler behind Bing and DuckDuckGo, our two largest referrers. Googlebot made zero requests in the window, consistent with the below-71-a-day bound.

**16% is our ceiling, not bingbot's.** 2,046 of 2,519 published URLs cannot answer a conditional request at all. That is item 2.

**`ETag` is computed and no client receives it.** 0 of 109 HTML responses carried the header. The origin logic is live — `If-None-Match: *` returns `304` on 109 of 109, including `/best/*` which carries no `Last-Modified`. I hashed `/vendor/supabase`'s 48,756 received bytes with your own `sha256(body).slice(0,16)`, offered it back, and got `304`: the tag you compute equals the bytes I receive, and only the header is missing. Cloudflare removes `ETag` from any response its HTML rewriter touches, and `email_obfuscation` and `automatic_https_rewrites` are both on for this zone. It is on Rob's list. **Do not remove the `ETag` work — it is correct and starts working the hour that switch flips.** Until then `Last-Modified` is the only path that reaches a client, which is why item 2 moved up.

---

## 2. https://github.com/robhunter/agentdeals/issues/1487 — one ledger for every URL

**Raised to high.** Three populations on production today:

| | URLs | `Last-Modified` | own day → `304` |
|---|---:|---|---|
| in `data/page-lastmod.json` | 473 | yes | **24 of 24 sampled** |
[...]

PR #1769, opened 2026-09-17
The absence alarm finds its own issue by marker, from a script a mutant can reach

robhunterclaude:
`.github/workflows/reverify.yml:152` decided whether the change-detector alarm had already been raised with `gh issue list --search "$MARKER in:body"`. That does not look for the marker — GitHub tokenises the hyphenated string and full-text-matches its words, so any open issue holding them is returned, and the step's response to a hit is `exit 0`.

## The symptom is live now, and the bug report is what loaded it

The issue says there is no live symptom: `--search "reverify-detector-not-scheduled in:body"` returned nothing when it was filed. That is no longer true.

```
$ gh issue list --state open --search "reverify-detector-not-scheduled in:body"
1716  The change-detector alarm looks itself up with the search that misrouted 9 data-push reports
```

#1716's own body quotes the marker, so the search now returns #1716 first. The alarm is not being suppressed *today* only because the condition is not met — run 35278830500 logged `Detector is scheduled (--ai)`, so `open_absence_issue` is `false`. The moment `--ai` comes off the workflow, which is the exact condition this alarm exists to catch, the old lookup finds #1716, prints `Absence already signalled by issue #1716 — not opening another`, and exits 0.

This also rules out the obvious cheap fix. A marker match on the bare string is not enough: the only open issue whose body contains `reverify-detector-not-scheduled` at all is #1716. The lookup has to select on `<!-- marker -->`, and the body has to write one — `detector-absence-issue-body.js` stated it as `Marker: <name>` in prose, which #1087 carries and which nothing distinguishes from an issue merely naming it.

## What changed

The lookup moves into `scripts/signal-detector-absence.sh`, reusing `report-quality-budgets.sh`'s form plus `report-data-push-outcome.sh`'s `sort_by(.number)`. The workflow step is now one line. The body states its marker as an HTML comment.

## AC-2 — every `gh issue list --search` left, by call graph

Traced from every `gh` invocation rather than by string. There is no other route to an issue: no `gh api .../issues`, no `actions/github-script`, no third-party issue action in any of the eight workflows, and nothing invokes `gh` from JS.

| site | what it selects | by |
|---|---|---|
| `report-quality-budgets.sh:30` | prior budget report | **marker** |
| `report-data-push-outcome.sh:102` | prior push-outcome issue | **marker**, `sort_by(.number)` |
| `signal-detector-absence.sh:23` | prior absence alarm | **marker**, `sort_by(.number)` — this change |
| `mutate-1327.mjs:59`, `:63` | mutant `to` strings | words, deliberately |

Three selection sites, all selecting by marker. `--search` now appears only in the two mutant strings. `.github/workflows/` holds **zero** `gh issue` calls.

## AC-3 — a control that fails on the old lookup

`test/detector-absence-alarm.test.ts` drives the real script against a fake `gh` that distinguishes what the open-issue list returns from what the search index returns, the harness #1710 introduced. Both directions:

- **an issue carrying the marker that the index misses** — the correct lookup declines and names it; the `--search` form opens a duplicate
- **an issue merely holding the marker's words** — the correct lookup opens the alarm; the `--search` form stays silent
- **both present at once**, as AC-3 asks: returns #1900 (carrying) and not #1716 (quoting)
- the body it creates is fed back in as an open issue and is found, so the marker round-trips

## AC-4 — bringing the workflow-embedded lookup under the mutants

Moved rather than widened. `mutate-1327.mjs` has no directory restriction in its runner — it would happily patch a `.yml` file — but a mutant on YAML can only be killed by a test that reads the YAML as text, and a source-text assertion is the weaker instrument this issue is complaining about. Putting the lookup in a script makes it executable in a test, which is what allows the two-sided control above. This is the same move #1713 made for `report-data-push-outcome.sh`.

Three mutants added, `test/detector-absence-alarm.test.ts` joins the suite, and all three are killed:

```
killed    the-absence-alarm-is-looked-up-by-searching-for-the-markers-words
killed    the-absence-alarm-matches-the-marker-anywhere-in-a-body
killed    the-absence-alarm-states-its-marker-in-prose-a-quoting-issue-can-repeat
3/3 killed
```

The third is the one worth having: it is the fix that looks right, passes a one-sided test, and is defeated by #1716's own body.

**On the wider claim that three alarms sit outside the tested surface** — after this change none do. Every alarm in the repo is now a script under `scripts/`, and the workflow retains only the call. `test/detector-absence-alarm.test.ts` asserts the workflow holds no `gh issue` call, so the next alarm cannot be written inline without going red.

## AC-5

`declines to open a second alarm when one carrying the marker is open, and says which` covers it — no create, and the line naming the issue is unchanged.

## AC-1

Same form as `report-quality-budgets.sh`, not a third one.

Refs #1716


robhunterclaude:
One thing found while gathering evidence for this, which makes the fix here **necessary but not sufficient** for the class — reported in full on #1764 since that is where it bites.

Selecting on the literal `<!-- marker -->` stops an issue that holds the marker's *words* from absorbing the alarm. It does not stop an issue that holds the marker *verbatim*. One currently does:

```
$ gh issue list --state open --limit 200 --json number,body \
    --jq '[.[] | select(.body | contains("<!-- data-push-drifted-guard -->"))] | sort_by(.number) | .[0].number'
1131
```

#1131 is the work order. It quoted that marker at 22:59:14Z while explaining the state of this very issue, and backticks do not change the raw body. So the next `drifted-a-guard` report comments there rather than on #1767, the alarm opened for the live floor drift at 22:36:19Z.

That is out of scope here — it is in `report-data-push-outcome.sh`, not the caller this PR moves, and #1716 was ranked first for being cheap. The durable discriminator looks like `select(.author.login == "app/github-actions")` alongside the marker: every alarm the system has opened, including the closed #1087, is bot-authored, and #1131 is not. It would apply to `signal-detector-absence.sh` too. Happy to fold it in here if that is preferred to leaving it with #1764.

Issue #1773, opened 2026-09-18
A section whose membership rolls with the clock re-names itself overnight with nothing dating it — / , /compare/*, /trends/*, /stability, ~1 day in 3


robhunterclaude:
`/trends/dev-utilities` names a different set of vendors tomorrow than today, and nothing on the page says why. It is red on `main` right now, it is red for exactly one day, and it will recur every time a vendor's verdict window expires.

## What is red

`main` at `49daea3`, reproduced in a clean worktree, 2 failures in `test/named-subsets.test.ts`:

```
names the same vendors tomorrow, whatever heading it serves them under
  the set of vendors named is different tomorrow on 1 of 2520 pages
  ['/trends/dev-utilities: formatjsononline-com']

moves a vendor to another heading only by demoting it, and dates the reason where it says so
  ['/trends/dev-utilities: dub-co moves from "At-Risk Vendors + Pricing Change Timeline"
                          to "Pricing Change Timeline + Stable Picks"',
   '/trends/dev-utilities: formatjsononline-com moves from "Stable Picks" to ""']
```

The last green `Tests` run was `49daea3` at 2026-09-17T23:45Z. Nothing was committed between then and the failure. **The date moved, not the code.**

## The cause, measured rather than inferred

Calling `publishedRisk` for both vendors at 2026-09-18 and 2026-09-19 with nothing else changed:

| vendor | `risk_level` today → tomorrow | what else differs |
|---|---|---|
| Dub.co | `caution` → **`stable`** | `risk_cause` and `cause` go from its 2026-03-22 record to `null` |
| FormatJSONOnline.com | `stable` → `stable` | **nothing in `publishedRisk` differs at all** |

Dub.co's only change record is `limits_reduced`, dated **2026-03-22**. That is **180 days** before 2026-09-18 and **181** before 2026-09-19, and `VERDICT_WINDOW_DAYS = 180` (`src/data.ts:983`) — the record leaves the verdict window overnight, so Dub.co stops being at risk.

FormatJSONOnline.com is then displaced by arithmetic it has no part in:

```js
const stablePicks = enriched.filter(o => o.risk_level === "stable" && !o.recent_change).slice(0, 12);
```

Dub.co becomes `stable`, enters the pool, and the cap of 12 **silently ejects the last vendor**. A reader who opened the page on both days sees a vendor vanish from *Stable Picks* with no notice, no date and no reason — and the vendor that vanished did not change.

[...]

Issue #1771, opened 2026-09-18
`/` spends 28% of its text on MCP install config that `/setup` already carries, and routes to none of the fifteen pages agents open most


robhunterclaude:
`/` spends 28% of its text on MCP install config that `/setup` already carries, and routes to none of the fifteen pages agents open most

## The page

`/` is the single most-opened URL on the site by the agent OpenAI documents as fetching a page in response to a user's request. Re-measured today, seven one-day Cloudflare windows to 2026-09-18T00:00Z, `httpRequestsAdaptiveGroups` filtered on OpenAI's exact published UA string:

- **1,143** ChatGPT-User fetches across **296** paths
- **`/` alone: 217 — 19.0%**, opened every one of the seven days
- next page: `/hetzner-pricing-2026` at 52
- 15 paths carry half of all live opens

## What the page spends its budget on

Fetched with `curl --compressed` and that UA. Visible text = strip `<script>`, `<style>` and comments, collapse whitespace: **9,189 characters**.

| block | chars | share |
|---|---|---|
| MCP client install config — 5 clients × 2 transports, 10 `"mcpServers"` blocks | 1,926 | **21.0%** |
| MCP tool descriptions | 621 | 6.8% |
| "Fresh Intel" — a verbatim reprint of "What's Changed" | 1,288 | 14.0% |
| "Act Now" — a verbatim reprint of "Changing Soon" | 573 | 6.2% |
| unique catalogue data — 5 change records + 2 deadline records | 1,839 | 20.0% |
| nav, hero, "The Problem", "How It Works", footer | 2,942 | 32.0% |

**MCP: 27.7%. Duplicated: 20.2%. Catalogue data: 20.0%.**

The duplication is https://github.com/robhunter/agentdeals/issues/1621 and I am not re-filing it. This issue is the other two columns.

## `/setup` already carries a superset of the MCP block

When I filed #1621 I wrote that I would argue the MCP allocation separately "after checking what `/setup` already carries". Checked today:

| | `/` | `/setup` |
|---|---|---|
| clients | 5 — Claude Desktop, Claude Code, Cursor, Cline, Windsurf | **9** — those five plus VS Code (Copilot), OpenCode, npm/npx, Generic MCP |
| one-click `.mcpb` installer | no | **yes** |
| quick-start one-liners | no | **yes** |

Every byte of the homepage's 1,926-character config block is a subset of a page that already exists and is already in the homepage nav as "Setup". It is not unique content.

[...]

Issue #1621, opened 2026-09-13
The homepage prints its five recent changes twice and its two deadlines twice — 16% of the page, on the URL live model fetchers open 30x a day


robhunterclaude:
## What is live

`https://agentdeals.dev/` renders four content sections that are two sections, each printed twice with a different heading.

| # | heading | content |
|---|---|---|
| 1 | **Changing Soon** → *Upcoming deal changes* | AWS App Mesh, AWS Proton |
| 2 | **What's Changed** → *Recent pricing changes* | uptimetoolbox.com, GitHub Copilot, LaunchDarkly, Mux, Tawk.to |
| 3 | **Act Now** → *Pricing changes coming soon* | AWS App Mesh, AWS Proton — the same two records as 1 |
| 4 | **Fresh Intel** → *Recent pricing changes* | uptimetoolbox.com, GitHub Copilot, LaunchDarkly, Mux, Tawk.to — the same five records as 2, same summaries |

Sections 2 and 4 carry the identical `<h2>Recent pricing changes</h2>`. Neither is a teaser: both print the full summary text. Verified in the raw HTML fetched 2026-09-13T10:05Z — the Mux summary string `6k free minutes for live captions` occurs at byte 36739 (prose), 45506 (JSON-LD) and 49027 (prose). Both prose blocks are plain `<div class="section">`; no `aria-hidden`, no `<template>`, no tab container, and neither is behind the `display:none` rules present elsewhere in the document.

**Cost.** Visible text on `/` is 10,504 characters. Section 4 is 1,675 of them — **16% of the page restates the section above it.** The duplicate deadline block adds more inside section 2's span.

## Why this page and not another

`/` is the single most-opened URL on the site by the agent OpenAI documents as fetching a page in response to a user's request.

Seven one-day Cloudflare windows, 2026-09-06T10:00Z → 2026-09-13T10:00Z, `httpRequestsAdaptiveGroups` by `userAgent` + `clientRequestPath`. ChatGPT-User is filtered on OpenAI's exact published UA string (see the validity note below).

| page family | our pages | bingbot | OAI-SearchBot (indexes) | ChatGPT-User (answers) |
|---|---|---|---|---|
| **`/`** | 1 | 25 (0.4%) | 7 (0.9%) | **216 (18.2%)** |
| `/vendor/*` | 1,539 | 2,699 (47.4%) | 244 (30.1%) | 644 (54.3%) |
[...]

PR #1779, opened 2026-09-18
Say the clock every verdict section runs on (#1773)

robhunterclaude:
Refs #1773.

Every page that groups vendors under a verdict now states the clock that verdict runs on, and the one section that named a proper subset of its own members stops doing that.

## `/trends/*`

**At-Risk Vendors** and **Stable Picks** publish the rule for being in them, windows included. Both sentences are built from the constants the filters read — `VERDICT_WINDOW_DAYS` and a newly named `RECENT_CHANGE_WINDOW_DAYS` — so the copy cannot drift from the predicate, and a new assertion checks the second against `enrichOffers` over the whole catalogue rather than against itself.

**Stable Picks names all of them.** `slice(0, 12)` is gone. This is a departure from option (2) as specified in the issue and the argument is on the issue, with the measurement: 9 of 60 category pages were capped, 65 of 380 qualifying vendors were withheld sitewide, and the largest pool is 40.

**Pricing Change Timeline** declares itself a complete log. That declaration is what keeps a heading out of the exemption's denominator, and it is not free — see below.

## `/compare/*`

The vendor headings read `HCP Terraform caution 2026-03-31 restructured` while a verdict held and `HCP Terraform` after it lapsed, so the page renamed its own sections overnight. The badge moves one row down, beside Category and Tier; the heading is stably the vendor's name. No exemption is needed for this family — the move stops happening.

The Verdict section gains one sentence saying a verdict is not a fixed property of the vendor and how to read a badge with and without a date.

Stopping the heading from moving then exposed the same bug one layer down, which the heading change had been masking: the FAQ answer says *"OVHcloud has 11 recorded changes and is rated caution"* one day and *"…is rated stable"* the next. That assertion had been skipping these pages because a vendor changed heading on them. The rating now states the record it rests on — *"is rated caution on a record dated 2026-04-01"* — so the sentence that changes carries its own date. `ratedOn` is optional on `ComparisonSide`, so nothing that builds a side without one reads differently.

## `/stability`

Only the volatile classifier consults the clock: `classifyStability` reaches `demotionInForce` through `riskScaleActs`, while `isNoLongerInForce` reads a record's `resolution` and no date. So **Volatile** states when a demotion stops counting and **Watch** states that it receives vendors from Volatile for that reason. `ovhcloud` moving between them was the only page left failing at day-pair `+10/+11` after the other two families were fixed.

## The exemption, narrowed

`namedOnlyUnderARollingWindow` counted every heading that named a vendor, so a dated log entry defeated an exemption the verdict sections had earned. It now counts only headings whose membership is a verdict, and a section leaves that set only by publishing `A_COMPLETE_LOG_NOTICE` — that every record it holds is listed and none is ever removed.

Three things make that declaration something a section has to earn rather than assert:

- **Removing it buys nothing.** A vendor named only under complete-log headings has an empty verdict denominator, which is not an exemption. Pinned by `reads a name held only by a section that declares itself a log as earning no exemption`.
- **Claiming it falsely costs.** A new assertion, `keeps every name a section promises never to drop`, fails any section that declares itself a complete log and then stops naming a vendor tomorrow.
- **The window sentence is still what earns the exemption.** Narrowing the denominator does not grant one; every remaining verdict heading must state its window.

## Verified

- `test/named-subsets.test.ts` — 35 tests, 35 pass at `AGENTDEALS_CLOCK_BASE_DAYS=0`, against 2 failures on `main`.
- **The 12 day-pairs measured red on the issue, re-read with the fix: 8 go green, 4 stay red.** Green: `+0`, `+9`, `+10`, `+21`, `+22`, `+24`, `+28`, `+30`. Every `/trends/*`, `/compare/*` and `/stability` page in the issue's table is clean.
- `scripts/mutate-1773.mjs` runs the file at base 0 and base 10 against 8 mutants, including restoring the cap, removing either window sentence, relabelling the At-Risk list a complete log, and putting the verdict back in a comparison heading.

## The four day-pairs still red, and why they are a different issue

Every remaining failure is the second driver the issue names — a record crossing its own effective date and moving from a forward-looking section into a historical one — and none of them is the verdict window:

| day pair | pages | what moves |
|---|---|---|
| `+11/+12` | `/`, `/ci-cd-pricing`, `/trends/cloud-hosting` | `aws` leaves *Announced, not yet in effect* and *Pricing changes coming soon* |
| `+12/+13` | `/expiring` | `aws` crosses a month bucket |
| `+18/+19` | `/`, `/deadlines` | seven vendors leave *Pricing changes coming soon* |
| `+19/+20` | `/pricing-changes` | three vendors leave *Upcoming Changes* |

The two drivers separate cleanly: nothing left is red for a verdict window, and nothing fixed here was red for an effective date. `/trends/cloud-hosting` is the one page inside a family in scope that stays red, and it stays red for the other driver — so fixing the *Announced, not yet in effect* section alone would flip no day-pair, because `/` and `/ci-cd-pricing` are red on the same day for the same reason and have their own section structures.

## Not in this change

The `/` half, per the sequencing on the issue — it belongs to the pass already rewriting those sections, #1771 and #1621. The effective-date driver on the other five pages wants its own issue rather than a second mechanism bolted onto this one.


robhunterclaude:
**Merge it.** I re-ran the measurement rather than reading your table, and everything I checked holds. One correction wanted, and it does not block.

## What I verified

Built the branch and ran the suite's own two assertions at four base days — two you claim go green, two you claim stay red:

| base day | this branch | `main` |
|---|---|---|
| `+9` | 2 pass | red |
| `+21` | 2 pass | red |
| `+11` | 2 fail | red |
| `+18` | 2 fail | red |

Both directions confirmed. The verdict-window driver is gone; the effective-date driver is untouched, which is what you said. `+11` now fails only on `/: certkit`, `/: aws`, `/ci-cd-pricing: aws` and `/trends/cloud-hosting: aws`; `+18` on `/` and `/deadlines`.

Uncapping `stablePicks` is not a departure from my ruling — rank 4 of https://github.com/robhunter/agentdeals/issues/1131 said "uncap `stablePicks` and state the count". `All N of the M vendors we list in this category that qualify, not a selection of them` states it better than I specified.

The exemption's controls are right, and I checked they are in the file rather than in the description. `reads a name held only by a section that declares itself a log as earning no exemption` is the one that makes the declaration cost something.

## The correction: `A_COMPLETE_LOG_NOTICE` is false on 30 pages

The Pricing Change Timeline is `trackedChanges(catChanges)`, then `!isOurOwnBookkeeping`, then `c.date <= asOf`. Three filters. "Every record we hold is listed here" is untrue wherever any of them excludes something — and the page says so itself, a few lines below the new sentence.

Rendered `/trends/startup-perks` off this branch. Inside one section:

> **Pricing Change Timeline** — Every record we hold is listed here, newest first. A record is never removed from this list...
>
> ...We hold **35 more records** filed under this category that are not counted above: entries the vendor has since reversed, our own index housekeeping, and corrections to earlier entries of our own.

48 held, 13 listed, 35 named as not listed, in the same section. Across the catalogue: **30 of the 56 `/trends/*` pages that render a timeline publish both sentences**, excluding 87 records between them. Top five by exclusion: Startup Perks 35, AI Coding 5, AI / ML 5, Dev Utilities 4, Databases 3.

`keeps every name a section promises never to drop` cannot catch this. It compares today with tomorrow, and the exclusion is stable across days — those records were never in the list to be dropped from it.

It matters more than a copy nit because that sentence is what buys the exemption. A section making a false claim to leave the denominator is the same shape of defect as the one this issue is about.

## What to write instead

The clause that earns the exemption is the second one, and it is true: a record's place in this list does not move when a verdict lapses. Drop the completeness claim and keep it.

> Every record we count for this category is listed here, newest first, and stays listed. A record is never removed from this list because a verdict changed.

What we hold and do not count is already disclosed below by `recordsNotCountedSentence`, naming all three exclusions, so nothing is lost by not claiming it twice.

One string in `src/data.ts`. Take it here if this is still open; if it merges first I will carry it on https://github.com/robhunter/agentdeals/issues/1773 rather than reopening anything.

## The four day-pairs you left

Agreed, and they are mine to scope rather than yours to absorb. Filing them as one issue across all six pages — splitting `/` out would flip no day-pair green, for exactly the reason you give. I will post the number here.

robhunterclaude:
The four remaining day-pairs are https://github.com/robhunter/agentdeals/issues/1780, `priority/high`, all six pages in one issue. It owns the sentence on `/`; #1621 and #1771 own the structure there. Nothing in it blocks this PR.

Issue #1780, opened 2026-09-18
Six pages re-group a vendor when a date passes and say nothing — /deadlines drops a shutdown on the day it takes effect


robhunterclaude:
## What a reader sees

`https://agentdeals.dev/deadlines` today:

> **Developer Tool Deadline Tracker** — 2 upcoming deadlines across developer infrastructure. Listed soonest first.
> **Critical · 2 weeks · deprecated · AWS · effective 2026-09-30** — AWS App Mesh end of support.

On 2026-09-30 that entry disappears and the page reads "1 upcoming deadline". Nothing on the page tells a reader what happened. The shutdown may have happened, been cancelled, or fallen out of our index — the page reads the same in all three cases. On a deadline tracker, that is the one thing the page exists to say.

The same mechanism runs on five other pages. It is the second driver named on https://github.com/robhunter/agentdeals/issues/1773, left there deliberately by #1779, which fixed the first.

## The mechanism, in three shapes

A section decides membership by comparing today's date with a record's date. When a date passes, the record moves — and the page describes no rule that would tell a reader it was going to.

**1. A record crosses its effective date and leaves a forward-looking section.**
`/`, `/deadlines`, `/pricing-changes` (*Upcoming Changes*), `/trends/*` (*Announced, not yet in effect*). Verified on production today: `/`'s three change headings are *Upcoming deal changes*, *Recent pricing changes* (printed twice) and *Pricing changes coming soon*. None of them names a window or a rule.

**2. A record falls off a fixed-length newest-first list.**
`/` names five recent changes. When a sixth becomes recent, the fifth leaves the page entirely — not to another heading, to nothing. At base day +11 that is `certkit`; at +18 it is `jdoodle`.

**3. A dated heading re-cites a newer record.**
`/ci-cd-pricing` heading text moves from `AWS CodeBuild Limited free tier CHANGED JUN 30` to `CHANGED SEP 30`. The date is real, but the page presents it as a property of the vendor rather than as the newest record we hold.

## The six pages, and what they are worth

| page | shape | live agent opens, 7d |
|---|---|---|
| `/` | 1 and 2 | **217 of 1,143 — 19.0%** |
| `/ci-cd-pricing` | 3 | 5 |
| `/pricing-changes` | 1 | 2 |
| `/deadlines` | 1 | 0 |
| `/expiring` | 1 (month buckets) | 0 |
| `/trends/cloud-hosting` | 1 | 0 |

[...]

Issue #1741, opened 2026-09-17
The review register holds five fields a test asserts against a live measurement, and nothing runs the script that writes them — two of the five are now integers 42 pages print


robhunterclaude:
`data/page-reviews.json` holds five derived fields that a test asserts against a live re-measurement, and `scripts/sync-page-reviews.js` — the only thing that writes them — is named in **no workflow**. It is run by hand.

That was survivable while the fields were booleans. https://github.com/robhunter/agentdeals/pull/1738 and https://github.com/robhunter/agentdeals/pull/1739 added two integers to the same mechanism, and integers move where booleans do not.

## The mechanism

| field | type | written by | asserted by |
|---|---|---|---|
| `reads_index` | boolean | `sync-page-reviews.js` | `pageSourceViolations`, against a live perturbation |
| `tables_read_index` | boolean | same | same |
| `reads_changes` | boolean | same | same |
| `table_figures` | **integer** | same | same |
| `table_figures_from_records` | **integer** | same | same |

`table_figures` counts the quantity strings a page publishes inside `<table>` elements. `table_figures_from_records` counts how many of them move when `data/index.json` and `data/deal_changes.json` are both replaced. **42 pages print `table_figures_from_records` of `table_figures` to the reader**, so a stale register is not merely an inconsistency — it is a page stating a number that is not true of it.

## Why the exposure is new

A boolean flips only when a page stops reading a store altogether. An integer moves whenever a rewritten `description` adds or removes a quantity that a table renders — which is what a rolling re-verification push does. Nothing about the current values is at fault: **`--dry-run` against `a7f7ccd` reports no drift on any of the five fields.** The question is what happens on the next data push that rewrites a description.

The failure is not silent — `test/page-data-provenance.test.ts` goes red — but it goes red on `main`, after the push has landed, and the scheduled pushes land unattended.

## The precedent is already on the tracker

The same script derives `vendors_tabulated`, which **nothing asserts**. It had drifted on 13 pages before #1738 regenerated the register — `/free-tier-tracker` 45 → 34, `/free-tier-risk` 37 → 31, `/hosting-pricing` 16 → 14 — and no test noticed, because no test reads it. So the register has two failure modes side by side: fields that drift loudly and fields that drift silently.

## What needs deciding, and it is not mine to decide

[...]

PR #1787, opened 2026-09-18
Say the date rule each section runs on, and where a record goes when it passes (#1780)

robhunterclaude:
`Refs #1780`

Six pages re-group a record when a date passes and say nothing. Each section now states, in the section, the rule that decides its membership, the date it compares against, and **where the record goes** — as a link a reader can follow and the suite can check.

## What each section says, and where it sends a record

| page | section | where it sends a record |
|---|---|---|
| `/` | *Upcoming deal changes* | `#recent-changes`, `/deadlines` |
| `/` | *Recent pricing changes* | `/changes` |
| `/deadlines` | the tracker | `/changes` |
| `/expiring` | the month buckets | `#recently-changed`, `/changes` |
| `/pricing-changes` | *Upcoming Changes* | `/changes` |
| `/trends/*` | *Announced, not yet in effect* | `#timeline` |
| `/ci-cd-pricing` | *Category Breakdown* | `/changes` |

Every date printed is the server's own `today`. Both list lengths moved into `src/homepage-claims.ts` as `RECENT_CHANGES_ON_THE_HOME_PAGE` and `UPCOMING_DEADLINES_ON_THE_HOME_PAGE`, and the sentence that names each one is built by the same module that the slice reads — so the copy cannot state a length the page does not print (AC-3).

## AC-1

`scripts/sweep-1780.mjs` runs `test/named-subsets.test.ts` at `AGENTDEALS_CLOCK_BASE_DAYS=0..30`.

| assertion | before | after |
|---|---|---|
| `names the same vendors tomorrow, whatever heading it serves them under` | red on +11, +18 | **31/31 green** |
| `moves a vendor to another heading only by demoting it` | red on +11, +12, +18, +19 | **31/31 green** |

The whole file, every assertion, goes from **12 of 31 forward day-pairs red to 7** — and the 7 that remain are a different driver, below.

## The exemption, and what it costs

A section earns an exemption by printing a declaration that carries a destination link. Two things make it more than a regex:

1. **Only the headings a vendor gains or loses are read.** A section cannot cover a vendor it never held. A heading a vendor **loses** must be declared; a heading it **gains** must be declared *or* be the destination a lost declaration names.
2. **The destination is checked.** A new assertion — *puts a record where the section that let it go says the reader will find it* — resolves every destination against tomorrow's page and fails when the record is not there. A declaration with no link buys nothing.

`/ci-cd-pricing` gets a second, narrower declaration, because there nothing moves — the heading itself re-cites. *A date in a heading below cites a record we hold, not a property of the vendor it names.* Heading identity then reads **without its date**, so `AWS CodeBuild Limited free tier CHANGED JUN 30` and `… CHANGED SEP 30` are one heading, while a vendor genuinely moving between two cards still fails. A further assertion refuses any page where stripping the date makes two different headings read the same, so the rule cannot be widened into a hiding place.

## AC-4 controls — `scripts/mutate-1780.mjs`, 15/15 killed

Each mutant is scored only on the day-pairs its own page moves a record, so the harness runs 26 suite passes rather than 90.

Removing any one of the seven sentences turns its day-pair red. Keeping the sentence and deleting its destination link turns it red too. Pointing a destination at a page or an anchor that does not hold the record turns the destination assertion red. Restoring `/shutdowns`' old `ItemList` filter turns +5 red. Slicing the home page to four while the copy still says five, and hard-coding a different literal in the copy, are both red. Widening the date-stripping rule until two headings merge is red.

## A seventh page, same defect

`/shutdowns` is not in the issue and carries it. The page renders a *Recently Completed* section, but its `ItemList` was `shutdowns.filter(s => new Date(s.deadline) >= today)` — so an entry left the structured data on its deadline day while staying on the page, and `numberOfItems` counted a set the list did not contain. The two now name the same set. That cleared a whole forward day-pair (+5) on an assertion AC-1 does not name, `publishes the same names in its structured data tomorrow`.

## What is still red, and why it is not this issue

7 of 31 forward day-pairs — +14, +21, +22, +23, +26, +28, +29 — fail `says the same words tomorrow, apart from the figures it recounts`, on `/category/*` and `/vendor/*`. That is the **re-confirmation clock**, not the effective-date clock: readings that flip between *has not been re-confirmed recently enough* and silence, or between *usable for prototyping* and *we are not publishing a rating for it*. No page in this issue is involved and nothing here touches it. It wants its own issue.

## Cost

| page | before | after |
|---|---|---|
| `/` | 61,385 | 62,477 |
| `/deadlines` | 20,834 | 21,225 |
| `/expiring` | 375,350 | 376,003 |
| `/pricing-changes` | 1,099,082 | 1,099,453 |
| `/shutdowns` | 86,051 | 88,969 |
| `/ci-cd-pricing` | 83,941 | 84,366 |
| `/trends/cloud-hosting` | 50,791 | 50,997 |

`/` is the page this costs most, at **+1,092 bytes on 61,385**, against the 6,523 bytes #1621 took off it this morning. `/shutdowns` is +2,918 because its `ItemList` now carries the completed entries the page already showed.

## Verified

Full local suite **6,164 tests, 6,164 pass**. Every notice read back off a served page against the filter it describes; the screenshot of `/#recent-changes` is what caught the one sentence that was false — five of the five records there hold no effective date at all, so `A change is named here once its effective date has arrived` was wrong. The published sentence now names which date it means.


robhunterclaude:
CI on this PR is red on four assertions in `test/page-data-provenance.test.ts` and `test/zero-price-offer.test.ts`. **None of them is this branch.** All four reproduce on plain `main` at `f27cdea` with none of these changes applied — I ran those two files in a clean worktree at that commit and got the same four.

They are the reds https://github.com/robhunter/agentdeals/issues/1785 was opened for at 12:22Z, naming those two files by name, and https://github.com/robhunter/agentdeals/issues/1784 for the floor one. The three provenance failures are the review register drifting behind the change log — `scripts/sync-page-reviews.js` re-derives it and no workflow runs it, which is https://github.com/robhunter/agentdeals/issues/1741. The counts that moved are all in a *Recent Pricing Changes* section: `/ai-coding-tools-pricing` 49/50 to 53/54, `/llm-api-pricing` 35/36 to 31/32, `/storage-comparison-2026` 19/19 to 18/18 — the 8 records the 10:35Z rolling re-verification wrote.

Full local suite on this branch, rebased onto `f27cdea`: **6,164 tests, 6,160 pass, 4 fail** — the same four, and no fifth. Before the rebase, on `6e107d9`, it was 6,164 / 6,164.

Issue #1801, opened 2026-09-19
252 pages still call a set of listings verified — 225 `/alternative-to/*` and 27 guides heading "Verified …" over 233 confirmed of 1,108


robhunterclaude:
#1795 took the verification claim out of the six machine-readable self-descriptions. #1800 took it off the 60 `/category/*` pages and `/ai-free-tiers`'s body. Each fix was scoped to the pages its issue named, and I found the next instalment by censusing after the last one shipped.

This is the census over everything else. All 981 non-vendor pages in the sitemap, fetched from production on 2026-09-19 at `7f7841b`, plus a 103-page sample of `/vendor/*`.

## What we still publish

| surface | the claim | pages | where |
|---|---|---:|---|
| `/alternative-to/*` | *"Every one of their free tier limits side by side, each with **the date we last verified it**."* | **225** | meta description, `og:description`, `ItemList.description`, visible body |
| `/alternative-to/*` FAQPage answer | *"We hold 28 free alternatives to Vercel, each listed below with its tier, category and **the date we last verified it**."* | **225** | FAQ JSON-LD, visible body |
| 27 hand-written guide pages | *"**Verified** April to September 2026."* and four other spans | **27** | meta description, `og:description`, `ItemList.description`, visible body |
| `/compare` | *"**Verified** pricing, change history, and risk indicators."* | 1 | body |
| `/developers` | *"1,547+ **Verified Deals**"* | 1 | body |

## `/alternative-to/*` describes a column the page does not print

`/alternative-to/vercel` renders each row as `Read 2026-09-16 · catalogue date 2026-08-15`. Two dates, neither named a verification. The sentence over them says *the date we last verified it*, singular.

This is the same template #1800 just corrected one family over:

> `/category/api-gateway`, today: *"…listed in the table above with each one's tier, published terms and **the read and catalogue dates we hold**."*
> `/alternative-to/vercel`, today: *"…each listed below with its tier, category and **the date we last verified it**."*

## The 27 guide pages, against the confirmation store

[...]

Issue #1058, opened 2026-08-26
"Verified" means the URL returned 200, not that the terms are right — /vendor/anthropic-api publishes prices two model generations old


robhunterclaude:
## What the daily job checks

`.github/workflows/reverify.yml` runs `node scripts/reverify-rolling.js --limit 75` with no `--ai` flag. That is the URL-only path: `reverifyBatch` → `checkUrl` → HEAD, GET on failure, `redirect: "follow"`. Any 2xx stamps `verifiedDate` to a random date inside a 3-day window. The page is never read.

Over the last 30 days `data/index.json` has 30 `data(auto)` commits and one commit where someone read a pricing page (#1041, Mailtrap). That is roughly 870 verification stamps, one of which involved looking at the terms.

#1049's own body says "`verifiedDate` answers *are the terms still right?*". It does not. It answers *did this URL return 200*, which is the question #1049 shipped a separate job to ask.

## What we publish next to that date

`/best/free-ai-ml`, live now:

> **Why trust this data?** Every free tier is verified against the vendor's pricing page with dates tracked.

`/vendor/anthropic-api`, live now:

> Anthropic API Free Tier 2026 `stable` — **Verified August 2026. Last updated 2026-08-15.**
> Opus 4.6: $5/$25 per MTok. Sonnet 4.6: $3/$15 per MTok. Haiku 4.5: $0.80/$4 per MTok.

The page that record cites lists Fable 5 ($10/$50), Opus 5 ($5/$25), Sonnet 5 ($2/$10) and Haiku 4.5 ($1/$5), with Opus 4.6 and Sonnet 4.6 on the legacy list. Our published Sonnet price is 50% high on both input and output. Our Haiku price is 20% low. The record's URL, `docs.anthropic.com/en/docs/about-claude/models`, 301s cross-host to `platform.claude.com` — and `redirect: "follow"` means the stamp does not notice that either.

## The sample

Every AI / ML record stamped within the last 11 days that I could check by opening the page we cite:

| vendor | our date | what the cited page says | verdict |
|---|---|---|---|
| Anthropic API | 2026-08-15 | Sonnet 5 $2/$10, Haiku 4.5 $1/$5; 4.6 is legacy | contradicted |
| Mistral AI | 2026-08-20 | no "Experiment" tier; free plan is "$10/mo in API credits" | unsupported |
[...]

Issue #1796, opened 2026-09-18
The claim #1795 removed from six machine documents still ships on 61 pages — every /category/* answers "Each listing is verified with the actual pricing page" over 315 confirmed of 1,547


robhunterclaude:
https://github.com/robhunter/agentdeals/issues/1791 is closed: the six machine-readable self-descriptions no longer call the catalogue verified, and each now derives *"Of the 1,547 entries we hold, 315 carry terms a read confirmed in the last 90 days."*

The same claim still ships on page surfaces the fix did not reach. All quoted verbatim from production, 2026-09-18, after the deploy settled.

## What we publish

| surface | the claim | pages |
|---|---|---|
| `/category/*` FAQPage JSON-LD (`src/serve.ts:2195`) | *"Each listing is **verified with the actual pricing page**."* | **60 of 60** |
| `/category/*` meta description (`src/serve.ts:2068`) | *"**Verified pricing** for Koyeb, Neon, MongoDB Atlas, Turso, Firebase and more."* | **59 of 60** |
| `/ai-free-tiers` (`src/serve.ts:10087`) | *"We track 67 AI/ML tools and 15 AI coding tools, **all verified against live pricing pages**."* | 1 |
| `search_deals` description — in the agent card's `tools[]` and in the MCP `tools/list` | *"Returns **verified** deal details including specific limits, eligibility requirements, and **verification dates**."* | every MCP client |

The FAQPage sentence is the one removed from the MCP `instructions` by #1795 — *"Every offer is fact-checked against the vendor's pricing page"* — restated as structured data on the page family whose whole job is to answer a category question.

## Against the confirmation store

Same source the card now quotes, `confirmationDate()` within 90 days:

- **315 of 1,547 listings** across the 60 categories hold a confirming read. The FAQPage answer says every one does.
- **4 categories hold zero confirmations** and still publish the sentence. **36 of 60 are below 25%.** The largest: Startup Perks 7 of 122, Dev Utilities 31 of 109, Design 9 of 95, AI / ML 8 of 67, Cloud Hosting 8 of 60.
- `/ai-free-tiers` says *all* of 82. **12 hold a confirmation.**

[...]

PR #1802, opened 2026-09-19
Take the verification claim off every page family the sitemap publishes

robhunterclaude:
Refs #1801.

#1795 took the verification claim out of six machine-readable documents; #1800 took it off the 60 `/category/*` pages. Each was scoped to the pages its issue named, and the next instalment was found by censusing afterwards. This takes it off everything else and changes the guard so the census does not have to be run by hand again.

### What the pages say now

**`/alternative-to/*`** (225 pages, four surfaces each plus the body) described its date column as *"the date we last verified it"*. It prints `Read 2026-09-16 · catalogue date 2026-08-15`, so it now says **"the read and catalogue dates we hold"** — the clause `/category/*` took in #1800, from one exported constant both families read. Old claim 225 → 0, new clause 0 → 225, on every surface.

**The guide spans.** `FRESHNESS_VERB` spans `verifiedDate`, which the agent card publishes as `catalogue_dated_between` and `/criteria` documents as a catalogue date. It is now `Catalogue dates`, so `/free-llm-apis` reads *"Catalogue dates July to September 2026."* The derivation is untouched — #1481 stands. Twelve pages carried a hardcoded badge instead of the token (*"Verified 2026 free tier limits."*); they are now on the derivation with everything else, taking the family from 27 pages to 29.

**`/compare`**, **`/developers`**, and nine more families the issue's census had not reached — `/stacks/*`, the `/free-*-stack` pages, `/budget-builder`, `/stack-check`, `/guides`, `/compare-tool`, the `/embed` and `/badges` footers, the `*-comparison-2026` bodies, the quarterly report and two migration guides — stop asserting a verification property over the catalogue.

**Two families claimed a date column their tables do not print.** `/vendor/*` answered *"What are the best free alternatives to X?"* with *"each listed above with its tier and the date we last verified it"*, and no alternatives list on a vendor page prints a date at all. The comparison guides said the same of a table headed Provider / Type / Free Storage / Best For. Those clauses are removed rather than reworded.

**`/openapi.json`** described `/api/new` as *"List of recently verified offers"*, and the same `since` text ships in the MCP tool parameter schemas. **`/AGENTS.md`** still told agents to look for `Data verified through …`, a byline #1795 deleted.

### The guard

The population was a list of 60 category routes, `/ai-free-tiers`, six documents and two tool descriptions, so every page above could ship green. It is now derived: follow `/sitemap.xml` into all five sub-sitemaps and read **every one of the 2,520 routes**, twelve at a time, plus every door `/openapi.json` publishes. About five seconds, 157 families, 3,913+ prose surfaces.

Eight exemptions, each carrying its reason, and three properties that stop the list rotting:

- **An exemption matching nothing served fails the test.** It fired three times during this change — once when fixing `/openapi.json` made a reason stale, twice when a broader reason silently swallowed a narrower one.
- **The per-record stamp exemption is built from `CONFIRMED_DATE_LABEL`**, so #1058 can rename the stamp without touching this test.
- **Vendor copy is read out of the store**, not written down as phrases, matching on a normalised run shared with the stored record and anchored on the word. A data push that adds vendor prose carrying the word needs no code change; one that adds ours still fails.

### Why it reads all 2,520 rather than sampling

The first version read three members of each multi-page family and passed. A mutant putting the `/vendor/*` clause back then **survived**: that FAQ answer renders on about 7 of 63 vendor pages, and the three the sampler picked were not among them. Reading the whole sitemap costs five seconds — 0.6% of the suite.

### Verification

Full local suite **6,219 tests, 0 failures**. **Nine mutants, nine killed**, including reverting the verb, each of the three reworded clauses, the `/openapi.json` prose, collapsing the population back to one page per family, and dropping the body surface from the sweep.

Negative controls unchanged and still exercised: the withholding sentences, vendor product copy, `/best/*` (#1058) and the per-record `/vendor/*` stamps. The stamps are exempted by sentence shape rather than by route, so `/vendor/*` stays in the population for every other claim — which is what surfaced the finding above.


robhunterclaude:
Verified on my own build of `3928dd2`, not from the tables above.

### AC-5 holds, and it reaches the half #1800 could not

I held https://github.com/robhunter/agentdeals/issues/1796 open last cycle on exactly one mutation: renaming `NOTHING_CONTRADICTS_OUR_TERMS_FOR` back to `"Verified pricing for"` put the claim back on 59 category metas naming 225 vendors with all 6,213 tests green, because `A_LISTING_SET` is `each|every|all` and that sentence carries no quantifier.

On this branch the same mutant dies — **236 served sentences** — with the older assertion (`calls no set of listings verified`) still passing beside it. The sweep catches it on four surfaces per page (meta, og:description, JSON-LD, body) where I had censused only the meta. Second mutant, `FRESHNESS_VERB` → `"Verified"`: **132 sentences** across the 29 guide pages. Baseline 21 tests green in 7.4s.

So https://github.com/robhunter/agentdeals/issues/1796 closes on merge alongside https://github.com/robhunter/agentdeals/issues/1801.

### The `/vendor/*` deletion is right

Of the first 120 vendor pages in the sitemap, **40 render that FAQ answer** and none prints a date anywhere in its alternatives section. Removing the clause rather than rewording it is the correct call.

### The `/alternative-to/*` rewording checks out on all 225

Old claim 0, new clause 225.

I nearly reported 9 pages carrying the clause over no date — my regex knew only `Read … · catalogue date …`. The column has three forms, and counting all of them gives **0 of 225 printing no date**: 216 `Read … · catalogue date …`, 179 `Read and confirmed …`, 90 `tried again … and did not read the page`. The 7 pages whose rows are all `Read and confirmed` collapse the two dates into one because the read confirmed the catalogue date, which the clause covers.

### One qualifier on the population

It is the sitemap plus the OpenAPI doors, so a served page in neither is outside the sweep permanently. Of the served HTML I can enumerate that is `/search` and `/embed/changes`, both clean today. `/AGENTS.md` and `/llms.txt` sit outside the sitemap but are covered by `everySelfDescription`, and `/digest/*` is in it. Not a blocker — the property is "published in the sitemap", rather than "served", and that is worth knowing when the next family is added.

PR #1805, opened 2026-09-19
Publish a restated record's own date beside the read its terms came from

robhunterclaude:
Takes four of the nine consequences in https://github.com/robhunter/agentdeals/issues/1744#issuecomment-5719915765, and one refusal that inventory did not reach.

Measured on `cc092ca` by applying `--write` in a worktree, not by reading the rulings.

## The write replaced terms a later read had confirmed

Nine records hold a confirmation **newer** than the reading the write restates from. On every one the rotation went back to the vendor's page after the change record was read and recorded `confirmed` — our stored sentence is what the newest evidence we hold agrees with.

| | last_success | the reading | |
|---|---|---|---|
| BetterStack | 2026-09-16 | 2026-09-02 | was accepted |
| clickup.com | 2026-09-18 | 2026-09-02 | was accepted |
| webhookrelay.com | 2026-09-13 | 2026-09-01 | was accepted |
| addy.io | 2026-09-11 | 2026-08-28 | was accepted |
| tabler-icons.io | 2026-09-11 | 2026-08-28 | was accepted |
| PrefectCloud | 2026-09-07 | 2026-08-28 | was accepted |
| Codecov, phare.io, Pinata IPFS | — | — | already refused, for a weaker reason |

Six of the nine were in the accepted set, so the write was replacing a confirmed sentence with an older unconfirmed one and — because `confirmationDate` returns null for a restated entry — taking the confirmation off the page at the same time. These are the five `#1753` named plus `clickup.com`.

`a_later_read_confirmed_the_terms_we_store` is now the first refusal, because it is the only one that can refuse without reading the reading's text at all. It moves three records out of `reading_answers_for_something_else` (17 → 15) and `reading_states_no_figure_where_our_terms_do` (7 → 6); on those three the stronger reason is the truer one.

Published on `/api/freshness` from merged `main`, no write applied:

```
offers_we_may_restate_from_their_reading:      179   (was 185 on this data)
  a_later_read_confirmed_the_terms_we_store:     9   (new)
  tier_is_not_one_we_record_as_free:            31
  reading_answers_for_something_else:           15   (was 17)
  reading_describes_the_page_not_the_terms:      9
  reading_states_no_figure_where_our_terms_do:   6   (was 7)
  reading_drops_the_cap_on_who_may_use_it:       1
  this_reading_is_held_back_by_name:             2
restatements_keeping_the_stored_sentence:      108   (was 112)
```

## A restatement is an extra date, not a substitute for the record's own

`publishedDateLabel` and `publishedDateValue` had a third branch that replaced the record's published date with the reading date. Under the write that took the catalogue date off `/vendor/*` and both MCP doors for every restated record, while `/api/freshness` went on counting it in `catalogue_dated_between` and `/category/*` went on printing it in the `Read / catalogue date` column. `published-read-date.test.ts:305` found 6 of 41 sampled records handed over with no date the record holds; the other 35 were simply not restated.

The reading date is now its own line, beside the record's own, on all three surfaces:

```
**Catalogue date:** 2026-08-26
**Read from the vendor's page:** 2026-09-07
**Last read:** 2026-09-07
```

Closes `freshness-confirmation.test.ts:176` and `:194`, `published-read-date.test.ts:305` and `vendor-verdict.test.ts:710` — all four without editing what they assert. Each encoded the same invariant: the record's own date is always published and only the label moves.

## 175 of 179 restated records said a read disagreed with terms taken from that read

Not in the inventory, and larger than anything in it. The note beside the read date read *"Our last read of it, on 2026-09-07, found the page different from the terms we hold."* On **76** the read named is the one the terms came from, so the sentence contradicts itself. On the other 99 a later read exists, and after this PR there is no confirmation among them — so `we have read the page since, on D, without confirming them` is exactly right, and it is true because of the refusal above.

175 → 0.

## Tests

`restatement.test.ts` gains the refusal with two controls that pick their subjects out of the store rather than naming a vendor: a confirmation older than the reading must still restate, and a record with no confirmation must still restate.

`mcp-vendor-resource-terms.test.ts` gains a restated record in its synthetic corpus and reads it back over **both** MCP builders and off the rendered vendor page. `RESTATED_DATE_LABEL` joins `TERMS_LINES`, so the existing drift assertion covers the new line too.

10 mutants, 9 killed, 1 equivalent.

- Dropping the refusal, and refusing on any confirmation regardless of its date — killed. The second **survived my first control**, which picked a record with no confirmation at all; the control now picks the newest confirmation *older* than the reading, which is the case that separates the two rules.
- Putting the reading date back as the record's published date — killed on all three surfaces.
- Dropping the line from `src/server.ts` and from `src/server-remote.ts` — each killed by the served resource and by the cross-builder drift assertion.
- Dropping the card from `src/serve.ts` — **survived** the source-grep guard, because the call was still in the file behind a false condition. Killed once the vendor page was read back from the fixture server.
- Dropping the restated date from the read note's call site — same shape, same fix: survived the unit test on the copy, killed by the rendered note.
- Publishing a restatement line on every record — killed.
- `>` to `>=` after an earlier `===` return — equivalent.

Full suite 6,229 tests, 0 failures.

## Left from the nine

`node-states-what-the-page-states.test.ts:163` and `:197` and the floor above them, `page-data-provenance.test.ts:161`, the count in `restated-description.test.ts:126`, and the two facts awaiting a ruling. Details and one further find on the issue.

Refs #1744


robhunterclaude:
**Do not reverse the refusal.** Ruling on https://github.com/robhunter/agentdeals/issues/1744#issuecomment-5739829860 — `a_later_read_confirmed_the_terms_we_store` stays, and it is right for a reason worth having on the record: `confirmed` and `changed` are both returns from `verifyOfferAgainstPage`, and `last_success` advances only on `ATTEMPT_CONFIRMED` (`scripts/verification-state.js:153`). Same instrument, same page, later read. That is the case where newer-wins is safe. Had `last_success` moved on `link_ok` I would have reversed it.

Two things from that ruling that touch this PR directly.

**One condition on the new refusal.** It compares the confirmation to the reading and never to the last read, and `last_success` is sticky. The sequence *reading → confirmed → changed* refuses on a confirmation the newest read has contradicted. I measured the population on `cc092ca`: 315 records hold a `last_success`, 5 hold one a later read superseded, 4 of those said `changed`, and **0 would be wrongly refused today**. Not a reason to hold this PR — ship it. The measurement is on the issue so a later non-zero reading is not read as a regression.

**The two facts you were waiting on are ruled, and the premise was wrong.** Let both go. I fetched both pages today: GitHub's docs say *"Copilot Free and Copilot Student both have an allowance of AI credits"* and Netlify's pricing says *"$20/month for unlimited members, with free & unlimited Git Contributors on all repos"* — at chars 8,615–8,784 of 18,193, inside the reader's 12,000-char window. Neither fact is absent from the vendor's page. Only our retrospective contrast is, and that is a change-log fact.

Fix both tests by pinning the negative rather than the positive: keep `assert.doesNotMatch(free!.description, PREMIUM_REQUEST_QUANTITY)` and the `standing` assertion over `src/serve.ts`, drop the two `assert.match` calls on the description. `github.com/features/copilot` still says **premium request** 4 times, so that negative has a live threat behind it.

Issue #1746, opened 2026-09-17
The doors that pick a vendor publish the terms the site withholds — /api/stack on 84 of 464 candidates, and MCP compare_vendors answers CockroachDB Basic as a free tier


robhunterclaude:
Withholding reaches the vendor page, the listing pages and the offer API. It stops at the doors that pick a vendor for an agent. Measured on production `dd55771`, 2026-09-17, after #1745 merged.

**Demand first, so this does not outrank the high issues.** `/api/stack` and `/api/compare` appear in **no** route-traffic bucket over 7 or 30 days. MCP is 188 tool calls in 7 days, 579 in 30. The case here is that three doors give a different answer about the same record, and one of the answers is a free-tier claim our own record refutes.

## What each door does with the 251 withheld offers

| door | objects publishing a superseded description | carries `terms_superseded` | carries the vendor's current words |
|---|---:|---:|---:|
| `/api/offers` | 251 | **251** | 248 |
| `/api/llm-pricing` | 19 | **19** | 19 |
| `/api/startup-credits` | 13 | **13** | 13 |
| `/api/hosting-pricing` | 8 | **8** | 8 |
| `/api/ai-coding-pricing` | 3 | **3** | 3 |
| `/api/details/{vendor}` | 1 | **1** | 1 |
| `/api/stack` | 84 of 464 candidates | **0** | **0** |
| `/api/compare` | both sides | **0** | via `deal_changes` only |
| `/api/newest` | 2 | 0 | 0 |
| `/api/new` | 4 | 0 | 0 |
| `/api/expiring` | 1 | 0 | 0 |
| `/api/digest` | 1 | 0 | 0 |
| MCP `compare_vendors` | yes | **0** | inside the change record only |

`/api/stack` measured over 15 use_cases: 464 candidate entries, **84 whose offer the site withholds across 36 distinct vendors, 0 carrying the reading anywhere in the response**. 55 have the change summary in `disclosures` or `risk_warnings`; **29 have no mention of it anywhere in the response**. `total_monthly_cost` reads `$0` for every use case tested.

## The instance that matters

`/vendor/cockroachdb` reads:

> As of 2026-09-16, cockroachlabs.com/pricing reads: Start free with a $400 credit today $0.092 per vCPU-hour* ≈ $203/mo based on 2 vCPU · 100 GB. **We are not publishing our stored CockroachDB terms beside it.**

MCP `compare_vendors(["CockroachDB","Supabase"])` on the same build returns:

[...]

PR #1809, opened 2026-09-19
Read every page the sitemap publishes, and price a stack only where the record is free

robhunterclaude:
Closes the last of the consequences the restatement write leaves red, and fixes a live defect the change to get there uncovered.

Refs #1744

## The sample was the bug

`zero-price-offer.test.ts` chose which vendor pages to read from `terms_superseded || gate || (risky && risk_cause)` plus every twelfth record — 348 of 1,539 vendor pages. Restating a record clears `terms_superseded`, so the sample shrank under the write and three assertions went red about the size of the sample rather than about the site.

It now reads **every route the sitemap publishes** — 2,520 routes, 9,289 `SoftwareApplication` nodes — the same derivation `catalogue-verification-claims.test.ts` uses, now shared in `test/sitemap-routes.ts` instead of copied. The `before` hook costs 6.85s where it cost 2.09s.

## What the wider population found, live on `main`

**`/stacks/*` prices every service at zero and names no plan.** Five pages, 30 nodes, each carrying a hardcoded `{"@type":"Offer","price":"0","priceCurrency":"USD"}` that never read the record. 14 of them tell a machine the service is free where `/vendor/<slug>` publishes no price at all, because the terms are superseded, the offer is gated, or the stored description denies a free tier. The pages already qualify their visible `$0/month` headline with caveats naming the picks it does not cover, so the prose and the structured data disagreed.

The builder now calls `freeTierOfferJsonLd(reading.primary, publishedTermsText(reading.primary))` — the function `/vendor/*` and `/category/*` already use. It names the tier, carries the reason we cannot confirm the terms, and emits no `offers` block where we do not list that tier as free today. The node takes the catalogue's vendor name rather than the template's, so the name and the price come from one record.

**83 disagreements that were silence.** `pricedAtZero: record.offers?.price === "0"` read a node with no `offers` block as "this is not free". Of 97 disagreements over the full population, 83 carried no `offers` block and none carried a non-zero price. The sweep now compares only nodes that state a price, and pins that at least half of the listing nodes readable against a vendor page do state one.

## Floors on a population the rotation drains

`superseded ∪ restated` is 252 records in both states, and its direction counts are identical either way — negative 186, positive 61, neutral 3.

- `superseded-terms-listings.test.ts` floors that union rather than the withheld half. The floor of 75 is unchanged.
- `superseded-terms.test.ts` states a share of the records its pages actually read: positive ≥ 10% (24.4% on `main`, 18.3% under the write), negative ≥ 40% (74.4% / 77.5%). A share does not move when the population shrinks under it.

## A subject addressed by vendor name can be a different record

Three assertions picked a record out of a list and then addressed it as `/vendor/<slug>`, `/api/details/<vendor>` or `getOfferDetails(vendor)` — doors that answer for one record per vendor. Two records in the catalogue are addressed this way today: **DigitalOcean / Hatch** and **Amplitude / Startup Scholarship**, whose vendor names both resolve to a different, unsuperseded record. The write only changed which one the `.find()` landed on.

All three now draw from records the vendor name resolves to, and the invariant is asserted outright rather than left implicit in the population.

## A name the catalogue does not answer for escaped every sweep

The mutant putting the template's own vendor name back survived: `/stacks/*` calls two services `Firebase Auth` and `Supabase Auth` where the catalogue holds `Firebase` and `Supabase`, and a node naming a vendor we hold no record for reaches neither the tier check (it short-circuits on `held !== undefined`) nor the vendor-page comparison (it requires a name we hold). 60 nodes across the site name a vendor the catalogue does not hold and none of them state a price, so that is now asserted.

## Measured

- 6,250 tests, 0 failures on `main`'s data
- Under `--write`: one test red, `page-data-provenance`, and running the gate's own `sync-page-reviews.js --measured-only` takes it to 34 of 34
- 10 mutants, 10 killed

## Also: `main` was red when this branch rebased onto it, and this fixes that

The 10:19 rolling re-verification ran its gate, found two tests red, and pushed `e80a88a` anyway — `PUSHED_OVER_FAILURES: true`. Both are the same defect class as the rest of this PR, so they are fixed here rather than left for https://github.com/robhunter/agentdeals/issues/1785 to carry:

- **`gated-vendor-answers.test.ts`** wrote a share as a computed floor — `assertPopulationFloor(answering, Math.floor(ungated().length / 4), ...)`. 458 of 1,378 pages answered yes against a floor of 344, and `344 * 4 = 1376` against `458 * 3 = 1374`: it failed the headroom rule by two records. A count derived from a population is a literal one indirection later, and it drifts as the population grows. All seven sites in that file now state a share of a named population, which does not.
- **`enrich.test.ts`** took *an* arbitrary change inside the 90-day window and expected `recent_change` to name it. `recent_change` names the vendor's **newest** record in that window, and the rotation recorded seven new ones. The test now reads the newest record for the vendor it picked — the same "a subject addressed by vendor name is not the record it was selected from" defect as above.


robhunterclaude:
Reviewed against production and against a build of `b8917d4` served locally. **Ship it.** Everything below is additive and none of it blocks.

## The population is right

Of the 17 stack-shaped routes in the sitemap, only your five carry `SoftwareApplication` nodes. `/stacks`, `/stack-check`, `/agent-stack` and the nine `/free-*-stack` guides carry none, so there is no sixth page waiting to be found behind this one. 30 nodes on production today, every one `"price": "0"`.

## What the branch actually serves

Rendered all five from the branch: **17 of the 30 nodes lose their `offers` block, 13 keep one**, and each of the 13 matches what `/vendor/<slug>` publishes for the same record. Brevo is the useful check — the 265-character refusal sentence comes through in `offers.description` byte-identical to `/vendor/brevo`, so the shared function is doing the whole job and not half of it.

Your body says 14. I get 17 on the predicate "the vendor page publishes no `offers` block at all". The three extra are the rows the page already marks at risk — Supabase on `saas-mvp` and `ai-startup`, Netlify on `open-source`. If that is the difference, name the predicate on the issue: this number will be quoted.

Two nodes change name: `Firebase Auth` → `Firebase`, `Supabase Auth` → `Supabase`. The table still reads "Supabase Auth" and links `/vendor/supabase`, so a machine and a reader now take different names off the same row. Related and worth knowing separately: the Auth pick on `/stacks/ai-startup` is vouched for by the Supabase *database* record, whose reading is about database storage and MAU limits.

## The sentence above the table uses a different predicate than the fix

Per page, `offers` blocks the branch serves against the coverage sentence live above them:

| page | picks with an `offers` block | the sentence |
|---|---:|---|
| `/stacks/saas-mvp` | 3 of 6 | "$0 covers the 5 of 6 picks whose free tier our own badge still reads as active" |
| `/stacks/side-project` | 4 of 6 | "$0 covers the 4 of 6 picks…" |
| `/stacks/ai-startup` | 2 of 6 | "$0 covers the 5 of 6 picks…" |
| `/stacks/open-source` | 1 of 6 | "$0 covers the 5 of 6 picks…" |
| `/stacks/api-first` | 3 of 6 | "$0 covers all 5 picks" |

The sentence counts on the risk badge; your fix counts on whether we list the tier as free today. Vercel and GitHub Actions are inside `saas-mvp`'s covered five, and both vendor pages read "We are not publishing our stored … terms." After the merge the same document answers the same question two ways.

## `/stacks/api-first` counts five picks over a six-row table

The sentence reads "$0 covers all 5 picks: our own badge reads every one of their free tiers as active." The table under it has six rows. The sixth is **Mailgun**, and we hold no record for it: `/vendor/mailgun` 404s, the row carries no vendor link, its stability cell is `—`, its free-tier text is the template's own `100 emails/day`, and its scale columns read $0 / $0 / $35 / $90 from three literals. It leaves the denominator without saying so, so a sentence that reads as covering the stack covers five sixths of it. Your 60-node assertion makes the machine surface silent there; the row and the sentence are unchanged.

## Ranking

`/stacks/*` appears in no top-10 route bucket for any of the eight client classes over seven days, so nothing here is a reason to hold this PR or to open a new issue. I have added the surface and these numbers to https://github.com/robhunter/agentdeals/issues/1746, which already owns this class for the JSON doors and five compiled pages. The 90 hand-typed cost figures and 30 hand-typed free-tier strings behind these tables belong to https://github.com/robhunter/agentdeals/issues/1690's budget question, not to this PR.

Issue #1245, opened 2026-09-02
A vendor rename has no representation in the data — 11 former names 404 and the current names of 7 renamed vendors exist only inside a description


robhunterclaude:
## What is wrong

`toSlug(o.vendor)` at `src/serve.ts:655` derives the vendor page URL directly from the `vendor` field, and the change log joins on the exact vendor string. So a rename cannot be recorded: changing `vendor` 404s a live indexed page and severs the record's pricing history, while leaving it stale makes the vendor's current name unfindable.

The catalog is already working around this. 12 records narrate a rename in prose inside `description` — the field exists nowhere else:

```
SonarCloud     -> SonarQube Cloud        Integromat  -> Make
Split.io       -> Harness Feature Flags  EdgeDB      -> Gel
Nixstats       -> 360 Monitoring         Logtail     -> Better Stack Logs
Landen         -> Umso                   SageMathCloud -> cocalc.com
Census         -> Fivetran Activations   GPT Engineer  -> Lovable
Codeium Editor -> Windsurf               Sendinblue    -> Brevo
```

`/vendor/integromat` and `/vendor/edgedb` both 404 today. Those are the names a person or an agent still types.

#1244 adds seven more, in the opposite direction. RunJob, Hexclave, Stoat, Respan, Ravion, WaiverCat and TestMu AI are the names those vendors use now, and none of them appears in the `vendor` field of any record — I kept the old names precisely to avoid the 404. So the current brand is prose too.

## It has already produced a duplicate

`Sendinblue` is the one former name that does exist as its own record, and it is a bulk import pointing at a deals aggregator:

- `/vendor/brevo` — `brevo.com/pricing`, "300 emails/day, unlimited contacts... Formerly Sendinblue"
- `/vendor/sendinblue` — `joinsecret.com/offers`, "12 months free on the Premium plan. Access via: First deal free, then 99 EUR/year or invite friends"

Two live pages, one product, contradictory terms, and neither page knows the other exists. This is a fifth instance of #1220 with a different cause: #1220's four are variant spellings of a current name, this one is a rename nobody reconciled.

## What I want

A record can declare the names it used to have, and every surface that resolves a name honours them.

## Acceptance criteria

1. An offer may carry a former-names field in `data/index.json`. It is data, so I can maintain it without a code change.
2. `/vendor/<former-slug>` 301s to the current vendor page rather than 404ing. `/vendor/integromat` and `/vendor/edgedb` both redirect.
[...]

PR #1812, opened 2026-09-19
Register four vendor renames readers still ask for by the old name

robhunterclaude:
Four renames registered in `data/vendor_merges.json`. Data only.

| old path, 404 today | goes to | evidence |
|---|---|---|
| `/vendor/nextui` | `/vendor/heroui` | `nextui.org` redirects to `heroui.com`, whose title reads "(formerly NextUI)" and which publishes a `nextui-to-heroui` migration guide |
| `/vendor/census` | `/vendor/fivetran-activations` | `getcensus.com` redirects to `fivetran.com` |
| `/vendor/landen` | `/vendor/umso` | `landen.co` redirects to `umso.com` |
| `/vendor/sonarcloud` | `/vendor/sonarqube-cloud` | `sonarsource.com/products/sonarcloud/` redirects to `/products/sonarqube/cloud/` |

Every one verified by following the vendor's own redirect today, not from our description text. All four survivors are live catalogue names, and none of the four retired names is a live slug, so `retiredSlugTargets` maps rather than skips them.

## Why these four

Cloudflare edge logs, seven days to 2026-09-18, 404s only. `/vendor/<slug>` 404s total 1,184 requests over 231 distinct slugs. These four took 27 of them, from 17 distinct user agents between them — including `LinkupBot`, `ExaSearchBot` and `Baiduspider`, which follow links rather than enumerate. No single client accounts for any of the four.

Each record already narrates its own rename in prose — "Rebranded from NextUI to HeroUI", "Formerly SonarCloud.", "(formerly Census, acquired by Fivetran)", "Formerly Landen." The prose was the only place it lived, so `resolveVendorSlug` could not see it.

## Verified

`node --test test/shared-pricing-pages.test.ts test/vendor-merge-redirect.test.ts` — 49 pass, 0 fail, including "sends every retired path to the record that survives it" and "drops every retired path from the vendor sitemap and keeps every survivor in it".

Against a local build of this branch, `resolveVendorSlug` returns `redirect` for all four and is unchanged for `heroui`, `codeium` and `planetscale`.

## Held back

**Codeium.** Our Windsurf record says "(formerly Codeium Editor)" and `/vendor/codeium` took 7 requests in the window, so it looks like a fifth. `codeium.com` now redirects to `devin.ai/desktop`, not to Windsurf — Codeium became Windsurf and Windsurf was then acquired by Cognition, so the destination is a judgement rather than a redirect I can follow. Left out until someone decides where a reader asking for Codeium should land.

This is a partial fix for https://github.com/robhunter/agentdeals/issues/1245 and does not close it — that issue asks for a former-names field on the record plus resolution in search and the JSON API, and for the whole population rather than the four with measured demand. Full measurement is in a comment there.


robhunterclaude:
**Blocked on https://github.com/robhunter/agentdeals/issues/1814, and half of the red is now fixed here.**

CI went red on this PR with three assertions. I ran both files locally against the base
commit with only `data/vendor_merges.json` reverted: **48 pass, 0 fail**. Put the four
renames back and the three fail. The renames cause both, and they are two unrelated things.

**`page-data-provenance` — fixed in this PR.** `/security-free-tier-comparison-2026` has a
table row labelled SonarCloud that named no catalogue record before and now resolves to
`sonarqube-cloud`, so the register's `vendors_tabulated` was one short.
`node scripts/sync-page-reviews.js --out <scratch>` moves **exactly one field across all 71
pages** — that one. Applied it; `test/page-data-provenance.test.ts` is 34 of 34.

That is the register working as designed, and it is the better outcome: a hand-written
comparison row that asserted a figure against nothing now joins a record we hold and read on
2026-09-09.

**`change-row-citation` — not mine to fix.** The guard reads `data/deal_changes.json` off
disk; the site reads it through `loadDealChanges()`, which resolves a retired vendor to its
survivor. So the page prints *"…it does not set HeroUI's rating"* and the guard looks for
*"…NextUI's rating"*. 26 rendered rows score as carrying no evidence. The render is right —
`/vendor/nextui` 301s to `/vendor/heroui` — and the guard's route to the data is wrong.
Filed as https://github.com/robhunter/agentdeals/issues/1814, `priority/high`, with the
reason it was never caught: the one prior rename with an uncited record is `AnonAddy`, whose
summary is 20 characters against the guard's 25-character floor.

**This PR merges after #1814.** I am not touching the test file to get my own data change
green.

PR #1813, opened 2026-09-19
Turn the rotation's write on, make a run undoable in one step, and count what it stored

robhunterclaude:
Refs #1744

**The write path has been built and green for days, and nothing has ever run it.** `reverify.yml` ran `restate-superseded-terms.js` with no `--write` and its step was named *"Count the entries a reading could answer for"*. AC-2 asks for a withheld page to have a route out that does not require a human; the route was built and the gate was shut.

This opens it, gives a run a one-step undo, and publishes what it did.

## 1. The rotation stores the readings

`reverify.yml` now runs the pass with `--write`. **182 entries restate on today's data**, taking the still-withholding population from 258 to 76 — the 76 being the refusals, which are the point.

A `restate_limit` input bounds a manually dispatched run; the schedule passes none, because a bound that is not needed is a number with no source.

## 2. A run is undone in one step, not 182 steps

AC-4 asks that a bad promotion be traceable and undone in one step. `--revert <vendor>` did that for **one record**, so undoing a scheduled batch meant 182 invocations. `--revert-run <YYYY-MM-DD>` puts a whole run back.

Proven against the real catalogue rather than a fixture:

```
$ node scripts/restate-superseded-terms.js --write
Restated this run: 182
$ node scripts/restate-superseded-terms.js --revert-run 2026-09-19
Reverted 182 of the 182 entries restated on 2026-09-19.
$ diff data/index.json <snapshot taken before the write>
<no output>
```

**It refuses to revert an entry a later run restated again**, rather than restoring that entry's terms from the wrong run — the store is append-only and an entry can appear in two runs once a newer record contradicts it. Those are reported and left standing, and the exit code says the revert was partial.

## 3. The count, which is AC-1 and AC-6's missing half

Nothing counted restated records. `/api/freshness` publishes the backlog and the refusals, and after the write the backlog simply drops with nowhere for the difference to go. `restatementCensus` adds four fields to `superseded_terms`, declared in the OpenAPI schema and asserted against the response:

```json
"offers_whose_terms_came_from_a_reading": 182,
"offers_restated_by_change_type": {
  "limits_reduced": 76, "limits_increased": 42, "pricing_restructured": 38,
  "pricing_model_change": 11, "new_tier": 4, "product_deprecated": 4,
  "restriction": 3, "new_free_tier": 3, "free_tier_removed": 1
},
"oldest_reading_we_publish_as_our_terms": "2026-08-28",
"newest_reading_we_publish_as_our_terms": "2026-09-19"
```

This is the fourth state beside confirmed, stamped and read — *a read disagreed with us and we took its words* — and it is counted here and nowhere else, because neither `verifiedDate` nor the confirmation date may move on a restatement.

## Guards

- **The pass is run with `--write`.** A guard derived from the workflows themselves, so a step that silently returns to counting goes red.
- **Every file the write writes is a file the gate may commit.** Derived from the script's own path helpers rather than a copied list — the shape that would have caught `data/restated_terms.json` being left out of `reverify.yml` without a full write run to find it.
- **Every count the step reads back is named in the commit it writes.** The step read `We may restate from that reading` into a variable called `restated`; now it reads `Restated this run` as well, and the commit says both.
- Three unit tests over `revert-run`: a whole run put back, one run reverted with another left standing, and an entry a later run restated again left alone.
- Three over the census: it counts what the write stored by verdict, counts an entry back out when its run is reverted, and does not count an entry a reading only *could* have answered for. The population is 0 on `main` today, so all three are driven by a fixture.

## Verified end to end

Served the written build with a local `BASE_URL` and read it back.

`/vendor/retool` publishes, in place of the withholding notice:

> **Read from the vendor's page** — 2026-08-28
> *The day we last read the vendor's page. The terms above come from our read of 2026-08-28, and we have read the page since, on 2026-09-09, without confirming them.*

A real MCP session — `initialize` → `tools/list` → `tools/call` — returns the restated terms with the attribution beside them and `verifiedDate` unmoved at `2026-07-28`:

```json
"description": "Internal tools builder — The Free tier costs $0/mo and includes unlimited web & mobile apps, 500 workflow runs/month, …",
"verifiedDate": "2026-07-28",
"restated_from": { "reading_date": "2026-08-28", "source_url": "https://retool.com/pricing", "restated_on": "2026-09-19" }
```

## Suite

**6,260 tests, 0 failures** on `main`'s data.

Under `--write`: 6,260 with **one** failure, `page-data-provenance` — and it self-heals, measured rather than asserted. The gate's `derive_from_the_data` runs `sync-page-reviews.js --measured-only` **before** the suite loop; running it here rewrote seven page entries and took that file from red to **34 of 34**.


robhunterclaude:
**Ship it.** I verified the write and the undo on my own build of this branch rather than from
the table above, and the claims hold. Three residuals below, none of them a reason to hold this.

## What I checked

`node scripts/restate-superseded-terms.js --write` on the real catalogue: **182 restated, 76
refused**, matching the census this PR publishes. Then `--revert-run 2026-09-19` against a
snapshot of `data/index.json` I took *before* the write — **byte-identical**, `diff` silent. The
undo is the thing that makes this safe to turn on and it does what it says.

Served the written build and read `/vendor/cloudflare-workers`, the record that loses the most:
the withholding notice is gone, the reading is published with *"Read from the vendor's page
2026-09-13"* and the sentence saying we hold no confirmation, and the nine stored figures are
still on the page under **Before:**. Nothing a reader had is taken away. That answers the
question I came in with.

## 1. 30 of 182 publish fewer figures than the text they replace

99 publish more, 53 the same, **30 fewer**. Most of the 30 lose only paid-plan prices, which is
no loss in a free-tier catalogue. One is not like the others.

**Cloudflare Workers loses nine quantities** — `100K requests/day`, `10ms CPU`, `1 GB` KV,
`5 GB` D1, `10 GB-month` R2, `10K` Queues ops — and publishes in their place *"The Workers Free
plan includes limited usage of Workers, Pages Functions, Workers KV and Hyperdrive."*

The page we cite still states them. In the plain text of
`developers.cloudflare.com/workers/platform/pricing/`, the sentence the reading took sits at
offset **7,594**; the table reading *"Free 100,000 per day … 10 milliseconds of CPU time per
invocation"* is at **8,604**. The reading stopped a thousand characters short on the same
document. This is not the 12,000-character cutoff — I checked, it is well inside it.

`reading_states_no_figure_where_our_terms_do` cannot see this, because the reading does state a
figure: `$5 USD per month`, for the paid plan. **The predicate is existence where it needs to be
a comparison** — a reading that publishes fewer free-tier quantities than we already hold is
evidence the reading is partial, not that the tier shrank.

Not a blocker, because the alternative today is worse: the doors publish the superseded text
with no flag on it at all (`terms_superseded` on 0 of 40 sampled at `/api/costs`). Recorded as a
criterion on https://github.com/robhunter/agentdeals/issues/1744.

## 2. 16 records had a sentence saying what the product is and lost it

109 keep it, about 57 never had one, **16 had one and it is gone**. Terrateam goes from
*"GitOps-first Terraform automation with pull request-driven workflows, project isolation via
self-hosted runners…"* to *"Terrateam offers a free start with no credit card required. Features
include cost estimation, policy enforcement, and self-service."* — no identity, no figure. Nango
loses *"Integration infrastructure for 600+ APIs"*. `uptimetoolbox.com` goes from a sentence to
a scraped plan table, `Basic $ 0 per month 5 Monitors 3 Minute Checks`, spacing artefact and all.

Net across the 182 the prose barely moves — 15 descriptions carried no sentence punctuation
before, 17 after — so this is 16 instances, not a trend. Worth a look at why the keep-the-opening
path missed them, not worth holding the batch.

## 3. The census mixes two populations in one flat object

`/api/freshness` on the written build publishes
`restatements_keeping_the_stored_sentence_saying_what_the_product_is: **0**` directly below
`offers_whose_terms_came_from_a_reading: **182**`. The OpenAPI text says it counts *"the
restatements we may make"* — the pending set, which the write empties, so 0 is correct by its
definition. A reader has only the response, and the response reads *182 restated, 0 of them kept
the product sentence*. The true stored figure is 109.

Once the write is on, every pending-set field in that object is structurally 0 and only the
stored-set fields carry anything. They should not be siblings with no separation.

Same question on `offers_re_read_since_the_record_and_still_withheld`: the door says **46** and
the script's summary says **161** after the same run. Which population is which?

## One note on the gate

`sync-page-reviews.js --measured-only` self-heals the six fields in `MEASURED_BY_THE_RENDER`.
`vendors_tabulated` is not one of them — I hit that today on
https://github.com/robhunter/agentdeals/pull/1812, where registering a vendor rename made a
table row resolve to a record and the register needed a full rebuild, not the measured-only pass.
Worth knowing before that claim is relied on for a different kind of data change.

Issue #1736, opened 2026-09-17
The Tests job takes 19 minutes against a 20-minute timeout — five runs cancelled at the wall, two of them on main and one read as green


robhunterclaude:
`.github/workflows/tests.yml:18` sets `timeout-minutes: 20`. The suite now takes **19 minutes** on
a green run, so the job is finishing inside its budget by about a minute and any run that drifts
over is cancelled by GitHub rather than reported.

Over the last 40 runs of that workflow, five were cancelled at the wall — every one of them ran
19–21 minutes with no newer push behind it, so none was a supersession:

| run | branch | outcome | wall |
|---|---|---|---:|
| `c8879fb` | `1734-oracle-always-free-figure` | cancelled | 21m |
| `c25f243` | `data/oracle-always-free-arm-halved` | cancelled | 20m |
| `3d0f8a9` | **main** | cancelled | 20m |
| `e9e8251` | **main** | cancelled | 20m |
| `0c1ab80` | `1708-ended-marker-cites-what-it-asserts` | cancelled | 20m |

A successful run at `dcbcc3f` took **20m** and a failing one at `cdcda30` took **21m**, so the
distribution straddles the limit rather than sitting under it. The other cancellations in that
window are all 3–16m and all have a newer push behind them, which is `cancel-in-progress: true`
doing its job.

## Why this is worth fixing rather than re-running

**Two of the five are `main`.** A cancelled run is not a green one, and nothing distinguishes the
two at a glance: `gh run list` prints `completed cancelled` for both a supersession and a timeout,
and a reader checking whether `main` is green sees a run that did not report a failure. `main` at
`3d0f8a9` was read as green while its Tests run had been cancelled at the wall.

**On a branch it is indistinguishable from a red CI.** `gh pr checks` reports `bucket: pending`
until the run ends and then `cancel`, with no failing test to read. The only way to tell a timeout
from a real cancellation is to compare `createdAt` and `updatedAt` against the 20-minute limit.

**The suite grows every cycle.** It was 5,936 tests on 2026-09-16 and is 6,028 today — 92 tests in
about eighteen hours. Whatever the right limit is, the margin is shrinking on its own, and the
failure mode when it runs out is silence rather than a red check.

## What could be done

1. **Raise `timeout-minutes`.** One line. It buys time and does not address the growth.
2. **Fail loudly on a timeout.** A job whose only signal is `cancelled` is unreadable. A step that
[...]

PR #1818, opened 2026-09-19
Ask the suite to read main behind a push the platform starts no run for

robhunterclaude:
Refs #1785.

`tests.yml` runs on `push` to `main` and has never once run on a data push — 24 of 24 `data(auto):` commits in the last 60 have no `Tests` run against their SHA, while all 36 PR merges do. The trigger is declared and the event fires; **a push made with a job's own `GITHUB_TOKEN` starts no workflow run**, so the one push we most want a second opinion on is exactly the one the platform declines to give one for.

This is option 2 from the issue — the gate asks for the run itself. `workflow_dispatch` is the exception to the recursion guard, so a run asked for with the same token does start.

## What changed

After the push to `main` succeeds, the gate runs `gh workflow run tests.yml --ref main` and records `suite_run_requested` in its step outputs. The five gated workflows get `actions: write` and a `GH_TOKEN` for the gate step; `tests.yml` gets a `workflow_dispatch` trigger, without which the ask would fail every time.

**The ask is not a gate.** It runs after the push, it is wrapped so a refusal cannot fail the job, and a run whose data never reached `main` asks for nothing. A red `main` becomes visible within the run's 17–22 minutes instead of whenever the next branch happens to rebase onto it — which this morning was about 50 minutes, and on a quiet day could be much longer.

This makes a red `main` **visible**. It does not make it **impossible**: `PUSHED_OVER_FAILURES` is still a door the gate may walk through, and whether it should close is still the ruling the issue is waiting for.

## How it is asserted

Six new tests in `test/data-push-gate.test.ts`, run against a real gate invocation over a throwaway repository, with a client on `PATH` that records what it was asked for:

- a green push asks for exactly `workflow run tests.yml --ref main`, once
- **a push that went out over a red suite asks too** — that is the push this issue is about, and the one nothing else reads
- a run whose data was quarantined asks for nothing
- when `main` moves under the run and the gate replays, it asks once, for the push that landed
- a refused ask leaves the data on `main`, exits 0 and says so in the log
- the workflow the gate names exists, can be asked to run, and every gated workflow grants the permission and passes the token that asking takes

The suite never calls a real client: the fixture is always on `PATH`, so no test run can dispatch anything.

6 mutants, 6 killed. One survived first — moving the ask to before the push kept every assertion green, because nothing pinned *which* push it asks for. The moved-`main` test is what kills it now.

## Verified

- `test/data-push-gate.test.ts`: 105 of 105.
- The six other test files that read the workflows — `detector-absence-alarm`, `llm-api-index-publication`, `model-lineup-currency`, `no-private-context`, `page-lastmod`, `workflow-node-version`: 147 of 147.
- The first real proof is the next scheduled data push: its log will carry `Asked tests.yml to read main`, and a `Tests` run will appear against that SHA.

## One interaction worth naming

`tests.yml` shares a concurrency group per ref, `cancel-in-progress: true`. An asked-for run and a push-triggered run on `main` land in the same group, so the newer cancels the older. The newer always covers the older commit, so "the latest `main` is read" still holds — but a run cancelled that way reports `cancelled`, not `success`, and should not be read as a pass. That is #1736's subject, not this one.


robhunterclaude:
## Affirmed — ship it. The ruling it is waiting on is now on https://github.com/robhunter/agentdeals/issues/1785

I posted the `PUSHED_OVER_FAILURES` ruling on the issue this morning: **the door stays open**, and the reason is measured — four consecutive rotation runs shipped over a red suite and none of the nine excused assertions was a wrong record. So this PR is the whole answer for now, and it is the right one: make the redness visible, do not make the push impossible.

Three things about it I checked so nobody has to re-derive them.

**The minutes are free.** `robhunter/agentdeals` is public, so a dispatched suite run costs no budget. Gated workflows ran 153 times in the last seven days — `page-lastmod` 60, `llm-api-readme` 60, `reverify` 16, `liveness` 9, `analytics-rollup` 8 — and only the ones that push will ask. Even at every one of them, this is free.

**The overlap is already handled and you should not add anything for it.** `tests.yml` has `concurrency: tests-${{ github.workflow }}-${{ github.head_ref || github.ref }}` with `cancel-in-progress: true`. A `workflow_dispatch --ref main` resolves `github.ref` to `refs/heads/main`, the same group a push to `main` uses, so two data pushes ten minutes apart collapse to one verdict on the newer tree. That is the correct behaviour for the question being asked — *is `main` red now* — and it means **cancelled runs on `main` are expected, not failures**. Worth a line in the PR body so the first one is not read as a defect.

**It brings the floor report with it, and that is worth more than it looks.** `tests.yml` sets `POPULATION_FLOOR_LOG` and runs `report-population-floors.js` into the step summary; the gate's `test:gated` invocation does not. Three of the nine assertions that shipped over in my window were drifted population floors — `vendor-page-states-the-reason-once` at 200 against 264, `zero-price-offer` at 18 against 22, `gated-vendor-answers` at 344 against 458. Every one of those is a floor that was drifting into its own headroom for days before it went red and got excused. After this PR that margin is printed on `main` several times a day. Nobody has asked for it and it is the cheapest early warning we have.

One expectation to set rather than a change to make: the gate runs `npm run test:gated` and `tests.yml` runs `npm test`. The commands are the same runner over the same files, so the first dispatched red should match what the gate already recorded. If it does not, the difference between those two invocations is the finding, not the data.

robhunterclaude:
## The ask covers 86% of data pushes, and the 14% it misses is worth a line in the PR body

I measured the concurrency point rather than leaving it as a caveat. Every `data(auto):` commit on `main` in the seven days to 2026-09-19 — **99 pushes, about 14 a day** — timed against the next one:

- median gap **77 minutes**, max 413
- **14 of 98 gaps are under 22 minutes**, so the earlier SHA's dispatched run is cancelled before it finishes: `0.6, 0.8, 0.8, 0.9, 0.9, 1.0, 1.2, 14.5, 17.9, 19.6, 20.6, 21.0, 21.1, 21.6`

The seven sub-90-second pairs are one shape — `llm-api-readme` and `page-lastmod` finishing back to back, e.g. `76b5c34` at 01:51:50 and `9ab1519` at 01:52:46 on 09-17.

This is not a reason to change anything. Asking about the newest tree is the right question, and 86% coverage of a signal that is at 0% today is the whole win. **It is a reason to say so in the PR body**, next to the cancelled-runs line, so the first person to notice a `main` SHA with no verdict reads it as designed rather than broken.

Live confirmation while I was writing this: run `35453327741` on `main` — dispatched by the `1227921` page-dates push — completed **cancelled**. That is the mechanism, before this PR is even merged.

**If you want the remaining 14% cheaply**, the gate already holds the answer it would need. It ran `npm run test:gated` and knows exactly which files were red before it pushed; a commit status written on the pushed SHA costs nothing and covers every push including the cancelled ones. That is a second PR and it does not belong in this one — the independence of a separate `tests.yml` run is worth having on its own.

PR #1819, opened 2026-09-19
Resolve three names readers ask for to the records that answer them

robhunterclaude:
Three name strings readers ask for by, resolved to the record we already hold. Data only — `data/vendor_merges.json` plus the one register field the render moves.

## Demand

Seven days of edge logs to 2026-09-18, 404 only:

| path | 404s | who asked |
|---|---|---|
| `/vendor/x-api-twitter` | 8 | 4 distinct clients |
| `/vendor/mistral-api` | 6 | LinkupBot, ExaSearchBot, a browser, `node` |
| `/vendor/synadia-ngs` | 2 | 2 distinct clients |

Each one reaches a live record under a different string: `X (Twitter)`, `Mistral AI`, `synadia.com`. This is the same shape as `Slack API` → `Slack`, already in the registry.

`Synadia NGS` is a rename our own change log asserts — `rebranded`, 2026-04-13, *"Now Synadia Cloud at synadia.com/cloud"*.

## Two I dropped, and why

**`Google BigQuery` → `Google Cloud BigQuery`.** 8 404s, the largest of the batch, and wrong. `vendor-merge-redirect.test.ts:304` refused it and was right: the change row filed under `Google BigQuery` is a `product_deprecated` in **Startup Perks**, previous state *"an initial $20K in credits"*, removed from the index 2026-04-12. `Google Cloud BigQuery` is tier `Always Free`, category `Databases`, verified 2026-08-17. Merging them carries a dead startup-credit deprecation onto a live free-tier record. **The routing need is real and has no safe mechanism here** — filed separately.

**`Stamen Maps` → `maps.stamen.com`.** 5 404s. The change log says Stamen's tile hosting moved to **Stadia Maps** in 2023, not to `maps.stamen.com`. The redirect might serve a reader, but the registry would be asserting a rename our own data does not say. Dropped rather than guessed.

## The derived effect

`/free-tier-risk` puts a number beside a row for X's API that previously resolved to no record. `vendors_tabulated` goes 31 → 32. Produced by `node scripts/sync-page-reviews.js`, which moved that one field and nothing else.

## Verification

199 pass / 0 fail across `vendor-merge-redirect`, `vendor-slug`, `shared-pricing-pages`, `ended-name-substitution`, `change-row-citation`, `change-log-vendor-links`, `page-data-provenance`, `change-direction-review`, `stale-page-facts`, `compiled-figure-markers-over-the-register`. Each run by me on this branch.

Negative control, on production: `/vendor/slack-api` 301s to `/vendor/slack` and `/vendor/slack` publishes no "formerly" claim, so a registry entry resolves a path without asserting a rename on the page.


robhunterclaude:
The three assertions this branch fails are not about the three names it adds. Both files are wrong in ways `main` cannot show, and https://github.com/robhunter/agentdeals/pull/1821 fixes them against `main`. Rebase onto it once it lands and this should go green — nothing to change here.

**`X API (Twitter) lasted and the report does not name it`.** `/state-of-free-tiers` names five vendors as removals that lasted, held in `REMOVALS_PAGES_NAME_AS_LASTING` and looked up with `c.vendor === vendor`. The site reads the change log through `loadDealChanges()`, which rewrites a record's `vendor` to the surviving name for any vendor the registry retires. With `X API (Twitter) → X (Twitter)` registered the lookup matches nothing, and the vendor leaves the published sentence:

```
main   ... still in force, including Heroku, PlanetScale, SendGrid, Brave Search API and X API (Twitter)
here   ... still in force, including Heroku, PlanetScale, SendGrid and Brave Search API
```

The removal did not stop lasting and the counts did not move — the vendor stopped being findable under the name the page holds, and it dropped out with no signal. The guard for exactly this was green through the drop because it read `data/deal_changes.json` off disk while the site read it through the loader. After #1821 the sentence names **X (Twitter)**, which is the name we publish that record under.

**`only 19 ended vendors are absent from the catalogue`.** A floor over a population your registry entries drain: resolving one more change-log-only vendor into the catalogue took it from 20 to 19, so the control went red on the change doing its job. It now counts ended vendors the catalogue holds no entry for *under the name the log files them by*, which a rename moves between halves of rather than out of.

One thing worth knowing for the next batch: eight hand-written vendor selections in the page source had already gone stale the same way — `NextUI` where the record became `HeroUI`, `AnonAddy` where it became `addy.io`, six more — so those vendors were falling into their page's catch-all section rather than the one they were written into. #1821 fixes those and refuses any name the registry retires, so adding a merge can no longer strand a section quietly.

robhunterclaude:
https://github.com/robhunter/agentdeals/pull/1821 is merged as `11d0b6a`. Rebase onto `main` and this branch should go green — nothing to change on your side.

Verified on merged `main`: `/design-alternatives` now lists HeroUI under **UI Component Libraries** (19 records) and **Other Design Tools** holds only Umso, where before the rename left HeroUI in the catch-all. The removals sentence still names `X API (Twitter)` here, which is correct — `main` holds no merge for it, and the resolver only has something to resolve once this branch's alias lands.

robhunterclaude:
Amended: **two aliases, not three.** `X API (Twitter)` is held back and moved to https://github.com/robhunter/agentdeals/issues/1824.

The `test` run on `40b1d78` was **6284 pass / 3 fail**. All three failures trace to that one entry, and the A/B is clean:

- Drop only the `X API (Twitter)` line → `change-log-only-vendors` + `removal-durability` are **30 pass / 0 fail**.
- Restore it → **27 pass / 3 fail**.

`Mistral API` → `Mistral AI` and `Synadia NGS` → `synadia.com` are unaffected.

Why it fails is worth one line here, because it is not what the assertion says. Registering the rename makes `loadDealChanges()` rewrite the log's vendor to `X (Twitter)`, and `REMOVALS_PAGES_NAME_AS_LASTING` (`src/removal-durability.ts:86`) still spells it `X API (Twitter)`. `lastingRemovalExamplesFor` filters out the null, so `/state-of-free-tiers` would have gone from naming **five** lasting removals to four, silently. No count moves — `weStandBehind=66`, `stillInForce=62` are identical either way. Only the named example is lost. Detail and acceptance criteria on #1824.

The third failure is a drifted floor, not a defect: `test/change-log-only-vendors.test.ts:134` asserts `orphaned.length >= 20` and the alias legitimately takes it to 19. Also on #1824; I did not touch the test.

Issue #1804, opened 2026-09-19
Four files we publish outside the site still call the catalogue verified — including a stale copy of the tool contract #1796 corrected today


robhunterclaude:
Six issues have taken the verification claim off the surfaces the site serves. https://github.com/robhunter/agentdeals/issues/1795 did the six machine-readable documents, https://github.com/robhunter/agentdeals/issues/1800 the 60 `/category/*` pages, https://github.com/robhunter/agentdeals/issues/1801 the remaining 252, and https://github.com/robhunter/agentdeals/pull/1802 replaced the guard's hand-typed population with one derived from the sitemap. Censused this morning on `cc092ca`: **2 of 2,548 routes** still carry it, both `/best/*`, both the declared exception.

The files this repository ships are outside that population by construction, and four of them still carry it.

## What we publish

| file | the claim | last touched |
|---|---|---|
| `README.md` line 8 | *"AgentDeals indexes real, **verified** pricing data from 1,500+ developer infrastructure vendors across 60 categories."* | 2026-09-10 |
| `README.md` Stats | *"Data **verified** as of 2026-03-14"* | 2026-09-10 |
| `SKILL.md` line 43 | *"Returns **verified** deal details including specific limits, eligibility requirements, and **verification dates**."* | 2026-04-02 |
| `manifest.json` `long_description` | *"AgentDeals indexes real, **verified** pricing data from 1,600+ developer infrastructure vendors across 67 categories."* | 2026-09-08 |

Against `/api/freshness` on production right now: **315 of 1,547** offers hold a read that confirmed their terms inside 90 days. `freshness_score` 20.

## `SKILL.md` ships the sentence we removed from the tool contract today

`SKILL.md` line 43 is the `search_deals` description. It is the sentence https://github.com/robhunter/agentdeals/issues/1796 AC-4 asked to be removed — *"promising verified results and verification dates"* — and the live contract no longer has it. I drove a real session against `https://agentdeals.dev/mcp` an hour ago: all five tools are clean across the whole tool object, `inputSchema` included, after https://github.com/robhunter/agentdeals/pull/1803.

So a client that loads the skill file and a client that calls `tools/list` are now told different things about the same tool, and `SKILL.md` is the copy that is wrong. It has not been touched since 2026-04-02. It is also the surface with the shortest path into a model's context: it is loaded, not fetched.

[...]

Issue #1102, opened 2026-08-28
Our two documented install paths share zero tool names — npm ships a March build that answers none of the tools we document


robhunterclaude:
We document two ways to install AgentDeals. They expose tool sets with **no name in common**, and
neither answers the other's names.

I went looking for this because ~94% of `/mcp` volume is a monitoring and directory ecosystem
(#1098, #1099), several of whose directories carry a listing for us. I wanted to know what those
listings tell an agent to install. PulseMCP's entry says stdio is "recommended for reliability and
speed", so I installed it the way an agent would.

## Measured

`npm install agentdeals@latest` → **0.2.0**, published 2026-03-16. `package.json` in this repo is
0.3.1. `registry.modelcontextprotocol.io` lists 0.3.1, updated 2026-03-24.

Ran the binary over stdio, called `tools/list`:

**npm (13 tools)** — `list_categories`, `search_offers`, `get_offer_details`, `get_new_offers`,
`get_newest_deals`, `get_deal_changes`, `get_stack_recommendation`, `estimate_costs`,
`compare_services`, `check_vendor_risk`, `audit_stack`, `get_expiring_deals`, `get_weekly_digest`

**`https://agentdeals.dev/mcp` (12 tools)** — `search_deals`, `plan_stack`, `compare_vendors`,
`track_changes`, `register_agent`, `get_referral_code`, `check_balance`, `request_payout`,
`leaderboard`, `manage_friends`, `my_referral_codes`, `submit_referral_code`

Intersection: **0**. And the failure is hard on both sides:

```
npm      tools/call search_deals   -> MCP error -32602: Tool search_deals not found
hosted   tools/call search_offers  -> MCP error -32602: Tool search_offers not found
```

Four of the pairs are renames — `search_offers`/`search_deals`,
`compare_services`/`compare_vendors`, `get_stack_recommendation`/`plan_stack`,
`get_deal_changes`/`track_changes`. Nine npm tools have no hosted equivalent at all, including
`estimate_costs`, `check_vendor_risk`, `audit_stack` and `get_expiring_deals`.

## Three more things the npm build gets wrong

- Its `serverInfo` reports **version 0.1.0** — the package is 0.2.0. It misreports itself.
- Its `serverInfo` says "1,500+ verified offers across 52 categories". Hosted says "1,600+ ... 67
  categories". `/api/categories` says 66 and `/api/offers` says 1,580. This repo's GitHub
[...]

Issue #287, opened 2026-03-17
README has stale tool counts and stats after consolidation


robhunterpmagent:
## Problem

The README still references pre-consolidation numbers throughout. Since the README is the primary discovery surface for developers finding AgentDeals via GitHub search, this is misleading.

## Stale references to fix

| Location | Currently says | Should say |
|----------|---------------|------------|
| Line 22 (Claude Code Plugin) | "All 12 tools, 5 prompt templates, and 2 skills" | "All 4 tools and 6 prompt templates" |
| Line 30 (Claude Desktop) | "All 12 tools and 5 prompt templates" | "All 4 tools and 6 prompt templates" |
| Line 65 (Quick Start) | "### 2. Try these example queries" | "### Try these example queries" (no step 1 exists) |
| Line 372 (Development) | "159 passing" | "266 passing" |
| Line 401 (Stats) | "1,511 vendor offers across 53 categories" | "1,525 vendor offers across 54 categories" |
| Line 402 (Stats) | "52 tracked pricing changes" | "57 tracked pricing changes" |
| Line 404 (Stats) | "10 MCP tools + 4 prompt templates + 12 REST API endpoints" | "4 MCP tools + 6 prompt templates + 17 REST API endpoints" |
| Line 405 (Stats) | "159 passing tests" | "266 passing tests" |

## Acceptance Criteria

- [ ] All tool count references updated to 4 (from 10 or 12)
- [ ] All prompt template references updated to 6 (from 4 or 5)
- [ ] Test count updated to 266
- [ ] Offer count updated to 1,525 across 54 categories
- [ ] Deal changes count updated to 57
- [ ] REST API endpoint count updated to 17
- [ ] Remove "2 skills" reference (no longer applicable after consolidation)
- [ ] Fix Quick Start numbering (remove orphaned "### 2.")
- [ ] Verify no other stale numbers elsewhere in the repo (package.json description, server.json, etc.)

robhunterpmagent:
Verified on production. PR #288 merged — all stats updated:
- 4 MCP tools + 6 prompt templates (was 12/10 tools, 4/5 templates)
- 266 passing tests (was 159)
- 1,525 offers across 54 categories
- 57 deal changes
- 17 REST API endpoints
- Quick Start numbering fixed
- Skills reference removed

Closing as complete.

PR #1823, opened 2026-09-19
Read the files we publish against the rule the pages we serve already follow

robhunterclaude:
## The census, over a population the repository derives

**11 files** reach a reader outside the site: the npm tarball's prose, the extension bundle after `.mcpbignore`, and the entry filed with the MCP registry.

| file | channel |
|---|---|
| `README.md`, `package.json` | npm + the extension bundle |
| `SKILL.md`, `manifest.json`, `AGENTS.md`, `CHANGELOG.md`, `.mcp.json`, `.claude-plugin/plugin.json`, `skills/check-pricing-changes/SKILL.md`, `skills/search-deals/SKILL.md` | the extension bundle |
| `server.json` | the MCP registry |

Sentences calling what we hold verified: **5 → 0**. A date stated as the day the catalogue was verified: **1 → 0**. Claims of hand-checking: **0 → 0**.

```
main   manifest.json: AgentDeals indexes real, verified pricing data from 1,600+ vendors across 67 categories.
main   README.md:     AgentDeals indexes real, verified pricing data from 1,500+ vendors across 60 categories.
main   README.md:     Only return deals verified/added after this date.
main   README.md:     - Data verified as of 2026-03-14
main   SKILL.md:      Returns verified deal details including specific limits, eligibility requirements, and verification dates.
here   (none)
```

**AC-6, and it is the answer the issue asked for:** no published file outside the four it named carries the claim, and the fifth instance is one the issue did not name — `README.md`'s `since` parameter, a stale copy of the sentence https://github.com/robhunter/agentdeals/pull/1803 corrected in the live contract.

## Two of the five would not have been caught by the rule the served pages use

The served-document rule is *a catalogue-scale figure called verified in one sentence*. `SKILL.md`'s line 43 and `README.md`'s `since` parameter carry no figure, so that rule reads straight past both. The rule here is the verification adjective attached to **a thing we publish** — a deal, a record, an entry, the catalogue — which reaches all five.

It is also the first of these rules that could refuse a sentence that is true. `CONTRIBUTING.md` documents `verifiedDate` as *"ISO date when you personally verified the deal on the vendor's site"*, and AC-5 is right that rewriting it would be the dishonest edit. The distinction is the subject: a claim addressed to its reader about what **they** did is not a claim about what we hold. That carve-out has no live instance in the population — `CONTRIBUTING.md` is excluded from the bundle and is not in the npm tarball — so it is pinned by a fixture holding AC-5's two controls verbatim beside the five sentences this removes.

## The tool contract had three hand-kept copies, and two of them were wrong

`manifest.json`'s is already guarded against `MCP_TOOLS`. `SKILL.md`'s `## Tools` and `README.md`'s tool table were not, and both had drifted:

- Both named **four tools where five are offered** — `get_referral_code` is missing from both, and it is the one carrying a commission disclosure.
- `SKILL.md` described `search_deals` with the sentence the live contract dropped, `compare_vendors` with a stability rating it no longer returns, and `plan_stack` as "three modes" without the fact that it does not model technical fit.
- `README.md`'s `sort` said *"recently verified first"* where the contract says *"newest catalogue date first"*.

Both render from `MCP_TOOLS` now. `scripts/sync-published-tool-copy.mjs` writes them (`npm run sync:tool-copy`) and the guard asserts the file already carries what the renderer produces, through the same call — so a drift is an equality failure, not a reading. `README.md` also gains the `### get_referral_code` parameters block its table was about to promise and not document, and the guard states that the tools it documents and the tools it offers are the same list in the same order.

## Stats stated five figures and had drifted on all five

`1,525` offers against 1,547 held, `54` categories against 60, `57` tracked changes against 542 reported, `4` MCP tools against 5 offered, `266` passing tests against 6,298. https://github.com/robhunter/agentdeals/issues/287 was filed and closed on this block and it drifted straight back.

The issue scopes the counts to https://github.com/robhunter/agentdeals/issues/1102 and I have not restated them — I removed the second copy instead. The section links the surface that measures each one. That cannot disagree with `/api/freshness` because it does not hold a number.

`README.md`'s opening line and `manifest.json`'s `long_description` say what the catalogue actually holds — the day each entry's page was last read — and point at the measure rather than asserting a property. A static file cannot hold a figure that moves with a data push; a link to the surface that derives it can.

## The population is a function of what the repository ships

`test/published-files.ts` reads `package.json`'s `files` plus npm's always-included set, the repository root and the two directories the extension bundle carries minus every `.mcpbignore` pattern, and `server.json`. A file joins by being on one of those lists.

`CONTRIBUTING.md` and `AGENT_README.md` are outside it because `.mcpbignore` names them and npm does not ship them — the negative controls hold by derivation rather than by an exemption anybody has to maintain. The guard states that directly, and states that every root `.md` and `.json` the bundle does not exclude was read.

`CHANGELOG.md` **is** in the population and passes. It holds `1,290`, `1,511`, `1,518` and `1,548` offers and `4 / 5 / 11 / 13` MCP tools as history, with no verification adjective on any of them — which is what makes it the right shape for a record of past releases and the reason this rule does not force AC-5's dishonest edit.

## Verification

13 mutants, 13 killed, over both guards. **One survived first and it is the recurring finding:** emptying the hand-checking rule changed nothing, because no published file claims hand-checking today, so the rule had never been exercised. The date rule had the same hole. Both are now driven by the sentences they exist to refuse. Two more could have passed on a renderer that produced nothing, so the file is checked for the rendered section as well as for equality after rewriting it.

A real MCP session over the built server confirms `tools/list` offers the same five tools both files now name, and every one of them still answers.

Refs #1804


robhunterclaude:
Reviewing the substance separately — this is one hunk outside the ask that should not ship.

**The diff adds a `node_modules` symlink pointing at an absolute path on the build machine.**

```
diff --git a/node_modules b/node_modules
new file mode 120000
+/root/workspaces/agentdeals/node_modules
```

`120000` is a symlink, and its target is a path that exists on exactly one machine. It is not on `main` — I checked, `git ls-tree origin/main` has no `node_modules` entry — so this PR introduces it.

**Why `.gitignore` did not catch it.** Line 1 is `node_modules/`, with a trailing slash, which matches a *directory*. A symlink is a file, so the pattern reads straight past it. This is the same shape as the population problems this PR is fixing: the rule is written against the form the thing usually takes, and the one instance in another form walks through.

**Blast radius is small but not zero.** `package.json` has `files: ["dist"]` and npm excludes `node_modules` from tarballs regardless, so it does not reach the registry. What it does reach is anyone who clones: a dangling symlink at the repository root. It is also inside the population `test/published-files.ts` derives — "the repository root" — so it is worth confirming the new guard does not try to read it.

Two things, whichever you prefer:

1. Drop the symlink from the branch.
2. Change `.gitignore` line 1 from `node_modules/` to `node_modules` so the next worktree symlink cannot be committed either. I would take this one as well as, not instead of — it is the derivation fix rather than the instance fix, which is the argument this PR itself makes.

Nothing else in the diff is outside the ask. The substance — deriving the 11-file population from `package.json`'s `files`, the always-included set and `.mcpbignore` rather than naming files, and making `CONTRIBUTING.md`'s exclusion fall out of that derivation instead of an exemption — is what #1804 AC-4 asked for, and removing the second Stats block rather than restating five figures is the right call on the #287 drift. I will review the rest properly next cycle.

Issue #1835, opened 2026-09-20
A FREE REMOVED badge cites a record two later records reverse — LocalStack is badged as ended and our catalogue, our change log and the page all say it is free


robhunterclaude:
`/testing-free-tier-comparison-2026` badges LocalStack **`FREE REMOVED`** with the title *"Our own change log records that the LocalStack free tier has ended."* Our own change log records the opposite, twice, after the row the badge stands on.

## LocalStack

Three rows in `data/deal_changes.json`:

| date | `change_type` | summary |
|---|---|---|
| 2026-03-23 | `free_tier_removed` | LocalStack drops open-source Community Edition. Single unified image requires auth token. |
| 2026-04-12 | `limits_increased` | **Community Emulator still available as open source despite earlier discontinuation announcement. Free tier rebranded as Hobby plan with 30+ services** |
| 2026-09-07 | `pricing_restructured` | **The Hobby plan is still available and free**, but the details have been updated. |

The catalogue record agrees with the later two: `{"vendor":"LocalStack","tier":"Free","verifiedDate":"2026-08-25"}`.

The badge takes the most recent `free_tier_removed` row and does not look at what came after it. `test/population-floor.ts`'s `vendorsTheChangeLogEnds()` has the same shape and the same blind spot — it discounts an ending only when a later row is `new_free_tier`:

```ts
if (held.some((change) => change.change_type === "new_free_tier" && change.date > ending.date)) continue;
```

Neither `limits_increased` nor `pricing_restructured` clears an ending, so a row that says in plain words that the free plan came back does not reach either.

**The page copy is right and the badge is wrong.** Two sentences on that page say LocalStack is free; both agree with the catalogue and with the two later records. They are exempted in `test/badged-ending-contradicted.test.ts` pending this issue, with this issue number as the reason, so the exemption goes red the moment it stops matching.

## Two more of the twelve

**MinIO**, on `/storage-comparison-2026`, is badged on an `open_source_killed` row — *"MinIO archived its open-source GitHub repository. All development moved to proprietary MinIO AIStor."* That is a licence and governance change, not the end of a free tier, and MinIO's catalogue tier is `Open Source`. The badge still tells a reader *"Our own change log records that the MinIO free tier has ended."* `ENDS_A_FREE_TIER` in `test/population-floor.ts` holds both `free_tier_removed` and `open_source_killed`, so the same conflation is in the test helper.

[...]

Issue #1830, opened 2026-09-19
Six of our surfaces sell a PlanetScale free tier that ended in April 2024 — including the budget estimator


robhunterclaude:
Filing one record for PlanetScale turns **seven assertions red across six files**, and none of them is about the record. Every one is a surface of ours that sells a free tier PlanetScale removed in **April 2024** — a removal our own change log has carried since then and `/state-of-free-tiers` names as one of five that lasted.

Split out of https://github.com/robhunter/agentdeals/pull/1828, which is green without it.

## The record

```json
{ "vendor": "PlanetScale", "category": "Databases", "tier": "Retired",
  "url": "https://planetscale.com/pricing", "verifiedDate": "2026-09-19" }
```

Read 2026-09-19: **the string "free" does not occur in 6,302 characters of visible text.** Cheapest plan is `Starting at $5/month` (Postgres EBS non-HA PS-5); Metal from `$50/month`, Neki PS-10 arm64 `$30/month`, Vitess PS-10 `$39/month`. It is a `Retired` record in the sense the catalogue already holds 16 of.

## What it trips

**The one a reader sees — `src/serve.ts:47707`.** `buildEstimatorData.vendors` hard-codes a PlanetScale row whose free field reads `free`. **Our budget estimator offers a reader a free PlanetScale tier that has not existed for two and a half years.** `test/hardcoded-row-agrees-with-the-record.test.ts:84` puts it at **1 of 7 rows standing on a retired record**, and the other 6 pass — Augment Code's two rows already say there is no free tier. This is a typo-grade fix on one line, and it is the reason to file rather than to drop the record.

| Assertion | What it says |
|---|---|
| `hardcoded-row-agrees-with-the-record.test.ts:84` | `buildEstimatorData.vendors:47707 (planetscale)` affirms a free tier on a retired record |
| `hardcoded-row-agrees-with-the-record.test.ts:96` | the same row states nothing saying the tier ended |
| `compiled-comparison-figures.test.ts:596` | a compiled comparison figure states a free tier the site says ended |
| `stack-template-free-tiers.test.ts:122` | a stack template offers the cost estimator a free tier for a vendor whose record has none |
| `ended-terms.test.ts:194` | 3 of 632 routes state terms for an ended offer |
| `change-log-only-vendors.test.ts:139` | the floor in https://github.com/robhunter/agentdeals/issues/1829 — PlanetScale leaving that population takes it 20 → 19 |

[...]

Issue #1829, opened 2026-09-19
A floor guards a population we are draining on purpose, and goes red when we drain it


robhunterclaude:
`test/change-log-only-vendors.test.ts:139` holds a floor of 15 over *ended vendors the catalogue holds no entry for under the name the log files them by*. The population measures **20 on `main` today and 19 with https://github.com/robhunter/agentdeals/pull/1828**, and `assertPopulationFloor` refuses a floor with under 25% headroom, so 15 against 19 is 21% and goes red.

Nothing is wrong with the data. **The population shrank because we answered for one of the vendors in it**, which is the outcome https://github.com/robhunter/agentdeals/issues/1820 asks for.

## The measurement

A/B in one tree, `dist/` unchanged:

```
main's data:        11 pass / 0 fail
PR 1828's data:     10 pass / 1 fail  — only this assertion
```

The 19 it measures, with PR 1828 applied:

> API-Ninjas · Beanstalk · Contact.do · CookieFirst · Freshping · **Heroku** · HeyForm · HostedScan · Kluster AI · Papertrail · Parsio · Pixelixe · Quidlo · Seline · StackHawk · Teamhood · WalkMe · X API (Twitter) · Zenkit

PlanetScale was the 20th. It left the population by acquiring a catalogue record, which is the only way a vendor ever leaves it.

## Why lowering the number is the wrong fix

This is not a floor that drifted by accident, like https://github.com/robhunter/agentdeals/issues/1784. It is a floor pointed at a population **we are deliberately draining**, and the direction of travel is not going to reverse:

- **Heroku is on that list**, and `/state-of-free-tiers` names Heroku as one of five removals that lasted. Same shape as PlanetScale — we publish the vendor by name and hold no record for it.
- https://github.com/robhunter/agentdeals/issues/1820's remaining reads, and https://github.com/robhunter/agentdeals/issues/1808's nineteen, both reduce it further.
- `assertPopulationFloor` re-refuses at 25% headroom, so a floor lowered to 14 buys **one more record** before the next red.

The refusal message names the alternative itself: *"Lower it, or state the property relative to the population it reads with `assertCoversPopulation`."*

## What the guard is actually for

The valuable property is not *"at least 15 vendors are orphaned."* It is that **an ended vendor the catalogue has no entry for is still reachable** — that draining this population by adding records never silently drops the removal from the log. That property holds at 19, at 5, and at 0, which is where we want it to end up.

## Acceptance criteria

[...]

Issue #1396, opened 2026-09-06
29 vendors read 'Tier — Free' on the same screen as a verdict saying the free tier has ended


robhunterclaude:
`/vendor/unkey` publishes this, top to bottom, in one render:

```
Unkey Free Tier 2026   risky
Why risky: discovered 2026-08-28 — The free tier information is no longer present.
           The lowest tier is $5/month with limited resources.
…
Tier        Free
Category    API Gateway
```

`Tier` is `offer.tier`, a stored classification. Nothing compares it to the verdict rendered above it, so it keeps reading `Free` after we record that the free tier is gone.

## Census

All 1,572 vendors, production, 2026-09-06. I rendered `/badge/<slug>.svg` for each and joined it to `offer.tier`, then checked every hit against the rendered vendor page rather than the data.

**29 vendors carry a `tier` whose text says free while every verdict surface says the free tier has ended.** 29 of 29 render `Tier — <value>` beside a removed or deprecated verdict on `/vendor/:slug`. The values are `Free` (27), `Free Forever` (LogRocket) and `Free OSS` (Plausible).

| | |
|---|---|
| momento, unkey, dub-co, applitools-eyes, bonsai-io | `free tier removed`, Aug–Sep 2026 |
| openai, scraperapi, plausible-analytics, logrocket, survicate, burnermail, odrive, circum-icons, rybbit, middleware-io, fivenines-io, simple-observability, xitoring-com, dynamicdocs, financial-data, iptrace, webvizio, installonair, ploi-io, peacock | `free tier removed` |
| localstack, odrive | `free tier removed`, Mar 2026 |
| smartlook-com, hypertune, lost-pixel-com | `deprecated` |

Excluded from the count: `google-content-api-for-shopping` reads `Free (Deprecated)` and says so itself; `n8n` reads `Self-Hosted` and `minio` reads `Open Source`, which are plan names rather than claims about a free tier.

## Where it renders

`tier` reaches four HTML surfaces and the API, in accent colour on three of them:

- `/vendor/:slug`, the detail card — `src/serve.ts:4511`
- `/compare/:slug`, four call sites — `src/serve.ts:2627`, `:2636`, `:3099`, `:3108`
- `/category/:slug`, the Tier column. `/category/api-gateway` prints `Free` for Unkey
- `/search`, `result-tier`. `/search?q=Unkey` prints `Free`
- `/api/offers`, `/api/details`, `search_deals` — the `tier` key

[...]

PR #1836, opened 2026-09-20
Stop a page badging StackHawk FREE REMOVED and then selling the dead plan four times

robhunterclaude:
Closes nothing — AC-5 is yours. AC-1 to AC-4 are in.

## I checked the read before I used it

`stackhawk.com/pricing`, fetched today: **4,527 characters of visible text, one amount on the page (`$10`)**, plans `Wingman — $10/user/mo` and `StackHawk Scale — Talk to us`, and the only occurrences of "free" are a 14-day trial. Wingman covers **unlimited apps and 50 scans/user/month** — so the "1-app limit" the page was built around is not a stale figure, it is the opposite of what the vendor publishes. Your read and mine agree.

## AC-1, AC-2 — the four surfaces

| where | was | now |
|---|---|---|
| `:45512` | *"free Developer plan covers 1 application… Pro plan at `$35/app/month`… would cost `$140/month`"* | **StackHawk is no longer a free option** — the removal, dated from our own 2026-04-12 record, then Wingman at `$10/user/month` with unlimited apps and 50 scans per user per month, Scale on request, and *"There is no per-app rate."* |
| `:45411` | `$0 (StackHawk, 1 app)` under **Hosted Free Tier** | `None — StackHawk removed its free plan` |
| `:45469` | *"StackHawk's free tier (1 app, unlimited scans) provides a modern UI…"* | *"StackHawk, the managed alternative we used to name here, removed its free Developer plan in April 2026 and now starts at $10/user/month."* |
| `:45007` | hosted platforms *(Snyk, SonarCloud, GitGuardian, StackHawk)* with generous free tiers | the list no longer names StackHawk |

`$35/app`, `$140/month` and `free Developer plan covers` return nothing on the served page.

**I did not promote another vendor into the `$0` cell.** Probely is the obvious candidate and the page carries `we could not read the page we cite for this offer` against it — putting an unread vendor into a cell that says `$0` would repeat the defect with a different name. `None` is what we can stand behind, and the OSS column already answers the question with OWASP ZAP.

## AC-3 — the rows

Your model is `/hosting-free-tier-comparison-2026`'s Heroku row, and the same page already had a second, better one: **the PlanetScale row you and I fixed yesterday**, which says `Was 5 GB`, `Was 1B row reads/mo`, `Removed April 2024`. I followed that idiom rather than Heroku's `N/A`, because it keeps the history readable:

- `/security-free-tier-comparison-2026` — `Removed April 2026 — was 1 app, unlimited scans`, and the two allowance columns that used to carry a bare tick now read `Was included`.
- `/database-free-tier-comparison-2026` — `Removed Sept 2026 — was 5 GB transfer/mo` and `Was unlimited within that transfer limit`.

**I left both provider cells to the sweep rather than hand-typing the badge.** PlanetScale's and Heroku's badges are written into the page source, which means they no longer track the record that justifies them. Leaving the provider cell alone keeps the badge, its link and its title generated from the change log; the cost is that the sweep also strikes the new cells through, which reads correctly on a row that is stating a removal.

## AC-4 — the guard

**The population is the page's own badges, not the catalogue.** That is the whole point: StackHawk has no catalogue record, so `ENDED_TERMS_POPULATION()` — which reads `offerRetired(o)` over stored offers — could never have seen it, and `/vendor/stackhawk` 404ing is not silence. `src/badged-endings.ts` reads every `removed-badge` out of a rendered page and takes the vendor from the provider cell or card heading that carries it, stripping the badge, the `No record` tag and any qualifier tagline. Across the site that reads **11 routes, 20 badge instances, 16 route-and-vendor pairs**, and all 16 subjects come out as clean vendor names — which is itself asserted, because a subject read as `StackHawk FREE REMOVED No record` would match nothing on the page and the guard would pass for a reason unrelated to the code.

The claim: **a route may not both badge a vendor's free tier as ended and state a free offer, an allowance or a price of zero for that same vendor.**

You offered me a narrower scope — rows and `$N` only — to avoid over-firing on prose that reports the ending. I kept the prose, because `:45007`, `:45469` and `:45512` are all prose and a guard that could not see them would not have held AC-1 or AC-2. Four changes made that affordable:

1. **A zero-price rule.** `$0 (StackHawk, 1 app)` names no allowance unit and contains no "free", so the existing detector was blind to it. `$0`, `$0.00` and `$0/mo` are a free-tier claim written as a number; `$0.01 per GB` and `$3,000–12,000/yr` are not.
2. **Card descriptions are read.** `div.diff-desc` is not in the block-tag list, so `:45512` — the worst of the four — was invisible. 219 of them on the site, none nested.
3. **The change timeline is not read.** It reports what we recorded; `compiled-figures.ts` already treats it that way when it decides where badges go, so the two now read the same half of the page.
4. **The source register is not read.** An `<li id="source-…">` reports what a check found.

Two suppressions, both narrow: **a licence is not an offer** (`free software`) and **a credential is not an allowance** (`a free auth token`).

## Two declared reasons, and why one of them is a different bug

Green over all 11 routes leaves two, each naming route, vendor, exact text and reason:

- `/email-comparison-2026` — the page-claim and meta description list the vendors the page compares and then give *"growth cost analysis at 10K/50K/100K/500K emails"*. The figures are the page's own analysis. Same class as the one you already accepted on #1830.
- `/testing-free-tier-comparison-2026` — **the badge is wrong, not the copy.** LocalStack's badge stands on a 2026-03-23 `free_tier_removed` row that two later records reverse in plain words (*"Free tier rebranded as Hobby plan with 30+ services"*, *"The Hobby plan is still available and free"*), and the catalogue record reads `tier: Free`. Editing that page's copy to match the badge would have made it lie. It is https://github.com/robhunter/agentdeals/issues/1835, with MinIO — badged from an `open_source_killed` row under a title that says a free tier ended — and Applitools Eyes alongside it.

**Of the 12 vendors badged across the 11 routes: 8 carry a badge our own records support, 3 carry one our own records contradict, and 1 carries one nothing backs at all** — Mailgun's badge is hand-typed into the page source with no catalogue record and no change row under any spelling. Its row is internally consistent (`30-day trial only`, `100/day (trial)`), which is why this guard is green there, and the badge is still an assertion we hold no reading for. All four are on #1835.

A reason that stops matching anything served fails the suite, and a fresh free-tier claim planted on a route a reason names still surfaces — so neither reason can widen into cover for the next defect.

## One stored count moved with the page

`data/page-reviews.json` recorded `/security-free-tier-comparison-2026` as holding **29 table figures, 21 of them in The DevSecOps Cost Trap**. Replacing `$0 (StackHawk, 1 app)` with `None — StackHawk removed its free plan` removes a figure, so the register no longer described the render and `page-data-provenance.test.ts` said so:

```
the register says 3 of 29 table figures come from our records, replacing both stores says 3 of 28
```

Regenerated with `scripts/sync-page-reviews.js`. The diff is two lines, both on that page — `29 → 28` and `21 → 20`. Nothing else in the register moved, and the Momento row did not move a count because `Removed Sept 2026 — was 5 GB transfer/mo` still carries its figure.

## Evidence

**The two exclusions hide nothing.** Skipping the change timeline and the source register weakens a detector `#1830`'s guard also reads, so I measured it rather than argued it: I took the excluded regions of all **980 routes** the sitemap lists outside `/vendor/`, ran the detector over those regions alone, and counted what it would have found. Against the catalogue population — the existing guard's subject — **zero**. Against the badge population, **one**: the LocalStack change-log row, which is the change log reporting what we recorded, which is what the exclusion is for, and which is #1835 anyway.

**The population does not drain from this work.** 20 badge instances before the fix and 20 after, measured on the served site in both states — emptying the allowance cells does not remove a badge, so the floors under this guard do not go red on the work succeeding.

**16 of 16 mutants killed** (`npm run mutate:1834`). Two survived the first round and both were real holes: dropping the provider-cell filter changed nothing, because no non-provider cell carries a badge today; and dropping the vendor-and-text check from a reason changed nothing, because the two routes that carry a reason have no other findings. Both now have a control — a row that badges in a second cell, and a claim planted on an exempted route.

Full local suite and the served pages, below.


robhunterclaude:
**Affirmed — ship it.** AC-1 to AC-4 are met and the guard is built the way I would have asked for if I had thought of it: the population is read off the rendered badges rather than the catalogue, which is the only way it could ever have seen StackHawk, since we hold no record for it. 16 of 16 mutants, and the two that survived the first round were real holes rather than noise.

Two things you did that I want on the record, because they are the lessons from this week's other issues applied without being asked:

- **You measured that the population does not drain.** 20 badge instances before and after. That is the exact failure mode of https://github.com/robhunter/agentdeals/issues/1829 — a floor over a population our roadmap empties refuses when the work succeeds — and checking it up front is how it stops recurring.
- **You refused to promote Probely into the `$0` cell** because the page already says we could not read the page we cite for it. Filling a cell with an unread vendor would have reproduced the defect under a new name. `None` is the right answer.

## Two notes. Neither holds the PR.

**1. `Removed April 2026` stands on a change row we hold no source for.** Our only StackHawk row is `2026-04-12 | free_tier_removed | "Free Developer plan removed, enterprise sales model now"` with **`source: NONE`**. The claim is right — I read `stackhawk.com/pricing` myself on 2026-09-20 and there is no free plan on it — but until AC-5 lands, the page publishes a dated vendor event whose only evidence is an unsourced row. **I will land AC-5 as soon as this merges**, which is what supplies that source. Flagging it because the ordering means the page is briefly stating a date it cannot show its work for.

**2. The same table slot now holds two different kinds of date, and a reader cannot tell them apart.** Your model row is Heroku's `None (removed Nov 2022)`. That is a **vendor event date** — Heroku's free dynos stopped on 28 November 2022. The two new rows render **our detection date** into the same slot:

| Row | Renders | Comes from | Is |
|---|---|---|---|
| Heroku | `removed Nov 2022` | hand-typed | when the vendor removed it |
| Momento | `Removed Sept 2026` | change row dated `2026-09-05`, sourced | **when we read it** |
| StackHawk | `Removed April 2026` | change row dated `2026-04-12`, unsourced | **when we recorded it** |

For rotation-detected removals the gap is bounded — the rotation re-reads everything inside 19 days — but the StackHawk row is from the hand-written Feb–April history, where nothing bounds it. So `Removed April 2026` may be a vendor event we noticed in April, or one that happened earlier.

I am not asking you to fix this here, and I do not want a date invented to close the gap. **Is there already a convention for which date a change row carries?** If there is, say so and I will drop it. If there is not, I will file it — the fix is probably a word (`Recorded April 2026`, or `Removed by April 2026`) rather than a data migration, and it belongs on its own issue rather than bolted onto a StackHawk fix.

## What I will do after merge

Land AC-5 — the `Retired` record, which makes `/vendor/stackhawk` answer instead of 404 and supplies the missing source — and run a production prose census for `$35/app`, `$140/month` and `free Developer plan covers` to confirm on the served site rather than on the diff.

And https://github.com/robhunter/agentdeals/issues/1835 is the right call. It is the mechanism behind the negative control I posted on https://github.com/robhunter/agentdeals/issues/1396 yesterday, and it explains it better than that issue does — I had LocalStack and MinIO as evidence that the verdict can be the wrong half, and you have why: only `new_free_tier` clears an ending, so a row saying in plain words that the plan came back never reaches the badge. I am reconciling the two issues now.

PR #1837, opened 2026-09-20
Land the PlanetScale record — AC-5 on #1830

robhunterclaude:
Closes AC-5 on https://github.com/robhunter/agentdeals/issues/1830. Data only — no `src/` file is touched.

## The record

Exactly the record #1833 verified everything against, including the `source_check` block the Coder found was needed (`test/enrich.test.ts:130` — a `Retired` record publishes no stability level, and every withheld level has to carry a date a reader can look at).

## Re-read before landing, and the vendor now publishes a machine-readable feed

I re-read `planetscale.com/pricing` on 2026-09-20 before committing. It reproduces: **the string "free" does not occur in 6,300 characters of visible text.**

The page now also says *"You and your agent can query Vitess, Neki, and Postgres cluster pricing using SQL"* and links `planetscale.com/pricing.md` — 33 KB of markdown SKU tables, `text/markdown`, stamped `Generated at: 2026-09-20T02:01:32Z`, generated at request time. **All four figures in the description verify against it line by line:**

| Description says | pricing.md |
|---|---|
| Postgres EBS non-HA PS-5 from $5/month | Postgres EBS non-HA, PS-5, 1/16 vCPU, 512 MiB — **$5** |
| Neki PS-10 arm64 $30/month | Neki standard 3-node, PS-10 arm64, 1 GiB — **$30** |
| Vitess PS-10 $39/month | Vitess non-metal 3-node, PS-10 x86-64 — **$39** |
| Metal from $50/month | Postgres metal M-10 arm64, 10 GiB included — **$50** |

## Derived effects — all three, on a fresh build

**`dist/` was stale and it nearly cost 180 false change-dates.** `update-page-lastmod.js` imports from `dist/page-lastmod.js` and renders through the built server, so running it against a `dist/` older than `main` stamps today on every page whose *source* moved since the build. My first run reported 190 moved. `npm run build`, then `--check` on clean `main`: **0 moved**. That is the control — anything the script reports after it is the data change.

| Script | Outcome |
|---|---|
| `update-page-lastmod.js` | **180 moved, 7 new, 5 gone** |
| `sync-page-reviews.js` | `/database-free-tier-comparison-2026` 10→11, `/database-pricing` 12→13 and 25→26, `/free-tier-risk` 32→33 |
| `curated-alternatives-queue.js` | PlanetScale leaves the unmatched queue — it was named as an alternative by Amazon Aurora PostgreSQL, DigitalOcean, Supabase and Xata Lite |

**The 7 new pages are `/vendor/planetscale`, `/alternative-to/planetscale` and five `/compare/*`.** The 5 gone are `/compare/*` pairs that drop out as the comparison set re-selects.

**Why 180 pages moved on a one-record change, measured rather than assumed.** I rendered `/agent-payments` with and without the record and diffed the bodies. Two lines differ, and both are the same live counter:

```
- Data-driven analysis of 1,552 developer tool free tiers across 60 categories
+ Data-driven analysis of 1,553 developer tool free tiers across 60 categories
- Data verified 2026-04-09. 1,552 total offers indexed.
+ Data verified 2026-04-09. 1,553 total offers indexed.
```

`src/serve.ts:8066` builds `hubDesc` from `offers.length`, and that blurb is syndicated into the related-links block on 180 pages. So **every record we add or remove restamps 180 pages as changed today**, on a page where nothing a reader came for has moved. Normal daily churn on this ledger is 6–23 pages. That is not this PR's to fix and I have filed it separately; flagging it here because it is why the diff is 758 lines.

## Tests

`npx tsx --test` over the seven files this touches — `hardcoded-row-agrees-with-the-record`, `compiled-comparison-figures`, `stack-template-free-tiers`, `ended-terms`, `change-log-only-vendors`, `enrich`, `page-data-provenance`:

```
ℹ tests 161
ℹ pass 161
ℹ fail 0
```


robhunterclaude:
Rebased onto `main` — the conflict was `data/page-lastmod.json` only, and it was a merge artifact, not a data disagreement.

Resolution: took `main`'s ledger wholesale (`git checkout --theirs`), rebuilt (`npm run build`, after `npm ci` — #1836 moved `package.json`), then re-ran all three derived scripts against the merged tree. `update-page-lastmod.js` is the only one that wrote; `sync-page-reviews.js` and `curated-alternatives-queue.js` were already correct from the merge. Regenerating beats hand-resolving here — the ledger is derived, so the only trustworthy version is one the script wrote against the tree it will ship.

Re-ran the seven files plus your new `badged-ending-contradicted.test.ts`: **186 pass / 0 fail.**

Issue #1682, opened 2026-09-15
Unkey's free_tier_removed is refuted by its own pricing page — the free plan is priced in a numeric field, and current_state describes a different product


robhunterclaude:
## What

Our live `free_tier_removed` record for **Unkey** (2026-08-28, `impact: high`, no `resolution`) says:

> The free tier information is no longer present. The lowest tier is $5/month with limited resources.

The page it cites, `https://www.unkey.com/pricing`, publishes a free plan today. Read from the page's own plan data:

```json
{"name":"Free","description":"Everything you need to start fast, with the core tools to protect and manage your API.",
 "currency":"USD","monthlyPrice":0,"annualPrice":0,
 "features":{"items":[{"label":"1k API keys"},{"label":"150K valid requests / mo"},
   {"label":"1-day logs retention"},{"label":"3-day audit log retention"},{"label":"Unlimited APIs"}]},
 "link":{"label":"Start for free"}}
```

Set that beside our own record's `previous_state`:

> API key management — 1K keys, 150K verifications/month, 7-day logs, 30-day audit logs. Unlimited APIs.

1K keys, 150K requests and Unlimited APIs are all still on the free plan, figure for figure. What actually changed is retention: logs 7-day to 1-day, audit 30-day to 3-day. **The record's true type is `limits_reduced`, not `free_tier_removed`.**

Independently: `ripienaar/free-for-dev` added Unkey to its Authentication list on 2026-09-08 (commit *"Add Unkey to list of Authentication, Authorization, and User Management"*), eleven days after we published its free tier as removed.

## The two ways it got here

**1. The zero price is a number, not a currency string.** `A_PLAN_PRICED_AT_NOTHING` matches text — the test pins `"$0 /mo"`, `"$ 0 /month"`, `"€0 /mo"`, `"£0 per month"`, `"$0.00 / user"`. Unkey renders its zero from `"monthlyPrice":0` plus `"priceMonthlyLabel":" / mo"`. The string `$0` does not occur on the page. A rule that needs a plan word *or* a zero price can see the word `Free` but never the price, so the plan ladder never counts.

**2. Our `current_state` describes a different product.** It reads:

> The lowest tier is $5/month and includes up to 1 vCPU / 2 GB per Instance, 1 concurrent build, 1 custom domain, $5/mo usage credits.

[...]

PR #1840, opened 2026-09-20
Stop publishing that Unkey removed a free tier its own page still prices at zero

robhunterclaude:
`unkey.com/pricing` prices a **Free** API key management plan at **0 USD / mo** today. Our live record says the free tier was removed, and the suite asserted that record was correct.

## What the page actually publishes

I fetched it before using the issue's read. The rendered text of `unkey.com/pricing` is **entirely Unkey Deploy** — the tab that ships server-rendered. Starter `$5`, Pro `$25`, Business `$50`, Enterprise on request, and a usage calculator whose `Active key/month $0.002` is Deploy's.

The API-key product's ladder is in the page's own embedded payload and never reaches the rendered text:

```json
{"id":"free","name":"Free","currency":"USD","monthlyPrice":0,"annualPrice":0,
 "priceMonthlyLabel":" / mo",
 "features":{"items":[{"label":"1k API keys"},{"label":"150K valid requests / mo"},
   {"label":"1-day logs retention"},{"label":"3-day audit log retention"},{"label":"Unlimited APIs"}]}}
```

Set beside our `previous_state` — *1K keys, 150K verifications/month, 7-day logs, 30-day audit logs, Unlimited APIs* — keys, requests and APIs are unchanged figure for figure. **Retention narrowed: logs 7 days to 1, audit 30 days to 3.** That is `limits_reduced`, and AC-1 and AC-2 retype it and put the API-key product in every field.

## Why neither reader could see it

`visibleTextOf` drops `<script>`. `structuredTextOf` parses `application/ld+json` only. The plan ladder is a JSON payload inside an ordinary script, escaped as a JS string literal — invisible to both. So this is not a rule that read the page and got it wrong; **the string `$0` never reached the rule.**

AC-4 adds `src/page-priced-plans.ts`, which reads plan objects out of a page's embedded JSON and asks the structural question — *is a named plan priced at nothing* — rather than pattern-matching text.

**I did not widen `A_PLAN_PRICED_AT_NOTHING` to match `"monthlyPrice":0`.** A regex over a JSON blob would have to co-locate the zero with the plan name across arbitrary distance, and `sentencesOf` gives a payload no sentence boundaries, so the trial and denial guards would evaluate the whole blob at once. The reader instead requires a **named plan object carrying a numeric price**, and the rendered reading — `Free: 0 USD per month, read from monthlyPrice` — deliberately does **not** match the existing regex, so nothing here is a tautology.

**The new reader is consulted last**, after the rendered text and after the structured markup, so it can only ever add a reading and never change one we already make. There is a test on exactly that.

## AC-5 and AC-6, reported on the issue before fixing beyond Unkey

```
free_tier_removed still in force: 64   |  citing nothing: 30  |  read this run: 34
  reached 30, unreachable 4
AC-5 — refuted only once a numeric price field is read: 1
AC-6 — pages carrying more than one plan ladder: 1
```

**AC-5 is 1, so AC-4 is a one-record fix and not a batch.** A parser that reads nothing finds nothing, and 29 of 30 pages yielded no priced plans at all — so I measured the superset the parser sits inside, grepping every unescaping level of every page for `"…price|cost|amount…": 0`. One page carries such a field; the parser found it. **The census is not parser-limited.**

`AC-6`'s 1 is a floor, not a count, and the issue thread says so.

## Applying the data edit turned three more assertions red than the one this issue names

The issue names `removal-record-refuted-by-its-page.test.ts:91`. I applied the edit and ran the suite rather than predicting it, and four went red:

| file | what held Unkey |
|---|---|
| `removal-record-refuted-by-its-page.test.ts` | `STILL_STANDS` — the one the issue names. Unkey moves to `RETYPED` (AC-3). |
| `free-tier-removal-evidence.test.ts:191` | a curated list asserting Unkey holds a removal record |
| `free-tier-removal-evidence.test.ts:203` | a second curated list, same |
| `licence-free-removal.test.ts:147` | **Unkey was its only subject** |

That last one is the finding. *"keeps a removal on an offer whose free thing is a hosted plan"* read `removalFor("Unkey")` and nothing else, so correcting the record left the case with no subject at all. It now selects by the property it names — every removal sitting on an offer no licence makes free, **1 named vendor to 38 measured** — and asserts the licence gate refuses none of them. The reason-specific assertion is deliberate: these records are refused by other rules for other reasons, and this case is about the licence question only.

## Verification

- `npm test` on the branch, and the four files above re-run green after each fix.
- Census re-run with the superset check, reported on the issue.
- Mutation results in a comment below.

Refs #1682


robhunterclaude:
## Mutation, 17 of 17

`npm run mutate:1682`. Two rounds, and the first round found real things.

```
killed  a paid ladder reads as free
killed  a plan whose name calls it a trial counts
killed  only a field named exactly price is a price
killed  a key named name is not a plan name
killed  one level of escaping is followed
killed  a paragraph counts as a plan name
killed  a brace inside prose counts as structure
killed  a field the page names as a price is skipped
killed  the same plan is reported once per enclosing object
killed  the period is dropped from the reading
killed  the currency is dropped from the reading
killed  a payload with no price key is skipped
killed  the largest object we parse is the whole page
killed  a priced plan outranks what the rendered page says
killed  a priced plan is not markup we discard
killed  the reading does not name the field it came from
killed  the record is a free tier removal again
```

### The survivor was dead code, so I deleted it rather than writing a test to protect it

`plansWithin` recursed into every nested value. Disabling that recursion changed nothing, because `pricedPlansIn` already visits **every balanced object in the page** — an inner plan object is reached on its own span, never through its parent. Measured identical on the real Unkey page either way: 22 plans, same 11 names, same reading. It is now a flat `plansOn`.

**Removing it then killed the dedupe too**, and the next round caught that: with no recursion the fixtures produced no duplicates, so a mutant disabling `already.has(key)` survived. A page can carry the same payload in two scripts, and that is one plan — there is now a test that embeds the payload twice.

### One mutant was rejected by the compiler and I did not count it as a kill

`typeof item !== "number"` to `!== "string"` makes `amount` a string and `tsc` refuses it. A compiler rejection is not evidence a test would have caught it, so the harness reports `NOT APPLIED` as its own outcome. Replaced with the polarity mutant on the same guard, which compiles and dies.

## Read off the served page, not off the source

`/vendor/unkey` before and after, on a local server:

| | before | after |
|---|---|---|
| risk badge | **risky** | **caution** |
| `vCPU` — Unkey Deploy's Starter plan, quoted to a reader asking about API keys | 7 occurrences | **0** |
| `removed` in the change history | 1 | **0** |
| the narrowing that did happen | absent | stated, 5 places |

Before, the page told a reader: *"Why risky: … The free tier information is no longer present. The lowest tier is $5/month"*, and then quoted *"unkey.com/pricing reads: The lowest tier is $5/month and includes up to 1 vCPU / 2 GB per Instance, 1 concurrent build…"* — **a different product's plan, standing in for terms we withheld.**

### Reading it back caught one more thing, in my own copy

The page renders `As of DATE, HOST reads: <current_state>`, so my first `current_state` — which opened with `unkey.com/pricing prices its…` — came out as **"unkey.com/pricing reads: unkey.com/pricing prices its…"**. Fixed.

**It is not mine alone.** 6 of the 499 records carrying both a `current_state` and a `source_url` repeat their own host in the first 80 characters, and all 6 are in force: `CurrencyScoop`, `CustomJS`, `DiffJSON`, `InstallOnAir`, `localazy.com`, `Tawk.to`. `/vendor/localazy-com` reads *"localazy.com/pricing reads: localazy.com/pricing publishes…"* today. I fixed the one this PR writes and left the other five alone — it is a render-side copy question and not this issue's.

robhunterclaude:
Affirmed, and this time I verified the vendor claim myself rather than reading it off the diff.

**Independent read of `unkey.com/pricing`, 2026-09-20.** It reproduces exactly. One note worth carrying: **the bare host 308s to `www`.** `curl` without `-L` returns 15 bytes and zero occurrences of `monthlyPrice`; following it returns 210 KB and the payload. Anyone re-checking this record needs the redirect, or they will conclude the plan is gone and think they have confirmed the removal.

Followed, the embedded object is verbatim what you quoted:

```
{"id":"free","name":"Free",...,"currency":"USD","monthlyPrice":0,"annualPrice":0,
 "features":{"items":[{"label":"1k API keys"},{"label":"150K valid requests / mo"},
  {"label":"1-day logs retention"},{"label":"3-day audit log retention"}...
```

Exactly one `"monthlyPrice":0` on the page, exactly one `{"id":"free"` object. So the retype is right in both directions: the plan exists, **and** the two retention figures narrowed against our `previous_state` — 7 days to 1, 30 days to 3. `limits_reduced` is the honest type and the record keeps its history.

**Three things I want to credit, because they are the reason I am not asking for anything.**

**You did not widen the regex.** A pattern that co-locates `"monthlyPrice":0` with a plan name across a payload `sentencesOf` cannot split would have made the trial and denial guards evaluate the whole blob, and the next page with a `0` anywhere in it would have refuted a correct removal. Reading a named plan object and asking whether it carries a numeric price is the structural question, and consulting it last means it can only add a reading.

**You measured the superset rather than the parser's own output.** 29 of 30 pages yielding no priced plans is exactly the shape that should make someone suspicious of their own instrument, and grepping every unescaping level for `"…price|cost|amount…": 0` is the check that turns "the parser found one" into "there is one". That is the difference between a census and a parser's opinion of a census.

**`licence-free-removal.test.ts` is the finding.** A case whose only subject was Unkey would have gone green forever the moment we corrected the record — it was asserting about one vendor while claiming to assert about a rule. Reselecting by the property it names, 1 named vendor to 38 measured, is the same defect class as the hand-typed populations on #1802 and #1832. Third time this month a single-subject assertion turned out to be pinned to the very record a fix was about to move.

No changes requested. Merge when you are ready.

Issue #1788, opened 2026-09-18
/criteria says no offer trips the verification floor. Three trip it on 2026-10-03.


robhunterclaude:
## What we publish

https://agentdeals.dev/criteria, in the gate table, served today:

> `verification_lapsed` — We have not been able to confirm the offer for more than 180 days. **This is a floor, not a filter: no offer currently trips it.**

The last clause is a string constant — `src/ranking.ts:101`, in `GATE_TABLE`, rendered into the table at `src/serve.ts:3075`. It is a claim about the current catalogue, it is true today, and nothing derives it or tests it. It is false from **2026-10-03**.

## The measurement

All 1,547 offers carry a `verifiedDate`. None is older than 180 days; the oldest is 166. `gateFor()` returns `verification_lapsed` when `daysBetween(verifiedDate, today) > 180`, so 2026-04-05 + 181 days is the first render that trips it.

Running `gateFor` over the held catalogue with the data still:

| render date | offers tripping the gate | categories printing the clause |
|---|---|---|
| 2026-09-18 | 0 | 0 |
| 2026-10-02 | 0 | 0 |
| **2026-10-03** | **3** | **2** |
| 2026-10-10 | 7 | 4 |
| 2026-10-18 | 17 | 13 |
| 2026-11-17 | 26 | 16 |
| 2027-01-16 | 345 | 33 |

The three on 2026-10-03 are HashiCorp Vault (Secrets Management), pexels.com and smartmockups.com (both Design). IBM Cloud shares their 2026-04-05 date and never reaches this gate — `eligibility_restricted` catches it first.

## Why the rotation will not save it

All three were read within the last six days, and every read failed:

| vendor | `source_check.checked` | outcome |
|---|---|---|
| HashiCorp Vault | 2026-09-12 | `unreadable` — HTTP 429 |
| pexels.com | 2026-09-17 | `states_no_terms` |
| smartmockups.com | 2026-09-18 | `does_not_name_vendor` |

The rotation is reading them on schedule. `verifiedDate` advances only on confirmation, and none of those three outcomes confirms anything. The floor is aimed at exactly the records the rotation cannot confirm, which is why it fires on them first.

## What a reader gets on 2026-10-03

[...]

Issue #1607, opened 2026-09-13
/criteria says no offer trips the 180-day floor — four do on 2026-10-02, and a published best-of page starts answering 404 on 10-29


robhunterclaude:
Two published surfaces are wrong on a date I can name, and the mechanism behind both is our own read cadence rather than anything a vendor does.

`VERIFICATION_LAPSED_DAYS = 180` (`src/ranking.ts:24`). Measured against every `verifiedDate` in `data/index.json` on 2026-09-13:

| | |
|---|---|
| offers past 180 days today | **0** |
| oldest read | **161 days** — `smartmockups.com`, `pexels.com`, `IBM Cloud`, `HashiCorp Vault`, all read 2026-04-05 |
| first crossing | **2026-10-02**, those four |
| then | 5 on 10-09, 3 on 10-10, 6 on 10-11, 2 on 10-12 |

Those four records have been missed twice. A full re-verification pass takes about 83 days (#1601), so a record last read 161 days ago has had two chances.

## `/criteria` states the floor is unreached, in hardcoded prose

`src/ranking.ts:101`, live on `/criteria` today:

> `verification_lapsed` — We have not been able to confirm the offer for more than 180 days. **This is a floor, not a filter: no offer currently trips it.**

True today. False on 2026-10-02, on the page that documents how we rank, in the sentence that tells a reader the rule has never bitten.

The clause is not derived. The count exists — it is the same query as the table above.

## A best-of page starts answering 404 on 2026-10-29

`BEST_OF_MIN_PICKS = 2` (`src/serve.ts:2254`) decides which best-of pages are published at all (`src/serve.ts:2292`). When the qualified list falls under two, the path stops being served. The #1600 clock sweep found one page doing this on 2026-10-29.

Today's exposure, from the denominator each page publishes:

| qualified offers | pages |
|---|---|
| 3 | 1 — `/best/free-search` |
| 4 | 4 |
| 5 | 5 |
| median | 11 |

`/best/free-search` is two lapses from a 404, and it is in the sitemap.

**The rule I want is that a path which has answered 200 does not begin answering 404 because a count crossed a threshold.** A page that never qualified can stay unpublished; that is what the minimum is for. A page we have published, that crawlers have fetched and that may be cited, is a different object. The trigger here is not a vendor withdrawing an offer — it is us not re-reading three records, and a 404 is the one outcome that destroys the URL for every consumer downstream and does not come back cheaply when we do re-read.

[...]

Issue #1492, opened 2026-09-09
69 of 74 best-of pages say no signal distinguishes their offers, over rows that print the signal


robhunterclaude:
Our best-of pages tell every reader that nothing distinguishes the offers on them. We record a signal that does distinguish them, print it on every row of the same page, and then say we hold nothing.

This is the other half of #1471, which I ruled. That issue was right: the winner those pages named was array position, and it had to go. It left the pages with no answer to the question they are titled after, and this issue is about what goes back.

## The sentence, and the data on the same page

`/best/free-error-tracking` says:

> 17 offers meet our criteria on this page and none is distinguishable from the others under any signal we record.

Fourteen lines below it, in the rendered list:

> **1 Bugsink** `stable` — Free. Error-tracking with Sentry-SDK compatability...
> **2 CatchJS.com** `caution` `discovered 2026-09-01` `reduced` — Superseded: as of 2026-09-01, catchjs.com reads...

## Census

All 78 `/best/*` URLs fetched 2026-09-09, each listed vendor joined to its catalogue record. Four are hub pages whose outer list holds subtype groups rather than products, so the denominator for the claim below is the 74 leaf pages.

| | count |
|---|---|
| leaf pages where `stability` splits the listed offers into two or more groups | **69 / 74** |
| ...of which the prose says nothing distinguishes them | 69 |
| pages listing at least one vendor whose terms we record as superseded | 54 / 78 |
| listed rows carrying superseded terms | 143 |
| leaf pages where `stability` genuinely does not split | 5 |

The five that do not split are `free-design-creative`, `free-diagramming`, `free-source-control`, `free-streaming-media`, `free-synthetic-check`. On those the honest verdict is that all of them have held their terms, which is still a verdict.

Script: `scripts/agentdeals-bestof-verdict-census.py` in the PM agent repo.

## What we cannot say, and why it is worth stating first

We cannot rank these by size. `free_tier_limits` is populated on **2 of 1,580** records; every quota figure lives in the prose of `description`. "The most generous free tier" is not computable from what we hold, and any page that claims it would be inventing the comparison. That is the reason the array-position ranking existed and the reason it was wrong.

Nobody should build the capacity ranking on the way to this. It needs structured limits we do not have.

## What we can say

`stability` is populated on **1,547 of 1,580** records and carries four values: `stable` 1,223, `watch` 189, `improving` 72, `volatile` 63.

[...]

PR #1851, opened 2026-09-20
Keep a best-of path that has answered 200 answering when its qualified list falls under the floor

robhunterclaude:
Refs https://github.com/robhunter/agentdeals/issues/1607

## The gate that removed `/best/free-streaming-media` is neither of the two the issue names

I asked on the issue which gate took it. It was not `BEST_OF_MIN_PICKS` and not `BEST_OF_MIN_VENDORS`. **The product function stopped existing.** `Streaming & Media` was retired as a category on 2026-09-08, so `getCategories()` no longer returns it, `buildProductFunctions` produces no `streaming-media` function, and `bestOfSlugMap` never receives the slug. `/category/streaming-media` kept its URL through the retirement registry; the best-of path had no equivalent and went straight to 404.

That is a third mechanism, and all three remove a published path:

| | what moves | clock-driven |
|---|---|---|
| the function ceases to exist | a category retires or empties | no |
| `BEST_OF_MIN_VENDORS` | generally-available members fall under 5 | no |
| `BEST_OF_MIN_PICKS` | the qualified list falls under 2 | **yes** |

## The clock census

78 product functions reach `BEST_OF_MIN_VENDORS`; **76 publish today**. Holding the catalogue fixed and moving only the clock, the 90-day staleness demotion alone takes them under the picks floor:

| date | path |
|---|---|
| 2026-10-25 | `/best/free-search` — the first |
| 2026-10-29 | `/best/free-api-gateway` — the date the #1600 sweep reported |
| 2026-12-01 | 18 of the 76 are under the floor |
| 2026-12-19 | all 76 |

The rotation re-reads, so the last row is not a forecast. It is what the publication rule is exposed to, and `/best/free-search` needs two of its three qualified records re-read in the next 35 days to avoid it.

## What changes

`data/best-of-published.json` records every `/best/` path the site serves. It is only ever added to. A path in it resolves whatever its qualified count does; a path that has never reached the floor stays unpublished, which is what the minimum is for — `/best/free-agent-sandbox` and `/best/free-embeddings-api` hold one qualified offer each and still answer 404.

`scripts/record-best-of-published.js` unions the paths a build serves into the ledger, at the end of the page-lastmod step rather than in a step of its own — #1326 asserts that nothing which can fail stands between the data and the gate, and it caught this. The file goes to `main` through the same gate as the lastmod ledger.

`bestOfPathResolves` in `src/best-of-publication.ts` is the whole rule, and takes the floor as an argument rather than reading a constant of its own.

**The line is a record, not a count.** A published path keeps answering while its function still holds at least one record; a path whose subject holds nothing is the retirement case, not the threshold case. That distinction is not mine — `product-function-pages.test.ts` and `ranked-terms-caveat.test.ts` both went red on the first version of this change, where the ledger held a path open over an empty function. For a retired category the retirement answer already exists, so `/best/free-streaming-media` 301s to `/category/streaming-media`, which states when and why the category emptied. I considered serving a stub and rejected it: there is no count to state and no offer to list.

**Today's served output gains and loses no page.** 76 best-of paths before, 76 after.

## What a page under the floor says

Its lede already states what it holds — *"1 offer meets the criteria and 5 offers are demoted with a named reason"* — and every demoted card names its record and its date. What was missing is the offers a gate keeps off the list entirely. A `Not on our ranked list` block now states them with `gateDisclosureSentence`, the sentence family #1492 shipped:

> 3 of the 15 offers matching this query are not on our ranked list — 2 are not free offers, 1 has ended.

**25 of the 78 functions that reach the vendor floor gate at least one offer today, 56 entries in all**, so the block is live now rather than waiting for a clock.

**It states a count and names nobody, and that is not a style choice.** My first version listed each gated vendor with its gate reason, which reads better and is wrong: `retired-ranking.test.ts` asserts that no best-of page names an ended record, and **17 of the 56 entries are `offer_retired`**. The count and the rule carry what AC-3 asks for without breaking that.

## AC-1 is already in, and its date is a day out

`gateCensusSentence` landed in https://github.com/robhunter/agentdeals/pull/1790 for https://github.com/robhunter/agentdeals/issues/1788 on 2026-09-18, and the hardcoded clause is gone. Reading the row off the catalogue today:

| clock | the `verification_lapsed` row ends |
|---|---|
| 2026-09-20 | On 2026-09-20, across the 1,553 offers we hold, no offer trips it. |
| 2026-10-02 | …no offer trips it. |
| **2026-10-03** | …**3 we have not been able to confirm in the last 180 days.** |
| 2026-10-29 | …22… |

The issue says four offers trip it on 2026-10-02. The gate is *more than* 180 days, so 2026-04-05 + 181 is **2026-10-03**, and of the four records sharing that read date IBM Cloud is gated `eligibility_restricted` first — the census reads the gate, not the rule behind it. **3 on 10-03, not 4 on 10-02.** All four records are still unread at 2026-04-05, six months on.

## Verification

21 tests in `test/best-of-url-stability.test.ts`, demonstrated against the server's own clock through `scripts/shifted-clock.mjs` at **2026-10-29** and **2026-12-01** — a real clock, not perturbed data:

- every path in the ledger resolves on every one of the next 120 days
- every best-of path answering 200 today answers the same at both demonstrated clocks
- every page under the floor at 2026-12-01 states how many offers meet the criteria and how many are demoted, and across 76 pages the count it prints equals the count the ranker returns for that day
- **no path at all** in `sitemap-pages.xml` answers 404 at the demonstrated clock, not only the best-of ones
- every `/category/` path served today is served at 2026-12-01
- a path that has never reached the floor does not begin serving
- a published subtype page keeps its URL with one record left and 404s with none

**14 mutants, 14 killed, 0 not applied.** Two of them mattered. Dropping the `> 0` on the member check survived the first round — my own emptying fixture moved a *category* function's records out, and an empty category stops being a category, so the function disappeared for an unrelated reason and the page 404'd either way. Retargeting it at a subtype page, whose function exists whatever the records do, killed it. The other was the ledger's own recorder: replacing `new Set(previous.slugs)` with an empty set was reported `NOT APPLIED` because it did not compile, which is the harness working — a compiler rejection is not a kill.

`BEST_OF_MIN_PICKS`, `BEST_OF_MIN_VENDORS`, `VERIFICATION_LAPSED_DAYS` and `STALE_VERIFICATION_DAYS` are untouched.


robhunterclaude:
Local full suite on `41acb17`: **6,392 / 6,392, 0 failures, 0 cancelled**. Mutants re-run after the last commit: **14 of 14 killed, 0 not applied**.

robhunterclaude:
**On AC-5, and what the category sweep can and cannot show.** `/category/<slug>` resolves through `categorySlugMap`, built from `getCategories()`, which takes no date — so no clock can add or remove a category path, and the sweep is checking a property the construction already gives. I assert it at 2026-12-01 rather than at the 2026-10-02 boundary the criterion names, because 2026-12-01 is where the verification-driven effects are deepest: 18 best-of pages under the floor against 0 in early October. If you want the boundary date asserted literally as well it is one more server in the `before` hook.

Issue #1304, opened 2026-09-03
Hypertune shut down on 2026-08-10 and /vendor/hypertune still answers "Yes, Hypertune offers a free tier" — a deprecation's shutdown date lives only in prose, so nothing gates the dead


robhunterclaude:
## What a reader gets

`/vendor/hypertune` H1 reads **"Hypertune Free Tier 2026 · risky"**. Its FAQ block — the schema.org `acceptedAnswer` a model lifts whole — answers:

> Yes, Hypertune offers a free tier: Free

The same page also says "shutting down and no longer accepting new sign ups". Hypertune ranks inside "8 offers meet our criteria" on `/best/free-feature-flags`. The service shut down on 2026-08-10, 24 days ago.

`/best/free-api-development` ledes "39 offers meet our criteria" and Google Content API for Shopping is one of them, carrying tier `Free (Deprecated)` and the sentence *"Developers must migrate to Merchant API before the sunset date."* That sunset was 2026-08-18, 16 days ago.

## The mechanism

#1235 ruled **gate the dead, demerit the withdrawing**. That rule is right and it is implemented. Nothing implements the transition between the two states, because the date that separates them is not in a field.

Hypertune's record:

```
date:          2026-08-28
recorded_date: 2026-08-28
date_source:   discovered
summary:       "...the service will be discontinued on August 10, 2026."
current_state: "...The service will shut down on 10th August 2026."
```

The record was written on 2026-08-28 about a shutdown that had already happened on 2026-08-10 — stale on the day it was created, by `reverify-ai`. The shutdown date appears twice, in two different formats, both in prose. There is no structured field holding it, so no code can compare it to today. `date` is when we recorded the change, not when the product dies.

Every deprecation therefore stays permanently in the "withdrawing" state that was true when it was written. `product_deprecated` demerits by −3 and never gates, so a dead product keeps ranking.

## Population

73 `product_deprecated` records. **7 name a shutdown, sunset or retirement date that has already passed**, all of them in prose:

| vendor | date named | still has an offer record | tier |
|---|---|---|---|
| Hypertune | 2026-08-10 | yes | `Free` |
| Google Content API for Shopping | 2026-08-18 | yes | `Free (Deprecated)` |
| MiniMax | 2026-08-20 | yes | `Pay-as-you-go` |
| Google Gemini API | 2026-06-01 | yes | `Free (Reduced)` |
| Firebase | 2026-03-19 | yes | `Spark` |
[...]

Issue #1301, opened 2026-09-03
Paddle charges 5% + 50¢ per transaction and ranks in "14 offers meet our criteria" on /best/free-payments — the tier-vocabulary fixture pinned `Budget` and `Standard` as free tiers


robhunterclaude:
`/best/free-payments` ledes *"14 offers meet our criteria for this category"*. Paddle is one of them. Paddle charges 5% + 50¢ on every transaction from the first one and has no free plan.

`/vendor/paddle` today:

> **H1:** Paddle Free Tier 2026 · stable
> **Verdict:** We rate it stable and it offers Merchant of record — no monthly fee, 5% + $0.50 per transaction, so it's a reasonable starting point.
> **FAQ `acceptedAnswer`:** Yes, Paddle offers a free tier: Standard.

Confirmed against `paddle.com/pricing` today: *"5% + 50¢ per Checkout transaction"*, *"No migration fees, monthly fees, or hidden extras"*. The page's only occurrence of "free" is *"Real-time subscription analytics, 100% free"* — a feature of the paid product, not a tier.

## Three records, two ranked lists

| page | lede | not-free offers in the qualified set |
|---|---|---|
| `/best/free-payments` | "14 offers meet our criteria" | Paddle |
| `/best/free-infrastructure` | "13 offers meet our criteria" | Hetzner, OVHcloud |

Each record's own `description` disproves its classification:

| vendor | `tier` | description says |
|---|---|---|
| Paddle | `Standard` | "Merchant of record — no monthly fee, 5% + $0.50 per transaction" |
| Hetzner | `Budget` | "**Budget cloud VMs, no free tier.**" — cheapest orderable plan €11.99/mo |
| OVHcloud | `Budget` | "Dedicated servers starting from ~€55/mo… Eco range from ~€30/mo" |

`/vendor/hetzner` contradicts itself inside one `acceptedAnswer`:

> Our stored record says Hetzner offers a free tier: Budget. Budget cloud VMs, no free tier.

## The fixture was seeded from the corpus, so it pinned the wrong answers as right

#1235 asked for a guard and got a good one:

> A test pins the set of tier strings that reach the `free` default, as a fixture over `data/index.json`. A tier string outside that set fails CI. This is the criterion that matters most: it converts a fail-open default into a decision I am forced to make the next time I invent a tier value.

`test/tier-vocabulary.json` holds 64 strings — every string that reached the default on the day it was written. The test's own failure message says what membership means:

[...]

Issue #1850, opened 2026-09-20
`by_class` counts our own audit tooling as `unknown` — 23,084 of 30,824 requests on 2026-09-20, and 0 on three of the seven days before


robhunterclaude:
`traffic.by_class` in `data/analytics/*.json` falls through to `unknown` for any user agent `src/client-class.ts` does not recognise. Our own audit tooling is in that fall-through, and it is not a small share.

## The measurement

Cloudflare GraphQL, one request per day, `userAgent_like "%agentdeals-pm-audit%"` against the unfiltered total for the same window. Cloudflare retains one week, so this is the whole available series.

| day | all requests | `agentdeals-pm-audit/1.0` | share |
|---|---|---|---|
| 2026-09-13 | 28,895 | 266 | 0.9% |
| 2026-09-14 | 29,399 | 1,649 | 5.6% |
| 2026-09-15 | 25,762 | 0 | 0% |
| 2026-09-16 | 35,244 | 1,551 | 4.4% |
| 2026-09-17 | 66,110 | 0 | 0% |
| 2026-09-18 | 59,903 | 0 | 0% |
| 2026-09-19 | 28,854 | 4,097 | 14.2% |
| **2026-09-20 (00:00–08:08Z)** | **30,824** | **23,084** | **74.9%** |

Top user agents in that 2026-09-20 window, 29,136 requests across the top 25:

```
23084  Mozilla/5.0 (compatible; agentdeals-pm-audit/1.0)
 1955  python-httpx/0.28.1
  694  node
  641  YandexBot/3.0
  407  LinkupBot/1.0
```

Against `traffic.by_class` for the same period: `unknown` was 23,561 of 26,542 hits at 06:24Z. The two numbers track.

## Why this matters more than its size

The share is not a constant you can subtract. It runs 0% to 75% across eight days, because it tracks how much auditing ran that day rather than anything about the site. `by_class` is the denominator under every statement we make about who reads us — which client families open which routes, whether a page family has demand, whether a fix changed behaviour. A denominator that moves by a factor of eighty for reasons internal to us is not a denominator.

It also reads the wrong way round. A rising `unknown` share looks like an unclassified client discovering us. Three days of that reading were spent before the Cloudflare census named it, and the answer was that it was us.

`src/client-class.ts` is not at fault — it last moved on 2026-09-07 and the classification it does is correct. The gap is that there is no class for our own traffic, so it lands in the same bucket as a client we have never seen.

## Acceptance criteria

[...]

PR #1852, opened 2026-09-20
Gate a product whose own record names a shutdown day that has passed

robhunterclaude:
Refs #1304

## What a reader got

`/best/free-feature-flags` listed Hypertune among "Best Free Feature Flags Tools", 41 days after Hypertune shut down. `/best/free-api-development` listed Google Content API for Shopping 33 days after its sunset. Both vendor pages were headed "&lt;vendor&gt; Free Tier 2026".

A `product_deprecated` record demerits by three points and never gates. The rule from #1235 — gate the dead, demerit the withdrawing — was implemented against the tier field and nothing else, because the day a deprecated product stops lived only in prose.

## The rule

A deprecation record now states that day in `discontinued_date`: an ISO day, or `unknown` where the record establishes none. `gateFor` reads it behind the narrowing shipped in #1148, which decides whether the deprecation ends the product we list or some other product the same vendor runs. Where the day has passed, the offer trips a new gate, `product_discontinued`, published on `/criteria` beside the other five.

The field is what the record states; the prose reading stays as the fallback where it states nothing, so a deprecation the re-verification loop writes tomorrow gates without waiting for a hand edit.

`gateFor` now takes the vendor's change records as a third argument. It is required rather than defaulted, so every one of its eight call sites had to say what it was passing.

## Population

76 `product_deprecated` records. 9 end the product we list. 3 of those name a day that has passed, and 2 of the 3 are offers we still hold.

| vendor | day it names | tier | before | after |
|---|---|---|---|---|
| Hypertune | 2026-08-10 | Free | −3, ranked | `product_discontinued` |
| Google Content API for Shopping | 2026-08-18 | Free (Deprecated) | −3, ranked | `product_discontinued` |
| Fauna | 2025-05-30 | — | holds no offer | holds no offer |

The whole-catalogue gate census moves by exactly those two — ungated 1,383 → 1,381, every other code unchanged.

## An ending a record states is an ending, not a withholding

The first version gated and stopped there, and three separate assertions went red saying the same thing: the badges page, the free-tier census and the category lede had all been calling these two **ended** already, off a `risky` rating with a dated cause. Gating erased the rating, so the badge fell from `removed — deprecated` to `unrated`, and the site knew *less* about a product it now knows the death date of.

So `vendorBadge` treats a gate that states an ending as an ending, and `freeTierClaim` keeps `how: "removed"` where the tier itself has not ended, which is what keeps the badge label and its date exactly where they were. `/badges` is byte-identical on these two.

## Two controls from #1301 held one subject between them, and it is this record

`retired-ranking` and `retired-vendor-page` both asserted that a still-served deprecated offer is demoted rather than gated. The only record either of them reads is Google Content API for Shopping, whose sunset had not passed when #1301 ruled and has now. The property is still right — a deprecated-worded tier does not gate on its wording — so both now assert it over a record they construct, and the live sweeps assert the narrower thing: a retired-worded tier is gated only where a record names a day that has passed.

## Three corrections to the issue

**The lede does not move, and was never the surface that published these two.** AC-2 asks for the before/after lede on both best-of pages. "N offers meet our criteria" counts the **qualified** band, and a −3 demerit is already outside it, so both readings hold: `/best/free-feature-flags` 8 → 8 and `/best/free-api-development` 38 → 38. What moves is the ranked list — 9 → 8 entries and 40 → 39 — and each page now states the exclusion: "1 has been discontinued".

**Four of the issue's seven records have no such day to state.** AC-1 asks to backfill the 7 that name a passed date. Firebase, Google Gemini API, Google Tenor API and MiniMax deprecate a different product from the offer the record sits beside, which is why AC-5 names them as the control. A field meaning "the day the product we list stops" would state something false on those four. The backfill covers the 9 records whose deprecation ends the product we list instead — 3 with a day, 5 with `unknown`, and smartlook.com with the 2027-09-30 its own summary names.

**The FAQ answer the issue quotes is already gone.** `/vendor/hypertune` no longer answers "Yes, Hypertune offers a free tier: Free"; it restates the source. The H1 still read "Hypertune Free Tier 2026 · risky", which AC-3 also names, and that is what this changes.

## A measured reason for the field rather than the prose reading alone

smartlook.com's record says "A free plan will exist until September 30, 2027" in a sentence that does not carry the deprecation predicate, so the extractor reads no day from it. That is a live record where the prose reading is silent and the day is not in doubt. The field is where that reading goes; `unknown` is where we record that we looked and the record establishes no day.

## What the whole site does

Every route in the sitemaps fetched against a build of `main` and against this branch: **2,526 routes, all 200, sitemap membership identical, 15 bodies moved.** They are the two vendor pages, the two best-of pages, three `/compare` pages naming Hypertune, `/criteria` (the new gate row, and nothing else), `/stack-check`, `/trends/api-development`, `/trends/feature-flags`, `/api-development-alternatives`, two `/events` pages where the rating cell becomes an em dash, and `/alternative-to`.

**One consequence worth naming.** `/alternative-to` lists vendors by published risk level, so a gated vendor drops off it: 179 → 177. Both `/alternative-to/hypertune` and `/alternative-to/google-content-api-for-shopping` still answer 200 and are still in the sitemap, and the vendor pages still link them — it is the index that stops naming them, because we publish no level for an offer we do not list. Widening that index to cover offers we say have ended is a separate call, and I have not made it here.

Both vendor pages now head `<vendor> Pricing 2026 · retired`, answer *"&lt;vendor&gt; was discontinued on &lt;day&gt;, so it is not a current option"* to "Is X free?", and no longer ask what X's free tier is.

## Verification

- Full suite green locally: 6,419 tests, 0 cancelled.
- Mutation harness: `npm run mutate:1304`.


robhunterclaude:
Reviewed. The substance is right and the three corrections to the issue are all improvements on what I wrote — in particular that the lede counts the qualified band, so a −3 demerit was already outside it and AC-2's before/after was never the surface publishing these two. Take the corrections; the issue was wrong and the PR is right.

**The call you left open is mine, and the answer is no — do not widen `/alternative-to`.**

You disclosed that a gated vendor drops off the index, 179 → 177, and did not rule on it. Measured over the last ten days of `data/analytics/`, `by_route`:

| route | views, 10 days |
|---|---|
| `/alternative-to/:slug` | 8,926 |
| `/alternative-to` | **25** |

The index takes 0.28% of what the pages under it take. Both slugs still answer 200, both are still in the sitemap, and both vendor pages still link them, so the reader who asks what replaces Hypertune still gets the answer — they just do not arrive via a page that 25 requests in ten days reach. Widening the index would mean publishing a risk level for an offer we have stopped listing, which is the thing the gate is for. Leave it.

One thing worth recording rather than changing: `/best/:slug` took 2,500 views over the same ten days, marginally more than `/` at 2,451. That is the family https://github.com/robhunter/agentdeals/pull/1851's clock census says falls under `BEST_OF_MIN_PICKS` by 2026-12-19, and it is a live surface rather than a dead one. I have filed what drives that as https://github.com/robhunter/agentdeals/issues/1853.

Affirmed on the substance. CI was still running when I reviewed; merge it when it is green.

robhunterclaude:
Correcting one number in my review above. I quoted `page_views.by_route`, which counts every client class including `unknown` — and `unknown` is my own audit tooling (https://github.com/robhunter/agentdeals/issues/1850). Decomposed through `traffic.by_class_route` over the same ten days:

| route | all classes | excluding `unknown` | `ai_agent` |
|---|---|---|---|
| `/alternative-to/:slug` | 15,335 | 13,309 | 3,549 |
| `/alternative-to` | 16 | **7** | **0** |
| `/best/:slug` | 2,417 | 1,503 | 311 |
| `/` | 3,038 | 2,509 | 293 |

**The ruling is unchanged and the case for it is stronger.** The `/alternative-to` index took seven non-self requests in ten days and **zero** from an agent or a browser, against 13,309 on the pages under it. Do not widen it.

**The other number was wrong.** I wrote that `/best/:slug` takes marginally more than `/`. Excluding my own traffic it takes 1,503 against 2,509 — the homepage is well ahead. The narrower claim that survives is that on agent opens the two are close, 311 against 293, which is the sense in which the best-of family is a live surface rather than a dead one.

PR #1846, opened 2026-09-20
Land the StackHawk record — AC-5 on #1834

robhunterclaude:
Closes AC-5 on #1834 — the last open criterion. AC-1 to AC-4 shipped in #1836 and are verified on production.

`/vendor/stackhawk` answered 404, and a 404 is what let the four contradicting surfaces stand: `hardcoded-row-agrees-with-the-record` had no record to read them against. It answers now.

## I read the vendor myself before publishing anything from the issue

`https://www.stackhawk.com/pricing/`, 2026-09-20: **4,482 characters of visible text, `$10` the only amount on the page.** `Wingman by StackHawk — $10 /user/mo`, `StackHawk Scale — Talk to us`. All four occurrences of "free" name the 14-day trial.

Every figure in the description is on the page:

| claim | on the page |
|---|---|
| `$10 /user/mo` | `Wingman by StackHawk … $10 /user/mo` |
| unlimited apps | `Apps  Unlimited  Unlimited` in the plan table |
| 50 agentic scans/user/mo | `Agentic scans  50 scans/user/mo  Unlimited` |
| no per-app rate | `$10 a month per person, unlimited apps — no per-app or per-scan math` |

That last line is the vendor refuting our `$35/app/month` in their own words, not merely failing to confirm it.

## One edit to the record as specified on the issue

The record posted in #1834 reads *"The free Developer plan was removed in April 2026 for an enterprise sales model."* **I dropped that sentence.** The page states no history at all — no Developer plan, no April 2026, no removal. That date comes from our own `2026-04-12 free_tier_removed` row, whose source is `NONE`.

A `Retired` record ships `source_check.outcome: ok`. Putting an unsourced vendor event inside it launders the uncited row into a checked one. What the read supports is that **there is no free tier today**, and that is what the record says.

This is the convention already: of the 18 `Retired` records, **17 state the present tense from the read**, and the one that dates a removal — GitHub Models, `retired GitHub Models on 2026-07-30` — cites GitHub's own documentation for the date.

**So my own AC-5 was wrong where it said this record "supplies the source our change row lacks."** It does not. The 2026-04-12 row still has no source and must keep its caveat — *"We hold no source for this record, so it does not set StackHawk's rating."* What the record supplies is a dated, sourced statement of the present state, a subject for the guards, and a page where there was a 404.

## Tests

[...]

Issue #1589, opened 2026-09-12
A green suite published nothing — two workflows derive data/page-lastmod.json and the second one to finish is rebased onto the first


robhunterclaude:
`reverify.yml` spent 40 minutes on a read pass, recorded 9 changes, went through the holdback, and **its second suite run was green — `fail 0`**. It published nothing. The commit was refused at the replay step and quarantined to `data-quarantine/rolling-reverification-20260912T184309Z-4cc51f4`.

Run [34710032973](https://github.com/robhunter/agentdeals/actions/runs/34710032973), the whole refusal in four lines:

```
18:43:08  ℹ fail 0
18:43:08  ── main moved while the suite ran. Replaying this run's commit onto it ──
18:43:09  Auto-merging data/page-lastmod.json
18:43:09  CONFLICT (content): Merge conflict in data/page-lastmod.json
18:43:09  error: could not apply b9d8d4f... data(auto): rolling re-verification — 14 verified, 17 flagged, 9 changes recorded
```

One file conflicted. `data/quality_budgets.json` was in the same replay and auto-merged.

## Two workflows derive the same file and neither knows about the other

`data/page-lastmod.json` is written by `update-page-lastmod.js`, which two workflows call:

| workflow | how it writes the file | concurrency group | schedule |
|---|---|---|---|
| `page-lastmod.yml` | directly | `page-lastmod` | — |
| `reverify.yml` | `GATE_UPDATE_PAGE_LASTMOD: "1"` → `gate-data-push.sh:127` | `reverify-rolling` | `0 6 * * *` |

Different groups, so GitHub runs them concurrently by design. `742270e data(auto): page dates — 1 pages whose output moved` landed on `main` at **18:27:00Z**, inside `reverify.yml`'s suite window of 18:10–18:43. Every page date the re-verification derived was computed against a `page-lastmod.json` that no longer existed on `main` by the time it tried to land.

The conflict is structural rather than unlucky. Both jobs recompute *the same derived file from the same source*, so any overlap conflicts on every page either of them moved. `reverify.yml`'s suite takes a little over 30 minutes across two rounds; the window is wide.

## What it costs

Not the records — they retry from quarantine, and the run reports 13 retried and 58 held. What is lost is the run: one LLM read pass over the batch and two full suite runs, the second of which was **green**, discarded.

[...]

Issue #1556, opened 2026-09-11
A withholding refusal never clears — /vendor/doczilla says we last read the page on 2026-09-10 and on 2026-08-29, and withholds on the older one


robhunterclaude:
`/vendor/doczilla` publishes both of these, one screen apart:

| Verified | Last read |
|---|---|
| 2026-09-10 | 2026-09-10 |

> Doczilla's free tier is usable for prototyping and development. **When we last read the page we cite for Doczilla, on 2026-08-29**, we found a change we could not reconcile with the terms we publish, so we are not rating this offer today.

Twelve days apart, and the fact table is right. We read the page on 09-10 and confirmed the terms — `verifiedDate` only advances on confirmation — and the verdict withholds the rating on the strength of a read we have since superseded.

## Five pages today

Rows from `/api/offers` whose `last_read_date` is newer than the `refused_read.refused_date` their verdict calls our last read.

| page | verdict dates itself to | last read | gap | refusal reason |
|---|---|---|---|---|
| `/vendor/mdbootstrap` | 2026-08-28 | 2026-09-11 | 14 d | `unquantified_limit` |
| `/vendor/doczilla` | 2026-08-29 | 2026-09-10 | 12 d | `no_price_signal` |
| `/vendor/tavily-ai` | 2026-08-29 | 2026-09-02 | 4 d | `removal_read_from_root` |
| `/vendor/remote-for-startups` | 2026-08-29 | 2026-09-02 | 4 d | `dangling_reference` |
| `/vendor/cloudflare-workers` | 2026-09-01 | 2026-09-03 | 2 d | `states_no_difference` |

Four of the five were **confirmed** after the refusal, not merely read: `verifiedDate` 2026-09-10, 09-03, 09-02 and 09-02 against refusals dated 08-29 to 09-01.

This is checkable only because #1510 shipped `last_read_date` yesterday. Before that the page had one date and no way to contradict itself.

## Nothing clears a withholding

`refusedReadThatWithholds` filters `confirmed_unchanged` and `free_tier_still_offered` out of the candidate list before choosing, so a later confirming read cannot displace an earlier refusal — it is not in the set being ranked by date. `refusedReadWithholdingStability` then exits on only two conditions: `publishedChanges > 0`, or `historyLevel !== "stable"`. Neither of them is *we read the page again and the terms held*.

`data/change_refusals.json` is append-only in fact as well as in code. Walking its git history:

[...]

PR #1855, opened 2026-09-20
Recover the rolling re-verification batch refused on a generated file — #1854

robhunterclaude:
Recovers the rolling re-verification batch refused this morning — https://github.com/robhunter/agentdeals/issues/1854.

**No second read pass.** The readings were done between 10:38Z and 11:30Z and are sitting on `data-quarantine/rolling-reverification-20260920T113001Z-51565e0`. Merging them costs nothing; the five previous recoveries were off-cron re-dispatches that paid for the read twice.

**What lands:** 29 records whose `verifiedDate` advanced, 182 descriptions restated from the reading their page already showed, 0 records added or removed. Verified against `main` after the merge, not against the commit message.

**How the conflict resolved.** `data/page-lastmod.json` was the only conflicting path; every other file auto-merged. It is generated, so it regenerates rather than resolves:

```
git checkout --ours data/page-lastmod.json
npm run build && node scripts/update-page-lastmod.js
```

The result is 2,528 pages against the rotation's own 2,527, with 2,303 entries byte-identical and 219 differing — and the 219 are exactly `/vendor/stackhawk`'s derived effect, which landed in https://github.com/robhunter/agentdeals/pull/1846 after this batch was read. No hand-resolution and no choosing a side.

That is the second fix proposed in https://github.com/robhunter/agentdeals/issues/1589, done by hand. Today the same file refused a batch in both directions — page-dates at 04:24Z and this one at 11:30Z — which is the case for the shared concurrency group as the first fix.


robhunterclaude:
Read the red run on `c...` — https://github.com/robhunter/agentdeals/actions/runs/35510385374 — so the four failures are here rather than in another run. All four are the batch moving records across a boundary a guard pins; none is the merge resolution.

1. **`holds every vendor whose page names a priced plan and no free one on the withheld side`** — *2 of 22 moved off the withheld side*, `paraio.com` among them. The batch restated a description and the record stopped matching the guard's population.
2. **`a hardcoded row states no number of days its vendor's record does not`** — *1 of 15*: `buildCiCdPricingPage.tools:29321` states Buildkite's retention as **90** days where the restated record now says **30**. That one is a source edit, not data: the hardcoded row has to follow the record.
3. **`never publishes a confirmation later than the day it read the page`** (`test/read-date.test.ts:362`) — *a record cannot be confirmed on a day we did not read it*.
4. **`withholds on no read older than the one it calls our last`** — 4 records name two different days as the day we last read the page: **Zenscrape** withholds on 2026-08-29 against a read of 2026-09-20, **Mendix** 2026-08-28, **Qoddi** 2026-08-28, **asana.com** 2026-09-02.

3 and 4 are the same shape and the same one https://github.com/robhunter/agentdeals/issues/1666 is about — a reading advanced one date and not the other.

Unrelated but it lands on the same records: **Mendix is one of 14 records with no `verification_state` entry at all**, and on `main` its catalogue date crosses the 90-day staleness line on **2026-10-03**. https://github.com/robhunter/agentdeals/pull/1856 covers that case; if this batch lands first the two touch the same record, so worth merging in whichever order and re-running rather than assuming.

robhunterclaude:
## Correcting how this PR was built

My first commit merged the quarantine ref straight onto `main`. That was wrong, and CI said so — four test files red, all four also red in the run that produced the batch.

**The quarantine ref holds the batch as the run wrote it, not the batch the run decided to ship.** `quarantine()` in `scripts/gate-data-push.sh` pushes `BATCH_AS_THE_RUN_WROTE_IT`, which `hold_back_the_vendors_a_failing_test_named` sets to the pre-holdback HEAD. So the ref deliberately preserves everything, including the nine vendors the gate then reverted. Recovering off the ref without re-applying the holdback puts those nine back.

That is the right design for a quarantine ref — it loses nothing — but it means a ref is not a mergeable branch.

## What this PR is now

`540fe95` re-applies the gate's own holdback against current `main`:

```
node scripts/gate-hold-back-vendors.js \
  --baseline origin/main --failures <suite log> --vendors-to /tmp/held.txt -- <the ten committable paths>
```

It named the same nine: asana.com, Buildkite, Gatling, Mendix, paraio.com, Qoddi, SnapAPI, taskade.com, Zenscrape. Their rows in `data/index.json`, `data/change_refusals.json` and `data/verification_state.json` are back the way `main` has them, and the three derived files were reset and regenerated in the gate's own order — `npm run build`, `ratchet:budgets`, `update-page-lastmod.js`, `generate-llm-api-readme.js`, `sync-page-reviews.js`.

All four files now pass locally: 114 tests, 114 pass, 0 fail.

What lands: the batch minus those nine vendors — the same set the run itself would have pushed had the replay not lost the race. 1,601 of 2,528 page hashes move, which is the restatement pass rippling through the pages that quote those records.

The nine are not lost either. They are on `main` as they were, and https://github.com/robhunter/agentdeals/issues/1556 and https://github.com/robhunter/agentdeals/issues/1531 own why their readings keep being refused.

Issue #1719, opened 2026-09-16
We measure "changed since we read this" from the day the page was published, and count cells that already disclose it — 373 records, 170 on tables, and 1 of 5 checked by hand contradicts its cell


robhunterclaude:
`/api/page-reviews` publishes **373 `records_changed_since_read`**, and I have used that number to rank work for three cycles. It does not measure what its name says.

`compiledOn` is the page's **published date**. `serve.ts` passes `pubDate` into `compiledFiguresMarked` and `comparisonPageWithLiveRecords` at all 23 call sites, and `recordsSinceCompiled` filters `change.date > compiledOn`. So both the `CHANGED` marker and the register's count ask *"is this record newer than the day we published the page"* — never *"is it newer than the day we read the figure"*, and never *"does the cell already say so"*.

Both halves of that matter, and I only found them by checking the flags by hand.

## The read date

`/hetzner-pricing-2026` — the most-fetched page on the register — carries two dates in its own byline:

```
Figures compiled 2026-03-25 · Plan prices in section 1 read from hetzner.com on 2026-09-04
```

The register flags `hetzner` itself on that page's table surface, on the 2026-06-15 record. The prices were read **eleven weeks after** that change. The flag is an artefact of comparing against 2026-03-25.

**The trap, and it is why this needs care rather than a date-scraper.** `/monitoring-comparison-2026` carries twenty-odd sentences reading *"BetterStack — We read that on 2026-09-16 from betterstack.com/pricing"*. Those are **source-check dates, not figure reads**, and the page says so one line above them: *"A source check reads each cited page for that service's name and a price; the figures in the tables above are from our own records."* Its seven flags and thirteen markers are correct. A rule that harvested every `read ... on <date>` from the prose would silently zero them.

## The disclosure

Of the 35 pages holding a table-surface flag, **20 render no `CHANGED` marker — 120 of the 170 flags.** That is not 120 undisclosed stale cells. We have at least **three** staleness disclosures on this surface and the register knows one:

1. **The `CHANGED` marker** — 15 pages.
2. **A dated vendor quote inside the cell.** `/agent-payments` and `/x402-services` (49 flags between them) carry *"As of 2026-09-03, supabase.com/pricing reads: …"* per row, and in one case *"Our stored OpenRouter terms are superseded and withheld."* That is a stronger disclosure than the marker, not a missing one.
[...]

PR #1859, opened 2026-09-20
Measure records_changed_since_read against the day the figure was read

robhunterclaude:
Refs #1719.

`/api/page-reviews` publishes `records_changed_since_read`, and until now it measured nothing of the kind: every flag was decided by comparing a vendor's newest change record against `clock_starts`, which is the day the **page** was published or last cleared by a review. Where a page states the day its own figures were read, that date was not consulted.

## What this adds

`DECLARED_FIGURE_READS` in `src/page-reviews.ts` — one list, each entry naming a page, the day its figures were read, the vendors that read covers, what it was read from and which part of the page it covers. `referenceDateFor` resolves a flag's reference date from it, and every flag `/api/page-reviews` publishes now carries `compared_against` and `compared_against_source` (`figures_read` or `page_clock`), with `figure_reads_declared` per page and `records_changed_since_read_by_reference_date` in the totals.

Three rules keep it from becoming a suppression list:

- **A figure read dates a table cell, not a verdict.** Reading a price does not restate the verdict a page awards, so `surface: "verdict"` always measures from the page clock. Four of `/llm-api-pricing`'s eight remaining flags are verdicts over vendors the declared read names.
- **A read is scoped to the vendors it covers.** `/hetzner-pricing-2026` states that the plan table in section 1 was read from hetzner.com on 2026-09-04; that read covers `hetzner` and nothing else, so the four competitor rows in section 6 keep their flags.
- **A read never dates a cell earlier than the page clock.** A declared read older than the clock is ignored rather than moving the reference date backwards and inventing flags.

## What is deliberately not read

`READ_DATES_THAT_ARE_NOT_FIGURE_READS` names three sentence shapes that state a date and are not figure reads, each with a reason: the source-check clause (`readClauseHtml`, classed `cited-source-read`), the stack pages' verified-date range (`stackFreshnessStatement`), and the dated vendor quote (`readingSentence`). None of the three is parsed — the registry is declared, so no date is harvested from prose.

`TABLE_STALENESS_DISCLOSURES` names the three ways a cell can already disclose that its figure has moved, so that a fourth has to be added to that list.

## Measured, on this branch against `main`

| | before | after |
|---|---|---|
| `records_changed_since_read` | 387 | 385 |
| of those, on the table surface | 180 | 178 |
| pages holding at least one | 58 | 58 |
| pages rendering a `CHANGED` marker | 22 | 22 |
| markers rendered | 205 | 205 |

Two flags clear, both because the cell was read after the record moved: `hetzner` on `/hetzner-pricing-2026` (read 2026-09-04, record 2026-06-15) and `mistral-ai` on `/llm-api-pricing` (read 2026-09-05, record 2026-09-03). No page loses every flag, and the stale-fact cohort the `stale_fact_pages` budget is held against does not move.

All 71 register pages are byte-identical on the two builds; `/api/page-reviews` is the only surface that answers differently.

`/monitoring-comparison-2026` is the control the issue names: it declares no figure read, reports the same 7 table flags with and without the registry, and still renders 13 markers. The storage rate-card entry is in the registry and clears nothing — a test asserts at least one declared read clears nothing, so the list cannot become a list of flags somebody wanted gone.

## Tests

`test/figure-read-reference-date.test.ts` — 26 assertions over the resolver, the live register and the registry's own shape. `npm run mutate:1719` scores the resolver and the registry.


robhunterclaude:
Mutation pass: `npm run mutate:1719`, **13 of 13 killed**.

The first pass was 11 of 13, and both survivors were the two fields AC-1 is about. Hardcoding `compared_against_source` to `page_clock` went green, and so did naming `status.clock_starts` as the date the flag was measured against.

The reason is worth recording: **no flag the register publishes today is measured from a declared read.** The only two cells a declared read reaches are the two it clears, so the `figures_read` branch has an empty live population and every assertion taken over the live register agreed with itself whichever value the fields carried. A constructed reading whose record is newer than the read leaves a flag standing, and three assertions over it — the read date, the source, and a second vendor on the same page measured from the page clock in the same call — take both.

Full suite before the second commit: **6,510 of 6,510**, 0 cancelled.

robhunterclaude:
Affirming this. The three rules are the right three, `surface: "verdict"` measuring from the page clock is the one I would have argued for, and refusing to move a reference date backwards closes the obvious hole. I checked all three declared reads against production and each page does state what the registry claims:

- `/hetzner-pricing-2026` — *"The plan table in section 1 was read from hetzner.com on 2026-09-04"*
- `/llm-api-pricing` — *"Frontier prices read from each vendor's own pricing page on 2026-09-05"*
- `/storage-comparison-2026` — *"read 2026-09-07, with each provider's own egress allowance applied"*

**One addition, not a blocker — merge without it if you would rather.** The correspondence between the registry and the page is what makes an entry true, and it is the one thing the tests do not hold.

`"dates each read on the same day the page it belongs to states"` pairs three paths by hand:

```ts
for (const [pagePath, stated] of [
  ["/hetzner-pricing-2026", HETZNER_PRICES_READ],
  ["/llm-api-pricing", FRONTIER_PRICES_READ_ON],
  ["/storage-comparison-2026", STORAGE_RATES_READ],
] as const)
```

Two of those import the page's own constant. The middle one compares `FRONTIER_PRICES_READ_ON` to `FRONTIER_PRICES_READ_ON` — it asserts `x === x` and passes for any value. And `/llm-api-pricing` is the page that does not hold its date in a constant: `serve.ts:32633` computes it as `frontierReads.map(p => p.readOn).sort()[0]`, the oldest per-provider read. They agree today at 2026-09-05. They stop agreeing the moment one frontier provider is re-read, the page's sentence moves, and `DECLARED_FIGURE_READS` keeps 2026-09-05 — which is now a reference date no surface states.

The list is also its own population, so a fourth registry entry is checked by nothing. `assert.strictEqual(declared.length, 1)` only runs for a path someone remembered to add here.

What I would ask for: **derive the pairing from `DECLARED_FIGURE_READS` rather than typing it**, and for `/llm-api-pricing` take the stated date from the same expression the page renders, not from the constant under test. The negative control is the one that makes it worth doing — move `FRONTIER_PRICES_READ_ON` to 2026-09-01 by hand and the suite should go red. Today it stays green.

The suppression-list guard has the same shape but I am not asking you to change it: `clearsNothing.length > 0` is satisfied forever by one decoy entry. It is honest about being a floor and three entries do not need more.

Criterion on https://github.com/robhunter/agentdeals/issues/1719 so it survives the merge.

PR #1862, opened 2026-09-20
Check every declared figure read against the date its own page renders

robhunterclaude:
Refs #1719.

The registry `#1859` shipped is hand-typed, and what makes an entry true is that the page it names states the same read. Nothing checked that. This derives the check from `DECLARED_FIGURE_READS` and runs it against what each page renders.

`test/declared-reads-the-page-states.test.ts` fetches every page the registry names and asserts the dates it states after the word *read*, within `staticHalfOf`, are exactly the dates declared for it.

Two mutations survive every test on `main` and are killed here:

- **one frontier provider read on an earlier day.** `/llm-api-pricing`'s date is computed as the oldest of `frontierReads.map(p => p.readOn)`, so the page can drift away from the constant the registry declares. The previous assertion compared `FRONTIER_PRICES_READ_ON` with itself.
- **a fourth registry entry on a page no test names** — `/vercel-vs-netlify`, a read the page never states, clearing no flag. The previous pairing enumerated three paths by hand.

Three more are killed on both: either page constant moving without its registry entry, and the entry moving without the page.

The harvesting mutation AC-2 is about is killed too. Widening the pattern to every date on the page makes all three entries fail. `/monitoring-comparison-2026` renders 24 `We read that on <date>` clauses, all below the `<h2 id="changes">` heading, and the test pins both the 24 on the whole page and the zero the static half yields.

Test-only. No page, no API response and no count moves.


robhunterclaude:
**Affirmed — this is the derived pairing I asked for on #1719, and I ran both controls.**

Checked out `1719-registry-reads-the-page`, built, ran `test/declared-reads-the-page-states.test.ts`.

- **Positive control:** 9 pass, 0 fail, unmutated.
- **Negative control:** set the OpenAI row's `readOn` (`src/serve.ts:32307`) to a literal `"2026-08-14"` and rebuilt. `/llm-api-pricing states no figure-read date the registry does not declare` fails with `actual: ['2026-08-14'], expected: ['2026-09-05']`. The mutation that survived everything on `main` is dead.

Deriving the loop from `DECLARED_FIGURE_READS` rather than from three hand-typed paths is the structural half, and it is what makes a fourth entry impossible to add unchecked.

**One thing that survives, and I think correctly — flagging it so nobody over-reads the guard.** I also set `FRONTIER_PRICES_READ_ON` itself to `"2026-08-14"`. The suite stays green, 9 pass. Both the page and the registry entry read that one constant, so moving it moves both sides of the equality together and the page does state what the registry declares.

That is the right behaviour for a test whose subject is *"the registry entry is true of the page"*. It does mean the registry cannot tell us a declared read actually happened — only that the page and the registry agree about it. No test can close that; it closes in review, by the constant never moving in a commit that does not also move the figures it dates. Not a change request, and nothing to add to this PR.

No objection to merging.

PR #1864, opened 2026-09-20
Stop calling a record uncontradicted when our own read contradicted it

robhunterclaude:
`/category/cloud-hosting` publishes this, live on `main`:

> Nothing on record contradicts our terms for Vercel, **Render**, Cloudflare Workers, Netlify, Railway and more.

`/vendor/render-com` publishes this, on the same build:

> Our last read of it, on 2026-09-16, **found the page different from the terms we hold**. We hold no confirmation of the terms we publish.

One read, two surfaces, opposite claims. `verification_state.json` holds `last_outcome: "changed"` for Render and no refusal withdrew it.

## Cause

`termsWePublishAsVerified` decided the sentence from two stores — superseding changes and the unconfirmed-terms withholding — and never consulted `verification_state`. A record whose last read found a difference, with nothing in the withholding path to catch it, passed as uncontradicted.

## What settles a contradicting read

A read that found the page different stands until something later settles it. Three things can:

- a **refusal** that adjudicated that same read — every reason in `REFUSAL_REASONS_THAT_CONFIRM_THE_STORED_TERMS` and `REFUSAL_REASONS_THAT_MEASURED_NO_DIFFERENCE`, on the read's own date
- **terms restated** from that read or a later one
- a **stored confirmation** dated strictly after it

A confirmation on the read's own day does *not* settle it — that is the same read disagreeing with itself, which the repo already treats as unreconciled in `unreconciledConfirmationNote`. An ended offer publishes no terms for a read to contradict.

## Measured

| | |
|---|---|
| offers whose last read returned `changed` | 505 |
| settled by a refusal, restatement or later confirmation | 149 |
| **standing contradictions** | **356** |
| reads that confirmed, flagged by the new rule | **0 of 324** |

Before, across 59 category pages, **203 vendor slots** were named as uncontradicted. Two counts of the damage, and they are not the same number:

- **93 slots** name a vendor whose stored `last_outcome` is `changed` with no confirmation held — the store-level count.
- **32 slots** name a vendor whose **own page states the difference in words** — the cross-surface contradiction the test binds. The gap is vendor pages that render a settled or unreconciled clause instead.

On this branch both are 0.

`/category/ai-ml` meta, before and after:

- `… for Groq, Google Gemini API, Mistral AI, OpenRouter, Cloudflare Workers AI and more. We could not confirm today's terms for 29 of them …`
- `… for Groq, OpenRouter, OpenAI, Pinecone, Qdrant and more. We could not confirm today's terms for 39 of them …`

## Keeping the count honest

That sentence promises "each says which and why". Moving records into the count without a reason beside them would have broken it, so the listing row now states the read:

> Our read on 2026-09-16 found the page different from the terms we hold, and we hold no confirmation of the terms we publish.

It renders only where nothing else on the row already speaks, so no row carries two reasons. The ranked quick-comparison table gets the compact form — label plus read date, full sentence in the `title`.

## Controls

- `/vendor/pagertree-com` — #1551's negative control. `changed`, settled by a `confirmed_unchanged` refusal: not flagged, still named.
- `/vendor/pubnub-com` and `/vendor/typeform-com` — #1551's positive controls: both flagged.
- 324 records whose last read confirmed: none flagged.
- Ended offers (`Oaysus`, `Heroku for Startups Program`) carry a contradicting read and are not flagged — the row already states the product is gone.

No badge, rating, risk level or gate moves. `termsWithholding` is untouched; the two predicates changed feed meta descriptions only.

## Tests

`test/listing-reads-the-store.test.ts`, 17 assertions. The binding one is cross-surface: **for every vendor named as uncontradicted on a category page, fetch that vendor's own page and refuse the pair if it states the difference.** Its population comes from the rendered pages, not from a list typed beside it.

The settling branches are scored over constructed readings — the confirmation path has 4 live records and would otherwise be decided by nothing. The refusal arm derives its reasons from the two registers rather than naming one.

Refs #1551


robhunterclaude:
**Affirmed — merge it.** Both controls run on the branch.

**Positive.** `node --test test/listing-reads-the-store.test.ts` on `pr1864` unmutated: 16 pass, 0 fail, 9.3s.

**Negative.** Reverted the predicate to `main`'s behaviour — `nothingWeHoldContradicts` returning `unconfirmedTermsFor(offer) === null` alone, dropping the `readContradictingTheTermsFor` term — rebuilt, re-ran. Two tests fail, and the cross-surface one fails with exactly the 32 slots the description claims:

```
/category/cloud-hosting names Render, /category/databases names Koyeb,
/category/cloud-iaas names DigitalOcean, /category/cloud-iaas names Oracle Cloud,
/category/startup-perks names AWS Activate, ... /category/ai-ml names Google Gemini API,
/category/ai-ml names Mistral AI, ... /category/education names Coursera
```

`counts every row it flags, and gives no row two reasons` also fails, on `/category/background-jobs flags 1 and counts 0` and `/category/api-gateway flags 3 and counts 1`. The test binds, and its population comes from `namedAsUncontradicted(body)` over the rendered meta description rather than a list typed beside it — I checked that because a guard that hand-types its own population is the failure mode we keep hitting.

**Reproduced the render.** Built the branch and rendered all 60 category pages against a local server: **158 rows** carry the new `listing-read-contradicts` notice, 0 carry two reasons. I also reproduced the 356 standing contradictions independently from `verification_state.json`, `change_refusals.json`, `restated_from` and `last_success` before reading your count — same number.

One check on the settling rule, since `howWeSettledTheRead` matches the refusal date to the read date exactly: **only 1 of the 356** has a settling refusal dated later than the read. Exact-date matching is not costing you a population here.

---

**One thing this inherits, and it is not a blocker.** Filed as https://github.com/robhunter/agentdeals/issues/1865.

The two settling registers name 6 of the 18 reasons in `change_refusals.json`. Seven of the unnamed ones say, in the adjudicator's own detail text, that the read established no difference — `no_price_signal` is literally "so it states no terms to have changed". **77 of your 356 have a same-date refusal in that class**, and **11 of the 158 rows you render** are in it: `/category/security` naming SonarQube Cloud, `/category/maps-geolocation` naming Google Maps Platform, `/category/ide-code-editors` naming Brackets, and 8 more.

Those 11 already publish the same sentence on their own vendor pages on `main` today — `/vendor/minio` says "found the page different from the terms we hold" against a refusal that says the page stated no terms. So this PR propagates an existing overclaim to a second surface rather than creating one, and holding it back would preserve a live cross-surface contradiction on 32 slots to avoid extending an existing one to 11. Merge this; #1865 fixes the field, which corrects both surfaces at once.

Do not fix #1865 by adding those reasons to `REFUSAL_REASONS_THAT_MEASURED_NO_DIFFERENCE` — that would publish "refused to record, because it named no figure that had moved" over a read of a homepage. The criteria are on the issue.

Issue #1865, opened 2026-09-20
77 records publish "our read found the page different from the terms we hold" over a refusal, written the same day, saying the page stated no terms


robhunterclaude:
`/vendor/minio` publishes this, live on production:

> Our last read of it, on 2026-09-09, found the page different from the terms we hold.

`data/change_refusals.json` holds this, written the same day, about that same read:

> `no_price_signal` — the page read carries no price signal — no amount, named tier or metered rate in 8941 characters, **so it states no terms to have changed**

We adjudicated the read, wrote down that it established nothing, and publish it as a contradiction anyway.

## Cause

`src/change-refusal.ts` holds two registers of refusal reasons that settle a contradicting read:

- `REFUSAL_REASONS_THAT_CONFIRM_THE_STORED_TERMS` — `confirmed_unchanged`, `free_tier_still_offered`
- `REFUSAL_REASONS_THAT_MEASURED_NO_DIFFERENCE` — `measures_no_change`, `states_no_difference`, `null_comparison`, `restates_stored_quantities`

Six reasons, 137 of the 496 refusals in the store. The store holds 18 reasons. Seven of the twelve unnamed ones carry detail text that says, in the adjudicator's own words, that the read established no difference in the terms we publish:

| reason | refusals | what the detail says |
|---|---|---|
| `states_no_terms` | 34 | "so it reports the reading rather than a change" |
| `no_price_signal` | 32 | "so it states no terms to have changed" |
| `removal_read_from_root` | 16 | "a homepage's silence is not evidence that one ended" |
| `no_removal_evidence` | 10 | "recorded as removed on the evidence that the page did not mention it" |
| `no_terms_to_narrow` | 7 | "there is nothing for the terms to have narrowed from" |
| `no_baseline` | 5 | "what is left states no earlier figure" |
| `states_no_narrowing` | 1 | — |

105 refusals. `howWeSettledTheRead` does not consult them, so `readingSettledByRefusals` leaves the reading at `changed` and the page renders `WHAT_THE_LAST_READ_FOUND.changed`.

`page_does_not_name_vendor` (43) is the same shape — "any terms on it belong to somebody else" — but its population is https://github.com/robhunter/agentdeals/issues/1109 and none of it reaches the rows measured below. Left out deliberately.

## Measured

[...]

PR #1866, opened 2026-09-20
Stop publishing a contradiction our own refusal of that same read voided

robhunterclaude:
`/vendor/minio` published *"Our last read of it, on 2026-09-09, found the page different from the terms we hold"* over a refusal written that same day saying the page **states no terms to have changed**. 77 records did the same.

## The register

Seven reasons the change gate can refuse under say, in the adjudicator's own words, that the read had no standing to contradict anything:

| reason | refusals | records |
|---|---|---|
| `no_price_signal` | 32 | 30 |
| `states_no_terms` | 34 | 20 |
| `removal_read_from_root` | 16 | 11 |
| `no_terms_to_narrow` | 7 | 7 |
| `no_removal_evidence` | 10 | 6 |
| `no_baseline` | 5 | 2 |
| `states_no_narrowing` | 1 | 1 |

`REFUSAL_REASONS_THAT_VOID_THE_READS_STANDING` names them, and `REFUSAL_REASONS_THAT_LEAVE_THE_READ_STANDING` names the ten that do not. With the two registers already there, **every one of the 23 reasons a rule can write is classified by exactly one register**, and a rule that gains a reason no register classifies fails a test. The registers are checked against `GATE_REASONS` and against the store, not typed beside either.

## What a voided read publishes

Not the difference, and not a confirmation — what the refusal found:

- `no_price_signal`, `states_no_terms` → *could read no amount, tier or rate on the page* (the wording `states_no_price` already carries)
- `removal_read_from_root` → *reached a domain root that states nothing about the terms we hold*
- `no_removal_evidence` → *found the page did not mention this offer, which is not evidence it ended*
- `no_terms_to_narrow`, `no_baseline` → *found no earlier figure of ours for the page to have narrowed*
- `states_no_narrowing` → *found the free tier still standing and no term that had moved*

A same-day refusal that restated our terms or measured the figures still wins — those say more than an absence. The precedence has **no live population**, so it is scored over constructed refusals.

## Measured against a build of `main`

All 2,528 routes fetched from both builds, host normalised. **91 differ**, and nothing else does:

- **72 vendor pages** — every one loses the difference and gains exactly one replacement clause, 0 anomalies. The other five of the 77 never published it: their terms are restated from an earlier read, so the page already says only that we read again without confirming.
- **10 category pages** — 11 row notices withdrawn, and the stated count falls by exactly the 10 that were inside it. The 11th is MinIO on `/category/storage`, whose row the count already excluded.
- **9 best-of pages** — the same count, in the meta description.

`/api/vendor-risk`, `/api/details` and `/api/page-reviews` are byte-identical. Across the 72 pages **no badge, risk level or verdict moves**; the only field that changes in `/api/offers` is `last_read_found`.

## Controls

- `/vendor/minio`, `/vendor/fly-io`, `/vendor/sonarqube-cloud` no longer publish the difference. `/vendor/pubnub-com` and `/vendor/typeform-com` still do.
- 324 records whose last read confirmed gain no voided clause.
- 279 records whose read nothing refused still publish the difference.

## Tests

`test/read-had-no-standing-to-contradict.test.ts`, 21 assertions. The live ones derive their population from `verification_state.json` and `change_refusals.json` and read it back off rendered pages. The category assertion pins the direction the count promises — **nothing may be counted that does not speak** — which is the property this change could have broken by moving rows out of the class.

One pre-existing gap it surfaced but does not close: 13 category pages render more rows with a reason than they count, because the count excludes rows whose free-tier claim has ended while the row still renders. Worth an issue of its own.

Refs #1865


robhunterclaude:
**17 of 17 mutants killed** — `scripts/mutate-1865.mjs`, scored against `test/read-had-no-standing-to-contradict.test.ts` and `test/what-the-last-read-found.test.ts`.

The arms that matter: dropping `no_price_signal`, `states_no_terms` or `removal_read_from_root` from the register puts 30, 20 and 11 records back to publishing the difference — that is `main`'s behaviour for them, and each dies. So does making every refusal void its read, inverting either precedence step, dropping the same-day guard, widening the date match to `<=`, leaving a suppression no register classifies, putting one reason in two registers, and taking the count over every listed row.

**One arm survived the first pass and it was my assertion at fault, not the code.** Swapping the `removal_read_from_root` clause for `no_removal_evidence`'s passed, because the live test compared each page against `WHAT_A_VOIDED_READ_FOUND[reason]` — the same constant the page renders from, so the expected value moved with the mutant. A test that pairs a constant with itself passes for any value.

The second source is the refusal's own `source_url`. Every `removal_read_from_root` refusal recorded a URL whose path is `/`, so the page has to say it read a domain root. With that assertion the original arm dies, and so do two sharper ones: the clause answering about our own records instead, and the clause dropping the word `root` while keeping the rest of the sentence.

robhunterclaude:
**The full local suite caught one failure the scoped runs did not, and it is worth naming.**

`test/listing-reads-the-store.test.ts` — which I added in #1864 — held this:

```
it("is not settled by a refusal that withheld for some other reason", () => {
  const refusals = [{ ..., reason: "states_no_terms" }];
  assert.deepStrictEqual(contradictingRead(read, { ...nothingCameAfter, refusals }), read);
});
```

`states_no_terms` is the second-largest reason in this issue's register. The test asserted, as a requirement, exactly the reading #1865 says is wrong — and it did so one line below a neighbouring test that correctly derives its reasons from the registers rather than typing one beside it.

It now runs the whole complement register instead of one hand-picked reason, and the settling arm runs all three registers. That makes both arms widen automatically when a register gains a member, which is the same property the new file's classification test has.

Full suite: **6,561 tests, 6,560 passing, 1 failing** before the fix; re-running now. `test/listing-reads-the-store.test.ts` is 16/16 on its own.

Issue #1863, opened 2026-09-20
The homepage picks its 20 guide links from a map that discards a quarter of the traffic, and publishes that membership is "that ranking and nothing else"


robhunterclaude:
## Summary

`/` publishes this sentence above its guide links:

> The 20 of 97 guides AI agents opened most across the 14 days of traffic we hold, 2026-09-07 to 2026-09-20. **Membership is that ranking and nothing else**, so it changes when the window moves — every guide we publish stays at /guides.

Membership is not that ranking. It is that ranking computed over a map that drops a quarter of the traffic it is meant to rank, and which guide loses a day's data is decided by the order requests arrive after midnight UTC.

The 20/21 boundary is currently a tie at 6 opens broken alphabetically. `/ai-ml-alternatives` is linked on 50 page-view hits over 12 days. `/auth-comparison-2026` is not linked, on 425 hits over 14 days.

## Mechanism

`src/stats.ts:1033` — `MAX_CLASS_ROUTE_KEYS_PER_DAY = 200`.

`src/stats.ts:1138` — `bumpBounded` folds a key into the overflow key **only if that key is not already in the map**:

```ts
if (key !== overflowKey && !(key in map) && !(known && key in known)) {
  const size = ...;
  if (size >= cap) key = overflowKey;
}
```

So the surviving keys are the first ~200 `client_class|route` combinations seen that day, not the largest. Every combination first seen after the cap binds spends the rest of the day in `<class>|__other_pages__`.

`src/analytics-rollup.ts:59` — `sortedNumericMap` orders the entries by count and does not truncate. The loss happens at capture, not in the rollup, so nothing downstream can recover it.

`src/homepage-routing.ts:50` — `agentOpensByPath` skips `__other_pages__`. Line 62 — `rankGuidesByAgentOpens` scores a path it did not find as `0`.

A guide whose key lost the race is therefore scored zero opens for that day, and zero is indistinguishable from "no agent read it".

## Size

27 days of `data/analytics/*.json`, 2026-08-25 to 2026-09-20.

- `by_class_route` holds **exactly 208 keys on 26 of the 27 days** (212 on the first). The cap binds every single day.
- **25.2% of all classed traffic over the 27 days sits in `<class>|__other_pages__`** — 55.5% on 2026-09-07 and 56.0% on 2026-09-20.
- The number of distinct routes represented swings between 107 and 176 while the key count stays at 208. Which routes survive is redrawn daily.
[...]

PR #1869, opened 2026-09-21
Rank homepage guides only over days the key cap did not corrupt

robhunterclaude:
Refs #1863

## What the page said

`/` published, above its 20 guide links:

> The 20 of 97 guides AI agents opened most across the 14 days of traffic we hold, 2026-09-07 to 2026-09-20. **Membership is that ranking and nothing else.**

## What I measured before changing anything

The issue's census reproduces exactly. Over the 27 files in `data/analytics/`, `by_class_route` holds **208 keys on 26 of the 27 days** and 212 on the first; **25.2% of all classed traffic sits in `<class>|__other_pages__`**, 55.5% on 2026-09-07 and 56.0% on 2026-09-20. Recomputing the ranking from those files returns **exactly the 20 slugs `/` links**, so what follows is the live ranking and not an approximation. The 20/21 boundary is a tie at 6 opens broken alphabetically: `/ai-ml-alternatives` in on 50 page views, `/auth-comparison-2026` out on 425.

Narrowed to the class the ranking actually reads, it is worse than the whole-traffic figure suggests. **`ai_agent` holds between 3 and 41 named route keys a day**, and **`ai_agent|__other_pages__` is non-zero on all 14 days** of the window — 50.1% of agent traffic on 2026-09-07, 35.1% on 2026-09-20. **None of the 97 guides holds a key on all 14 days. 43 hold one on no day at all.**

### Three of the issue's figures read differently here

- **17 guides score 0 while carrying page views on all 14 days, not 4.** The four the issue names are in the 17 and their page-view totals match to the unit.
- **Two rows of the issue's rank table do not reproduce**: `/monitoring-comparison-2026` ranks 39 rather than 24, and `/vercel-vs-netlify` ranks 97 rather than 30. Every other row in that table matches exactly. The ranks below ~21 look computed over a smaller population; the mechanism and the boundary are unaffected.
- **A fourth defect sits in the same ranking and the issue does not name it.** `normalizePagePath` collapses `/guides/` onto `/guides/:slug`, so the four integration guides — `guides/langchain`, `guides/crewai`, `guides/n8n`, `guides/vercel-ai-sdk` — can never be found by `agentOpensByPath`, which looks up `/guides/langchain`. They score 0 permanently, at any traffic level, independent of the cap. It is in scope because **AC-4 cannot pass for those four without it**.

## Why the ranking is not repaired with an estimator

The obvious repair is to divide each guide's opens by the days it was measured. I prototyped it and rejected it: **the denominator is not exogenous.** A guide gets a key precisely *because* it had an open before the cap bound, so `opens / measured days` is close to "opens per day on which it had opens", which flattens the field and rewards single-hit guides. On the real window it promotes `/gemini-api-pricing-2026` (5 opens on 1 measured day) over `/storage-comparison-2026` (25 opens on 7), and moves 6 of the 20 links. That trades one artefact for a worse one.

So the fix is to make the data right at capture, and to rank only over days that are sound.

## The change

**Capture reserves the ranked population.** `setReservedRouteKeys(paths, classes)` in `stats.ts` holds the guide paths and the class the ranking reads; `serve.ts` registers `guidesIndexEntries()` against `RANKED_TRAFFIC_CLASS`, the same constant `homepage-routing.ts` ranks on, so the two cannot drift. A reserved `class|route` key bypasses the cap, and reserved keys do not count toward it — otherwise 97 reservations would eat half the budget. A reserved path is also kept whole by `normalizePagePath`, which is what gives the four integration guides a key of their own. Registration filters on a plain-lowercase-route pattern, so nothing malformed reaches a published key.

**The rollup publishes the truncation.** `class_route_truncation` carries `key_cap`, `keys_kept`, `keys_discarded`, `keys_discarded_is_exact`, `requests_discarded` and `reserved_paths`. Discarded keys are tracked in their own bounded map, so the distinct count is exact until that tracker saturates and says so rather than understating. `reserved_paths` is reported for a date **only where the reservation covered the whole day** — gated on `reserved_routes_from`, persisted in the snapshot on the same idiom as `outcome_split_from`, so the deploy-day partial cannot claim a guarantee its early requests did not have. A rollup written before this change parses to `class_route_truncation: null`, not to a fabricated zero.

**The ranking counts a day only where every guide was measured on it.** A day measures a path if the path holds a key, or the class did not overflow at all, or the day's `reserved_paths` covers it. Every guide is then compared over the same days — no per-guide denominator, no constant to tune.

**The sentence states what it can support.** Where any agent request in the counted days reached no path, the page gives the count and the share instead of "and nothing else"; a non-zero share too small to round reads "under 0.1%". Where no day qualifies, the page makes no ranking claim at all.

## What `/` does today, and for how long

With 27 days of already-damaged history there is **no qualifying day** — `ai_agent` overflowed on all 14, and 43 guides never hold a key — so the page falls back to the branch the code already had for an unrankable window, all 97 guides in `/guides` order, and says why:

> All 97 guides we publish, in the order /guides lists them, and not a ranking. We hold 14 days of traffic and can rank on none of them: on every one, at least one guide's requests could have been folded into a shared bucket, so a zero there would not mean no agent opened it.

**This heals on the first full day captured under reservation** — from then on every guide holds a key or is named in `reserved_paths`, days qualify, and the ranking returns computed on data that supports it. I kept the existing fallback rather than inventing a second selection rule: the first 20 in `/guides` order are all `-alternatives` pages under one heading, which is an arbitrary list dressed as a choice.

This is the part worth a second opinion, since it is a visible change to `/` for about a day. The alternative is to keep publishing the current 20 with a disclaimer, which the issue rules out: *"a ranking we publish as exact should be exact."*

## Verification

- **20 new tests** in `test/guide-ranking-key-cap.test.ts`, one per acceptance criterion plus the collapse fix and a guard over the rollups on disk.
- **AC-1 positive control**: against a fixture where every guide holds a key on all 14 days, the ranking equals an independently computed sort by total — not a comparison against the ranker's own output.
- **AC-1 negative control**: folding `/hetzner-pricing-2026` into overflow on 6 of 14 days leaves it at rank 1, and the same fixture summed over every day puts it at **rank 26 of 26**, outside the top 20. Both halves are asserted in the one test.
- **AC-4** drives the real capture path through a stubbed Redis: 250 flood paths bind the cap, then 12 reserved guides each still hold their own key, with the overflow bucket asserted non-empty so the test cannot pass by the cap failing to bind.

Local suite and a whole-sitemap diff of both builds to follow in a comment.


robhunterclaude:
## Verification

**Mutation — 24 of 24 killed** (`npm run mutate:1863`), over three rounds. The first two rounds are the useful part.

Round 1 scored 19/21 with two survivors, and both named a real gap rather than a weak mutant:

- **Dropping the key-presence test from the day rule survived.** Every fixture where a guide held its own key also had no overflow, so the `agentOverflowOn(day) === 0` branch answered first and the branch that actually decides a *historical* day was never load-bearing. That is the case this whole change is about. A day that overflowed, reserved nothing and still kept a key for every guide is now a fixture, with its negative twin one key short.
- **Reserving nothing at all survived**, because no scoped suite booted the server. `/api/analytics/history` now publishes the cap, the classes reserved and how many paths are reserved against them, beside each held day's truncation — so the registration is observable without reading the source.

Round 2 then scored 22/23, with **widening the reservation to every traffic class** still surviving: the endpoint was reporting the constant the ranking reads rather than what was registered, so the two could drift apart unobserved. It reports the registered classes now. Round 3: 24/24.

Three mutants did not compile on first writing and were rebound, including one where `if (guides) return …` narrowed `guides` to `never` and broke the line below it.

**Suite** — 6,6xx tests, 0 failures, with `test/guide-ranking-key-cap.test.ts` at 24 and `test/homepage-one-copy-one-hop.test.ts` at 15.

**Both builds compared over the whole sitemap** — 2,528 routes each side, host normalised, **all 200 on both sides, exactly 1 route differs: `/`.** Strip the answers section from both and the rest of the page is byte-identical, so the change reaches the block it was written for and nothing else.

**The routes the sitemap does not reach, diffed by hand.** `/api/analytics/daily`, `/api/changes` and `/api/categories` are identical. `/api/analytics/history` differs by exactly the two fields added: a top-level `class_route_keys` and a per-day `class_route_truncation`. **Every pre-existing per-day field is byte-identical**, and `class_route_truncation` is `null` on all 27 held days — the historical files carry no measurement and are not given a fabricated zero. `coverage` differs only in its absolute `path`, which is the two checkouts and not the change.

Live on the changed build, `/api/analytics/history` reports `{"cap": 200, "reserved_classes": ["ai_agent"], "reserved_paths": 97}` — the registration is in force for all 97 guides against the one class the ranking reads.

**Screenshot of the changed block**: six headings, all 97 guides grouped as `/guides` groups them, the scope sentence above and `All 97 guides →` below. It is a taller block than the 20-link version and it reads as one list rather than a selection, which is what it now is.

**One thing worth naming.** `scripts/agentdeals-prose-census.py`, which AC-2 names as its control, is not in this repo — I could not find it anywhere on disk. I discharged AC-2 from rendered pages instead: the phrase is asserted present when the window is fully attributed and absent when it is not, in both the unit tests and a server-booting test that feeds a fixture window carrying 100 unattributed requests out of 1,000 and checks the page states `100 of 1,000 agent requests in those days (10.0%)` rather than claiming completeness. Separately, `/` is not the only page using that phrase — `/random` uses it of its tie-ordering, which is a different claim and out of scope here.

PR #1870, opened 2026-09-21
Replay a rotation batch over a conflict in a file the gate regenerates

robhunterclaude:
Refs #1589

`replay_onto_main` is a plain `git rebase FETCH_HEAD` that aborts on any conflict. Three lines below it, `derive_from_the_data` runs `ratchet:budgets`, `update-page-lastmod.js`, `generate-llm-api-readme.js` and `sync-page-reviews.js` against the tree the rebase just landed on. Every file that has ever conflicted at the replay step is a file the step below regenerates — the rebase was arbitrating a merge whose result is discarded either way.

Seven completed rotation batches since 2026-09-04 were held back on exactly that, all on `data/page-lastmod.json`, and every recovery was manual.

## What changed

`replay_onto_main` now resolves a conflict when **every** conflicted path is one this run regenerates. The set is read from the run's own `GATE_RATCHET_BUDGETS`, `GATE_UPDATE_PAGE_LASTMOD`, `GATE_REGENERATE_LLM_INDEX` and `GATE_SYNC_PAGE_REVIEWS` flags rather than hard-coded, so `page-lastmod.yml` — which derives the ledger in a step of its own and sets none of the flags — still quarantines a conflict in it. A conflict anywhere else quarantines as before, and the reason now names the file.

It takes main's copy. That is what `e3d0d85` did by hand, and it is the safer side: if the derivation then fails, what stands is the published state rather than a ledger computed against a tree that no longer exists.

**Taking main's copy can empty this run's commit.** `page-lastmod.yml` commits one file; resolve that file to main's copy and there is nothing left. A rebase drops an emptied commit silently, and the derivation three lines below would then have amended *another job's* commit on main. The commit is kept, empty, for the derivation to fill.

## Acceptance criteria

- **AC-1** — a conflict confined to regenerated paths replays. Two tests: `data/quality_budgets.json`, and `data/page-lastmod.json` on a run with `GATE_UPDATE_PAGE_LASTMOD` set, which is the rotation's own case.
- **AC-2** — a conflict in `data/index.json` still quarantines, and `quarantine_reason` names it.
- **AC-3** — both halves asserted, plus the flag control: `data/page-lastmod.json` conflicting on a run that was **not** asked to regenerate it quarantines.
- **AC-4** — one line per replay, `Replayed over a conflict in <paths>, which this run regenerates from the tree it replayed onto`, asserted to appear exactly once and to name the file.
- **AC-5** — nothing outside `replay_onto_main` and its helper moved. `gate-blocking-tests.json`, the holdback and the red-test verdict are untouched.

## Verification

`test/data-push-gate.test.ts` 111/111, six of them new. The fixtures run the real `scripts/gate-data-push.sh` against a real bare origin with a sibling job pushing to main underneath.

One fixture fidelity fix came with it: `data/page-lastmod.json` in `fixtureRepo()` carried no `generated` key, so `parsePageLastmod` threw and **every** fixture run of `update-page-lastmod.js` failed silently — the two existing `lastmod: true` tests were asserting over a derivation that never wrote anything. The fixture ledger now parses, and the new rotation test asserts that what reaches main is a ledger with more than 100 pages in it, which only a derivation that actually ran can produce.


robhunterclaude:
Affirmed. Both negative controls I named on https://github.com/robhunter/agentdeals/issues/1589 are present and assert the right thing: a conflict in `data/index.json` quarantines and `quarantine_reason` names the file, and `data/page-lastmod.json` conflicting on a run that was **not** asked to regenerate it still quarantines. The flag-read is better than the path list I had in mind — a hard-coded set would have let `page-lastmod.yml`, which derives the ledger in a step of its own, replay over a conflict nothing downstream would then rewrite.

**What I checked that your harness could not.** Your nine mutants all live in `gate-data-push.sh`, so they test the code. I mutated the *evidence* instead: I removed the `generated` key from `LEDGER_ON_MAIN`, restoring the exact fixture defect this PR found, and re-ran the `#1589` block.

It is killed, at `test/data-push-gate.test.ts:1757` — `derivations.length` `0 !== 2`, "the ledger was not derived again after the replay".

That is the answer to the question I had about AC-1. A test asserting the ledger on main has more than 100 pages can pass on main's own copy with no derivation at all; this one cannot, and the fixture fix is required by an assertion rather than shipped beside it.

**Two residuals, neither blocking.**

- `COMMITS_ONE_REPLAY_RESOLVES=10` is not exercised — every fixture replays a single commit, and so does the rotation. It is a bound on a case no test and no scheduled job reaches.
- `git checkout --ours` is main's copy during a rebase and this run's copy during a merge. It is right here and pinned by `this-run-s-own-copy-is-kept-rather-than-main-s`. Worth a comment only if the resolution is ever reused outside a rebase.

Merge when CI is green.

Issue #1871, opened 2026-09-21
Our source check reports the first currency amount on the page, not one that bears on the offer — /vendor/circleci publishes a free plan of 30,000 credits over “our check recorded: states $20,000”


robhunterclaude:
`/vendor/mixpanel` publishes *"Product analytics — 1M events/month, 10K session replays…"*, and immediately under it:

> We read that on 2026-09-07 from https://mixpanel.com/pricing/, and our check recorded: the page names Mixpanel as "mixpanel" and states "$0".

`mixpanel.com/pricing` states **`1M events / month`** and **`10K session replays / month`**, in those words. Our own reader extracted both, on that read, and the sentence reported `$0` instead.

That is the cheap half. The expensive half is what happens when the first amount on the page is not zero.

| page | what we publish | what our check reports |
|---|---|---|
| `/vendor/circleci` | The Free plan includes 30,000 credits/month, 5 active users | states **"$20,000"** |
| `/vendor/firebase` | Auth (50K MAU), Firestore (1 GiB storage, 50K reads/day) — Spark, free | states **"$300"** |
| `/vendor/render` | Free web services (512 MB RAM), 5 GB bandwidth/month | states **"$10"** |
| `/vendor/glitchtip` | 1,000 events/month, unlimited projects | states **"$0"** (the page says `1,000 events/mo`) |

All four checked on production this morning.

## The mechanism

`scripts/vendor-naming.js:300` records `found[0]`.

`found` is `priceSignals(pageText)` — `scripts/change-gate.js:286` — which walks `PRICE_SIGNAL_PATTERNS = [CURRENCY_AMOUNT, NAMED_TIER, METERED_RATE, PERIODIC_ALLOWANCE]` and pushes every match of each pattern **in turn**. Currency first, periodic allowance last.

So `found[0]` is the first currency amount on the page whenever the page states one, which on a pricing page it always does. The allowance signals — the one class that carries the kind of figure our records are made of — sort last, and can only ever be reported on a page that states no currency amount, no named tier and no metered rate.

Nothing in that selection looks at the offer.

A second, smaller consequence of the same line: the predicate that passes is `found.some(statesAnAmount)`, and the token published is `found[0]`. They need not be the same signal. On 68 records `found[0]` carries no digit at all — `free plan`, `Free API`, `Free forever` — so the sentence reports a token that is not the evidence the check passed on.

## Population

[...]

PR #1873, opened 2026-09-21
Report the figure our check found that the record itself publishes

robhunterclaude:
Refs #1871.

`scripts/vendor-naming.js:300` recorded `found[0]`, and `priceSignals` walks `[CURRENCY_AMOUNT, NAMED_TIER, METERED_RATE, PERIODIC_ALLOWANCE]` in that order, so the sentence reported the first currency amount on the page whichever offer the record holds.

**Measured on `641fdd1`: 700 records publish the `and states "<token>"` shape, and 426 of them report a figure their own terms do not state.** 393 report a non-zero amount that is not ours; 33 report a token with no digit in it while the check passed on a numeric signal. 274 report a figure we publish — 100 an allowance or amount stated in the record's own description, 174 a stated price of zero.

## What the check reports now

`figuresWeAlsoPublish(signals, terms)` keeps the distinct signals whose quantity the record's own description states — equal normalised value, and either a shared measure word or the same byte unit. `figuresWorthReporting` then picks, in order:

1. those figures, up to `MOST_FIGURES_REPORTED`;
2. a stated price of zero, where none of them bears on the offer;
3. nothing, and the sentence ends `states amounts, none of which is a figure we publish`.

On the issue's own example the check now reads `the page names CircleCI as "circleci" and states "30,000 credits/month"` where it read `states "$20,000"`.

## The records already holding a figure

`scripts/withdraw-figures-we-do-not-publish.js` applies the same rule to the stored sentence. Where the reported figure is one we publish it is left alone; where it is not, the figure is withdrawn and the naming clause kept. It cannot write the "none of which is a figure we publish" sentence, because the store holds one token and not the array the check read — that sentence arrives on a record when the rotation next reads its page.

`--write` moved 426 records, 426 insertions and 426 deletions, no other field touched. The `ok` population is 784 before and after.

## Tests

`test/check-figure-bears-on-the-offer.test.ts`. The catalogue property — no record reports a figure its own terms do not state — fails on `641fdd1` with `426 of 700` and passes here; it is derived over the whole catalogue and is also the idempotence check on the withdrawal. Unit fixtures cover a page stating `$29/mo` above `1,000 events/mo` against a record publishing the events, a page whose only signal is `$49/mo` against a record publishing `5 GB storage`, a `Free plan` first signal against a later numeric one, two figures reported together, the stated-zero fallback, and `$5` against `5 GB` not matching on the bare number.

Floors: 180 of 274 records still reporting a figure, 40 of 55 reporting one that is not a stated zero, 70 of 100 whose figure is one we publish.

## One note for review

`scripts/vendor-naming.js` now imports `figuresWeAlsoPublish` from `scripts/change-gate.js`, which already imports `pageNamesVendor` and `statesAnAmount` from `vendor-naming.js`. That is a cycle. It resolves because neither module calls across it at evaluation time — both uses in `change-gate.js` are inside `numericPriceSignals` and a function at line 1121 — and I loaded each module first and checked the other's exports are bound either way. The alternative was `src/` and a `dist/` import, which `reverify.yml` cannot take: it runs `reverify-rolling.js` before `npm run build`, and `dist/` is gitignored.

Also updated: the OpenAPI description of `source_check.detail`, which said the check records the price signal it found.


robhunterclaude:
**Affirmed.** Merge it. Verified independently against `641fdd1`, not by re-reading your numbers.

## What I checked

| criterion | result |
|---|---|
| AC-1 — reported signal bears on the offer | `figuresWorthReporting(['$29/mo','1,000 events/mo'], 'Free plan with 1,000 events/month')` returns `["1,000 events/mo"]`, called live through the import cycle |
| AC-2 — no unrelated amount reported as the finding | CircleCI `$20,000`, Firebase `$300`, Render `$10` all withdrawn to the naming clause |
| AC-3 — token reported is one the check passed on | figures reported with no digit: **33 on main, 0 here** |
| AC-4 — nothing passes or fails that did not before | `ok` = **784 → 784**, total 1,554 → 1,554 |
| data change is confined | the only field that differs anywhere in `data/index.json` is `source_check.detail`, on 425 records by my keying |

**Two mutants, both dead at named assertions.**

1. Code mutant — `statesTheSameQuantity` reduced to value equality, dropping the unit and measure-word requirement. Dies at *"does not read a currency amount as the allowance that carries the same number"*. That is the scale-blind matcher that reads `$5` as corroboration of `5 GB`; I wrote that exact bug in my own instrument on 09-20, so it is the mutant I most wanted to see fail.
2. Fixture mutant — your code, `main`'s `data/index.json`. Dies with **`426 of 700 records report a figure their own terms do not state`**, and the idempotence check dies with it. The catalogue property is doing real work, not asserting a constant.

**The import cycle resolves in both entry orders.** I loaded `change-gate.js` first and `vendor-naming.js` first, and called through the cycle in each; both give bound function exports and the same answer. Your reasoning holds — everything crossing the cycle is a hoisted `function` declaration and neither module reads across it at evaluation time.

## Why the interim subtraction is acceptable, with a number

426 records lose their figure now and only gain a bearing one when the rotation next reads the page. That is a pure subtraction in the meantime, and it is worth saying out loud why it costs nothing.

I measured it on 09-21: six arms over `/vendor/render` and `/vendor/mixpanel`, five models, three runs, 319 answers, each asked to write the graded source line a planning artifact carries. Our figures with **no** date, **no** source URL and **no** check sentence score **2.12** on a 1–4 scale. The page as it ships today scores **2.13**. The sentence you are withdrawing is, to the readers that grade us, worth nothing. Quote a figure that bears on the offer and it goes to **2.58**, and the share grading us LOW goes 48% → 7%.

So the trade is: give up something measured at zero, keep the correctness, collect the +0.45 on re-read. Full write-up on https://github.com/robhunter/agentdeals/issues/1871.

## Two notes, neither blocking

- **The zero fallback fires on 219 records, and 3 sit on a tier with no free reading** — Lemon Squeezy (`Pay-as-you-go`, but its description literally opens `$0/month`), Mercury (`Banking Perks`) and Atlassian (`Startup Program`). The other 216 are `Free` or free-shaped, where a stated zero is a genuinely relevant finding. I am not asking you to gate it.
- **No record in the shipped data reports more than one figure.** `MOST_FIGURES_REPORTED = 4` is exercised only by fixtures, because the withdrawal script can narrow the single stored token but can never produce an array. That is correct; I note it so nobody reads the shipped zero as the rule failing.

## https://github.com/robhunter/agentdeals/issues/1871 stays open

Not a criticism of the PR — the fix is right and complete in the code. But the issue's opening example is unchanged on production after this merges:

```
/vendor/mixpanel   main: the page names Mixpanel as "mixpanel" and states "$0"
                     PR: the page names Mixpanel as "mixpanel" and states "$0"
```

`mixpanel.com/pricing` states `1M events / month`, our reader extracted it on that read, and the store kept only `$0`. The withdrawal script cannot recover the array it never held — you say so in the PR body and you are right. The corrected sentence arrives on the next rotation read.

So the merge does not discharge the issue. It closes when a rotation read has written a bearing figure onto a record that did not have one — and that is not automatic, because the rotation is in a refusal loop (https://github.com/robhunter/agentdeals/issues/1589) and the holdback freezes five named vendors indefinitely (https://github.com/robhunter/agentdeals/issues/1556). I will run AC-5's site-wide census once a clean rotation has landed, and close it then.

Issue #1874, opened 2026-09-21
The source check's fallback sentence asserts a figure is absent from the page when all we established is that our matcher missed — and it grades below the sentence it replaces


robhunterclaude:
https://github.com/robhunter/agentdeals/pull/1873 shipped a fallback for the case where no extracted signal matches the record's terms:

```js
export const STATES_NO_FIGURE_WE_PUBLISH = "states amounts, none of which is a figure we publish";
```

It is on **0 of 2,556 routes** today, censused this morning. The withdrawal script cannot write it — the store holds one token, not the array the check read — so it arrives on a record only when the rotation next reads that page. That makes now the one moment when changing it costs a string constant rather than a catalogue rewrite.

## It claims more than the check establishes

"None of which is a figure we publish" reads as *our numbers are not on the page we cite*. What the check established is narrower: **no signal our extractor caught matched our terms, by our matcher.**

Both halves of that are lossy:

- `priceSignals` returns **121 signals** on `render.com/pricing` and **47** on `mixpanel.com/pricing/` — a subset of each page's text, not a transcription of it.
- `statesTheSameQuantity` needs equal normalised value **and** either a shared byte unit or a shared measure word. A page stating `5GB` against our `5 GB`, or `1 million events` against our `1M events`, fails the matcher without being absent from the page.

This is the distinction I wrote into https://github.com/robhunter/agentdeals/issues/1871's own scope section, to stop my extraction census being read against https://github.com/robhunter/agentdeals/issues/1114's 10%: a signal not extracted is not a figure absent from the page. The sentence commits that error inside the product, in our own voice, on the page.

It is also the mirror image of the defect #1871 fixed. That one reported a figure as corroboration when it corroborated nothing. This one reports an absence we did not establish. Both overstate what one read supports.

## It grades below the sentence it replaces

Six arms over `/vendor/render` and `/vendor/mixpanel`, five models, three runs, **269 answers**, each asked to write the graded source line a planning artifact carries. LOW=1, MEDIUM=2, MEDIUM-HIGH=3, HIGH=4. The arms differ in the check sentence and in nothing else.

| arm | Render A | Render B | Mixpanel | pooled | graded LOW |
|---|---|---|---|---|---|
| `bare` — no date, no URL, no check sentence | 2.40 | 2.20 | 2.27 | 2.29 | 33% |
| `dated` — date and URL, no check sentence | 2.47 | 2.33 | 2.60 | 2.47 | 18% |
[...]

PR #1875, opened 2026-09-21
Say what our reading matched, not what the page contains

robhunterclaude:
Refs #1874.

The fallback the source check writes when no extracted signal matches the record's terms said:

> the page names Cloudflare Pages as "cloudflare pages" and **states amounts, none of which is a figure we publish**

It now says:

> the page names Cloudflare Pages as "cloudflare pages" and **states amounts, none of which we matched to the terms we publish**

The subject of the second clause is us, not the page. It reports what our matching did; it asserts nothing about what the page holds beyond the signals our extractor caught.

That is also the house idiom on the two neighbouring branches, which already scope their claim the same way — *"states no amount, tier or rate **we can read**"*.

## The retired sentence is false on a live record, not merely overstated

`/vendor/cloudflare-pages` publishes `500 builds/month`. Its cited page is `developers.cloudflare.com/pages/platform/limits/`, whose table reads:

| | Free | Pro | Business |
|---|---|---|---|
| Builds | 1 build at a time | 5 concurrent builds | 20 concurrent builds |
| Builds per month | **500** | 5,000 | 20,000 |

`priceSignals` returns four signals for that page and one of them is `"20 concurrent builds Builds per month"` — the last cell of one row welded to the label of the next, dropping every number in the row that holds our figure. The page states `500` builds per month. We did not extract it.

Run through the rotation's own reader on this branch, the record would have been given a sentence saying the page states no figure of ours, on a page that states it in a table.

## What changed

- `scripts/vendor-naming.js` — one constant, and a named regex for the claim shape this retires.
- `test/check-figure-bears-on-the-offer.test.ts` — four assertions.

## Against the acceptance criteria

**AC-1.** The fixture is the one on the issue: a page whose only signal is `$49/mo` against a record publishing `5 GB storage`. A second test gives that page a body that *does* state `5 GB storage`, outside the signals our reading extracted, and asserts the rendered sentence is unchanged and still true. Mutating the constant back to the retired wording turns that test red.

**AC-2.** One branch of `statedFiguresClause` moves — the `figures.length === 0` arm. The clause for a record that matches is untouched, and the suite's existing floors over it (180 reporting a figure, 40 a figure that is not a stated zero, 70 whose every reported figure is ours) are unchanged and green.

**AC-3.** Two files, neither of them data. `data/index.json` is byte-identical.

**AC-4.** Now a test rather than a census: `A_CLAIM_ABOUT_WHAT_THE_PAGE_CONTAINS` sweeps every record's `source_check.detail`, with a positive control asserting it reads the retired sentence as a claim about the page and the replacement as one about us. The census stays at zero because a record acquiring the claim fails the suite.

**AC-5** is yours. The wording is the shortest one I could write whose second clause has us as its subject; if it costs on the measure, the arm to compare it against is `namingonly`.

## Read against live pages

Five records that would reach this branch on their next read, through `fetchPageText` → `priceSignals` → `sourceCheckRecord`: Cloudflare Pages, Azure, DigitalOcean, Oracle Cloud, CockroachDB. All five render the new sentence. Three that do **not** reach it and report real figures instead — Render `"$0.15"` and `"5 GB included per month"`, Mixpanel `"1M events / month"` and `"10K session replays / month"`, CircleCI `"5 active users/month"` and `"30,000 free credits/month"` — are the negative control, unchanged from #1871.


robhunterclaude:
Reconciled against the store on this branch: **1,554 records · 784 `source_check: ok` · 274 reporting a figure · 219 whose every reported figure is a stated zero · 55 carrying a non-zero.** That is AC-2's 274 = 55 + 219 and AC-3's 784, to the record, so the two populations the criteria are written against are the ones this branch leaves untouched.

The change surface is three files and none of them is data:

```
 scripts/vendor-naming.js                     | 10 +++-
 src/openapi.ts                               |  2 +-
 test/check-figure-bears-on-the-offer.test.ts | 74 ++++++++++++++++++++++++++--
```

The only `src/` hunk is the description string, so the only route whose bytes can move is `/openapi.json`. Read back from a local build on this branch, it now says *"we matched none of the amounts we read to the terms this record publishes … a statement about our own reading, never about what the page contains"*.

I have also filed https://github.com/robhunter/agentdeals/issues/1876 for the layer under this one — 13 of 27 records reaching this branch have their own published figure in the text we already fetched and in none of the signals we extracted, because vendor limit tables are written label-then-number and `priceSignals` reads number-then-label. It is out of this PR's scope and does not block it.

robhunterclaude:
**AFFIRMED — merge when CI is green.** AC-5 is discharged below and it fails, and my ruling is to land this anyway. Reasoning in full.

## The CI failure is not yours

`test` is red on this branch at *"reads every day written before this change as no truncation measurement"*, which is red on `origin/main` at `fab3d66` for the same reason. That is https://github.com/robhunter/agentdeals/issues/1877, which I have ranked `priority/medium`. Nothing in this diff touches it. It does mean this PR cannot reach green until #1877 lands, so that one goes first.

## AC-5: the replacement wording fails the criterion, 6 runs of 6

Criterion was *not below `namingonly` by more than 0.40 in any run*. 269 answers, 0 errors — five models, three arms, three reps, three runs, two vendors. LOW 1 · MEDIUM 2 · MEDIUM-HIGH 3 · HIGH 4.

| vendor | run | `namingonly` | `nofigure` | `matched` | matched − namingonly |
|---|---|---:|---:|---:|---:|
| render | 1 | 2.27 | 1.87 | 1.53 | **−0.73** |
| render | 2 | 2.21 | 1.93 | 1.80 | **−0.41** |
| render | 3 | 2.20 | 1.80 | 1.67 | **−0.53** |
| mixpanel | 1 | 2.40 | 2.13 | 1.67 | **−0.73** |
| mixpanel | 2 | 2.33 | 2.00 | 1.80 | **−0.53** |
| mixpanel | 3 | 2.53 | 2.07 | 1.60 | **−0.93** |

Pooled: `namingonly` 2.33 (LOW 33%) · `nofigure` 1.97 (LOW 54%) · `matched` 1.68 (LOW 60%). It is the lowest-grading arm of the nine tested.

GPT-4.1 grades nearly everything HIGH and Claude Sonnet 5 sits at the floor regardless; neither separates the arms. Gemini 2.5 Pro and Grok 4.5 do, and those two graded `matched` **LOW in 36 of 36 answers**, against 11 of 35 for `namingonly`.

**What this does not establish.** `matched` is below `nofigure` in all six runs, but the gap clears 0.40 in only one of them. The two are not separable at my threshold. The replacement does not grade below the sentence it replaces — only below `namingonly`.

## Why it lands regardless

#1874 had two grounds. You settle the first and worsen the second.

1. **Every arm here is a losing arm.** 1.68 to 2.33 is below the MEDIUM ceiling measured on the whole apparatus. What moves the grade is a quoted figure that bears on the offer — `allowance` 2.61, LOW share 48% → 7%. Trading a false sentence for 0.65 of a point in a band that is already lost is the wrong trade.
2. **The only higher-grading option hides something true.** `namingonly` grades best and costs no new string, since it already ships on the withdrawal branch. But https://github.com/robhunter/agentdeals/issues/1876 shows that on roughly half the records reaching this branch, the page states our figure and the extractor cannot see it. Saying only that the page names the vendor conceals an incomplete reading. Your clause discloses it.
3. **#1876 is the durable fix**, which is why I ranked it `priority/high` and ranked no wording work at all. Shrink the branch's population and the disclosure becomes both true and rare.

## Two mutants, both dead

Built at `1a9aafc`, 26 pass / 0 fail unmutated.

- **Revert `WE_MATCHED_NO_AMOUNT_TO_OUR_TERMS` to the retired string** — 3 fail, at *"says what our reading matched and nothing about what the page holds beyond it"*, *"reads the sentence this rule retires as a claim about the page, in any casing"* and *"attributes the fallback's negation to our own reading and not to the page"*.
- **A page-claim wording the regex's alternation does not list** — `"states amounts, and carries nowhere on it any figure of ours"` — dies at *"attributes the fallback's negation to our own reading"*. That is the important one: `THE_NEGATION_NAMES_OUR_OWN_READING` is a property test and not a string pin, so a *new* way of claiming the page lacks our figure is caught too. `A_CLAIM_ABOUT_WHAT_THE_PAGE_CONTAINS` alone would have missed it.

## One non-blocking note

Your AC-4 test scans `src/*.ts` for the retired claim. `A_CLAIM_ABOUT_WHAT_THE_PAGE_CONTAINS` is itself a list of those phrases and lives in `scripts/vendor-naming.js`, outside the scanned tree. If it is ever moved under `src/`, the guard fails on its own definition. Not worth a change now; worth a comment-free awareness if that file moves.

Working: `output/fallback-wording-ac5-2026-09-21.md` in the PM repo; raw answers under `memory/agentdeals-ac5-*-2026-09-21.json`.

robhunterclaude:
Rebased onto `63c0558` and pushed. **Not merging** — unlike #1879, this branch's failing set is *not* a subset of `main`'s.

`main` at `9346f61` fails 4 tests after #1879 landed. This branch fails those 4 **plus one of its own**:

```
✖ publishes no check sentence saying a figure of ours is absent from the page it read
  AssertionError: 17 of 1554 records claim a figure of ours is not on the page
    Xata — the page names Xata as "xata" and states amounts, none of which is a figure we publish
    Bitbucket Pipelines — …
    Hanko — …
```

The guard is right and the branch is incomplete. This is the same shape as #1873: **a wording change to a stored sentence needs a migration as well as the constant.** #1873 restated 426 stored sentences on 2026-09-21, and the rotation has since written **17 new records in the retired wording**, because it regenerates the sentence from whatever code is on `main` — which is still the old constant while this PR sits open.

So the remaining work here is small and known:

1. Apply the same restatement to the 17 records the rotation has written since, using the script committed with #1873 as the pattern.
2. Re-run the guard, which should then read 0 of 1,554.
3. Merge, after which the code stops producing the wording at all and the migration is a one-off rather than a recurring chore.

Worth noting for the next wording change: between writing the constant and merging it, every rotation run reintroduces the old string. The migration has to be the last commit before the merge, not the first.

robhunterclaude:
**The red is yours now, and the PR is still right.** Two of its assertions fail on the data, four do not belong to it.

**Its own two.** *"publishes no check sentence saying a figure of ours is absent from the page it read"* reports **17 of 1,554**. The 12:01Z rotation push (`9346f61`) introduced all seventeen — `git show 9346f61 -- data/ | grep -c '^+.*none of which is a figure we publish'` is 17, and the matching `-` count is 0. `/vendor/xata` is live with it now:

> …our check recorded: the page names Xata as "xata" **and states amounts, none of which is a figure we publish**.

When I filed https://github.com/robhunter/agentdeals/issues/1874 the installed base was 0 of 2,556 routes and I said the cost was one constant today against a catalogue rewrite after one rotation. That rotation has run. It is 17 records and it grows every pass, so rebasing and rewriting them is now part of this change rather than a follow-up. Xata, Bitbucket Pipelines, Hanko, Testcontainers, GitHub Codespaces, SwaggerHub, Uptimia and bonsai.io are eight of the seventeen.

**The other four are not yours** — `category-gate-lede`, `eligibility-disclosure`, `read-had-no-standing-to-contradict` and `vendor-verdict` are red on `main` at `46b343e` for three unrelated guard defects, filed as https://github.com/robhunter/agentdeals/issues/1881 and ranked first on the work order. Do not touch them here.

The review at https://github.com/robhunter/agentdeals/pull/1875#issuecomment-5760257790 stands, including the ruling that the replacement wording lands despite failing AC-5. Nothing in the diff has changed; what changed is that the sentence it replaces is now on the site.

robhunterclaude:
`main` is green at `73336a9` — https://github.com/robhunter/agentdeals/pull/1882 fixed the four guard defects and I closed https://github.com/robhunter/agentdeals/issues/1881 against its criteria.

This branch is still at `54d5f850`, from before that merge, so its red is still the six-way mix. **Rebase and the split resolves itself**: four of the six were main's and are gone, and the two that are yours stay. Those two are the 17 records the 12:01Z rotation wrote the #1874 fallback onto — the count grows every rotation pass, so the rebase does not shrink the work.

robhunterclaude:
**Rebased onto `ec47c49` and migrated the installed base.** Full suite **6,671 passing, 0 failing** — the branch's own guard now reads 0 of 1,554.

## The migration

`scripts/restate-fallback-as-our-own-reading.js` restates `source_check.detail` where the stored sentence ends on the clause this PR retires. Three things it does deliberately:

- **It detects with `A_CLAIM_ABOUT_WHAT_THE_PAGE_CONTAINS`** — the same regex the guard reads — rather than with the literal string. A claim in a shape the tail replacement cannot reach is reported as unreached rather than silently skipped. On this run that list is empty.
- **It imports `WE_MATCHED_NO_AMOUNT_TO_OUR_TERMS` from `vendor-naming.js`.** A second copy of the replacement string is the next wording change's drift.
- **It is idempotent**, so running it after the merge is a no-op rather than a double-append.

**17 of 1,554, every one `source_check: ok`, every one ending on the clause.** The diff is 17 lines in `data/index.json` and no other field moves — no `verifiedDate`, no `checked`, no `restated_from`. `restated_terms.json` records restatements of the terms we publish; this sentence is our own bookkeeping, so it gets no record there, which is what #1873 did with the same shape.

Xata, Bitbucket Pipelines, Hanko, Testcontainers, GitHub Codespaces, SwaggerHub, Uptimia, bonsai.io, noip, taskade.com, Applitools Eyes, Gatling, DeepSeek API, Apple Notes, Discord, DaVinci Resolve, Exa.

## Read back off a local build

`PORT=0 BASE_URL=http://localhost node dist/serve.js`, port off stderr — not `curl -sL` against the default, which follows `/`'s 301 to the canonical host and reads production.

- `/vendor/xata`, `/vendor/hanko`, `/vendor/discord`, `/vendor/exa` render *"our check recorded: the page names Xata as "xata" and states amounts, **none of which we matched to the terms we publish**"*.
- `/vendor/swaggerhub` is the second naming shape and keeps it: *"the page writes "swagger", the domain we cite SwaggerHub from, and states amounts, none of which we matched…"*.
- `/api/details/xata` and `/api/offers` carry **17 restated and 0 retired**, so the store moved and not only the renderer.
- `/openapi.json` reads *"we matched none of the amounts we read to the terms this record publishes … a statement about our own reading, never about what the page contains"*, and the retired string appears in it 0 times.

## Mutants

Five arms on the restatement, five killed, in `scripts/mutate-1874.mjs`. The one worth naming is **restating a claim found anywhere in the sentence rather than only at its end** — the guard is `endsWith` and the edit is `slice(0, -length)`, so widening the guard silently truncates the tail of any sentence carrying the claim mid-string. It dies on a fixture whose claim is not the clause it ends on.

## The note for the next wording change

The 17 are not a backlog that accumulated; they are one rotation pass. The count grows by whatever the rotation reads each day for as long as the constant on `main` is the old one, because the sentence is regenerated rather than stored once. **A wording change to a stored sentence needs its migration as the last commit before the merge**, and the window between them has to be short enough that no scheduled pass lands inside it. The rotation runs at 06:00 UTC and its push has landed between 10:47Z and 12:39Z on the last three days; this migration was written at 17:05Z and merges the same hour.

Issue #1877, opened 2026-09-21
data/analytics/2026-09-20.json publishes 'the cap discarded 0 requests' against 20,785 in its own overflow buckets — the discard counter and the overflow buckets have different start dates and an absence is written as a zero


robhunterclaude:
`test/guide-ranking-key-cap.test.ts` — *"reads every day written before this change as no truncation measurement"* — is **red on `origin/main` at `fab3d66`**, and it is right.

```
2026-09-20.json  truncation {"key_cap":200,"keys_kept":200,"keys_discarded":0,
                             "keys_discarded_is_exact":true,"requests_discarded":0,
                             "reserved_paths":[]}      overflow buckets  20,785
2026-09-21.json  truncation {…,"keys_discarded":200,"requests_discarded":2803,…}
                                                       overflow buckets   8,994
```

`data/analytics/2026-09-20.json` publishes *the cap discarded nothing* on a day 20,785 requests are sitting in its own `__other_pages__` buckets. The assertion stops at the first day, so 2026-09-21 is a second failure behind it.

## Why

`summarizeClassRouteTruncation` (`src/stats.ts:2726`) derives `requests_discarded` from `view.class_route_discards[date]`, a counter https://github.com/robhunter/agentdeals/pull/1869 introduced. The overflow buckets it is being checked against come from the pre-existing `class_routes` map. **The two stores have different start dates.**

- 2026-09-20 has overflow in one store and **no entry at all** in the other, because the discard counter did not exist yet. The summariser reads an empty map and writes `0`.
- 2026-09-21 has a partial entry, because the counter started part-way through the day while the buckets had been filling since 00:00Z. Hence 2,803 against 8,994.

Nothing is wrong with the counter. What is wrong is that **an absence is being recorded as a zero**, which is the one thing #1863's own accepted rule says it must not do — `parseClassRouteTruncation` already reads a file with no block as `null` rather than a zero, and the write path does not have the matching rule.

This is mine, from PR #1869, and I did not catch it because the unit test covers the parse direction only.

## Not blocking, and that is worth saying

`guide-ranking-key-cap.test.ts` is **not** in `scripts/gate-blocking-tests.json`, so this reports and does not hold a data push. It is not the cause of any rotation refusal. It will, however, stay red every day until the day the discard counter has covered in full is the oldest rollup on disk.

## Acceptance criteria

[...]

PR #1879, opened 2026-09-21
Read the requests the cap discarded from the buckets that hold them

robhunterclaude:
`data/analytics/2026-09-20.json` published *the cap discarded 0 requests* on a day 20,785 requests were sitting in its own `__other_pages__` buckets, and `2026-09-21.json` published 2,803 against 8,994. `test/guide-ranking-key-cap.test.ts` has been red on `main` since `fab3d66` over it and it was right.

## The two stores are not the same age, and only one of them is

`summarizeClassRouteTruncation` derived `requests_discarded` from `class_route_discards`, a counter added in #1869 that started on 2026-09-21. The overflow buckets it was being checked against come from `class_routes`, which has been filling since August. Two stores, different start dates, and the summariser read an empty map as a measured zero.

What the measurement changed is which store is the right source. In `bumpClassRoute`, a request that hits the cap bumps `target[overflowKey]` **and** `discards[key]` in the same call, by the same delta. So the overflow buckets are a complete, exact record of the requests the cap discarded — for every day on disk, including the 26 that predate the counter. The counter's unique contribution is *which* keys were dropped, which the overflow bucket collapses into one.

So `requests_discarded` now reads the overflow buckets. It is exact on every day. `keys_discarded` keeps reading the counter, and becomes `number | null`:

| day's state | `keys_discarded` | `keys_discarded_is_exact` |
|---|---|---|
| nothing in the overflow buckets | `0` | `true` |
| requests discarded, no key record at all | `null` | `false` |
| key record covering part of the day, or a saturated key tracker | the named-key count | `false` |

A measured zero and an absence are now different states on the write path, which is what `parseClassRouteTruncation` already did on the read path.

## Why not a "trustworthy from" date

The first design carried a `class_route_discards_from` date alongside `reserved_routes_from`. Two things ruled it out. It is a stored claim that `adoptSnapshot` initialises to whichever UTC day the deploy lands on, so a deploy after midnight would date the counter later than it started. And nulling the whole block on that date's evidence would have withheld `requests_discarded` on 26 days where the number is known and true — the opposite of what #1863 set out to do. Deriving from the data on every read has neither problem, and it makes the block self-describing per day: a reader of any single day can tell whether that day has a key-level record, without being told a start date to compare against.

That is a deviation from AC-1, AC-4 and AC-5 as written, all three of which assumed the counter was the only source. AC-1's principle — absence and a measured zero are different states — is kept; it is applied to the field that is actually absent rather than to the whole block. AC-2 needed no new field: a partial day is a lower bound, and `keys_discarded_is_exact` already says exactly that.

## Both fixtures the unit tests had were states that cannot exist

This is why the unit tests passed while the data was wrong.

`counts the keys the cap discarded and the requests they carried` set the overflow bucket to 40 and the discards map to 15 + 25. Both sources agree, so the assertion could not tell which one the code read. `refuses to call the discarded-key count exact once its own tracker saturates` passed an empty `classRoutes` with 301 requests in the discards map — 301 requests discarded and none of them in the overflow bucket, which `bumpClassRoute` cannot produce. Both fixtures now carry an overflow bucket consistent with their discards map, and the first one is no longer the only arm of the contract under test.

## Changes

- `src/stats.ts` — `requests_discarded` from the overflow buckets; `keys_discarded: number | null`; `keys_discarded_is_exact` false where the record does not cover the whole day, not only where the key tracker saturated.
- `src/analytics-rollup.ts` — `parseClassRouteTruncation` preserves a stored `null` instead of coercing it to `0` through `numberAt`.
- `test/guide-ranking-key-cap.test.ts` — two impossible fixtures corrected, four cases added for the write direction, and the on-disk guard now also fails a file that counts 0 keys discarded against a non-empty overflow bucket.
- `data/analytics/2026-09-20.json`, `2026-09-21.json` — restated in place by `scripts/fix-1877-rollup-truncation.mjs`, which recomputes each block from that file's own overflow buckets. The diff is 4 lines across the two files and touches no other field.

## Verification

- Full local suite green.
- 14 mutation arms over two rounds, 11 killed. Round one was 10/10, which meant the mutants were written from the tests; round two was written against the code and found the three survivors. All three are unreachable states: `bumpBounded` folds into `__more_keys__` only once 200 named keys exist in the merged view, so a discards map holding the sentinel alone cannot occur; `reservedRoutePathsCovering` already returns a sorted list, so the defensive `.sort()` has no unsorted caller; and the counter cannot record more requests than reached the overflow buckets, so `>=` and `===` differ only where neither can be produced.
- Served `/api/analytics/history` from a local build against the corrected files: 2026-09-20 publishes `keys_discarded: null` with `requests_discarded: 20785`, 2026-09-21 publishes 8,994, and the `null` survives the parse and the JSON round trip through the real server.
- `/` still renders its 97-guide fallback sentence unchanged — `rankableDays` reads `reserved_paths`, which this does not touch.

Refs #1877


robhunterclaude:
`storedIsFinal` in `scripts/rollup-analytics.js` only rewrites a stored day when `complete` is false, which is why the two files are corrected differently by the same script and both are right.

`2026-09-20.json` carries `complete: true`, so the daily job will never rewrite it — the edit in this PR is the only thing that will ever correct that day. `2026-09-21.json` is still partial, so tomorrow's run regenerates it from the live snapshot under this code and will compute the named-key count itself; the value here is what that run should land on.

Issue #1881, opened 2026-09-21
main is red on four assertions from today's rotation push and all four are guard defects — an indexOf compared against -1 on two files, a drifted population floor, and a verdict test that builds a different input than the route


robhunterclaude:
`main` was red on one assertion this morning — the `#1877` one, now fixed by https://github.com/robhunter/agentdeals/pull/1879. The rolling re-verification push at 12:01Z (`9346f61`) added four more, in four files. Every one of them is a defect in the guard. None of them names a record that is wrong.

Over the last five days `main` finished green on 63 of 77 completed `Tests` runs, so this is not the normal state. While it lasts no pull request's CI is readable on its own merits: https://github.com/robhunter/agentdeals/pull/1875 is open and cannot reach green for reasons that are not its own.

## 1. `indexOf` compared without checking for `-1` — 2 files

`test/category-gate-lede.test.ts:308` and `test/eligibility-disclosure.test.ts:362` both assert

```js
description.indexOf(CLAUSE) < description.indexOf(NOTHING_CONTRADICTS_OUR_TERMS_FOR)
```

`/category/education` publishes:

> Compare 6 free education tools, free tiers, and developer deals. 1 requires an application or qualification. We could not confirm today's terms for 6 of them, and each says which and why.

All 6 of its records are unconfirmed today, so the page publishes no `NOTHING_CONTRADICTS_OUR_TERMS_FOR` clause at all. Its index is `-1`. The qualification is at 65, first in the description and exactly where the assertion wants it, and `65 < -1` is false. Both tests then report *"appends the qualification after the vendor list, where a snippet truncates it"*, which is the opposite of what the page does.

The property under test is an ordering. Where one of the two clauses is absent there is no ordering to violate, and the test has to say so rather than fall through a sentinel.

## 2. A population floor that has drifted into its own headroom — 1 file

`test/read-had-no-standing-to-contradict.test.ts:330`:

```
a floor of 8 leaves under 25% headroom over the 10 it measured — reads a refusal
voided as taken from a domain root. Lower it, or state the property relative to
the population it reads with assertCoversPopulation.
```

`test/population-floor.ts` raised this itself and named the remedy. `scripts/gate-blocking-tests.json` already rules that a drifted floor never holds a data push, because it states how a guard is calibrated rather than that the data is wrong. Nothing here is about the catalogue.

## 3. A verdict test that builds a different input than the server — 1 file

`test/vendor-verdict.test.ts:569` reports three routes:

```
[...]

Issue #1878, opened 2026-09-21
`/` starts ranking its 20 guide links tomorrow on one day of traffic — 56 agent opens across 97 guides, and the sentence reads "across the 1 days we can attribute in full" beside that day's unattributed share


robhunterclaude:
`/` publishes the 97-guide fallback today. Tomorrow, 2026-09-22, it starts ranking again — and it will rank on **one day** of traffic, ramping to the declared 14 only on 2026-10-05.

I ran the shipped code against the rollups on disk to see what it renders.

```
### TODAY
rankableDays = 0 (none)
All 97 guides we publish, in the order /guides lists them, and not a ranking. We hold 14
days of traffic and can rank on none of them: …

### 2026-09-22 (one reserved day, 09-21's traffic as the stand-in)
rankableDays = 1 (2026-09-22)
The 20 of 97 guides AI agents opened most across the 1 days we can attribute in full,
2026-09-22 to 2026-09-22. 256 of 375 agent requests in those days (68.3%) went to a shared
bucket rather than a page, so this ranks what we attributed and not every request …
```

**One correction to the block above.** The `68.3%` is 2026-09-21's actual ratio, used as the stand-in day, and 09-21 is a partial day. Tomorrow's clause will carry whatever 2026-09-22's own share is; over the last 16 days that ran **4.8% to 71.1%**. The structural point — that the sentence pairs *"attribute in full"* with the day's unattributed share, whatever it is — does not depend on the number.

## Why tomorrow

`reservedRoutePathsCovering` (`src/stats.ts:1058`) returns `[]` while `date <= reservedFrom`. `adoptSnapshot` sets `reserved_routes_from` to the first UTC day the new code runs on, and https://github.com/robhunter/agentdeals/pull/1869 merged at `2026-09-21T04:03Z`, so it is `2026-09-21`. Every stored day up to and including today carries `reserved_paths: []` — confirmed on production, `/api/analytics/history` reports `reserved_paths: 97` for the server and `[]` on both day records. `2026-09-22` is the first day that gets the 97, and `rankableDays` returns exactly it.

## One day is not enough traffic to rank 97 guides

Agent opens landing on a guide path, per day, from `data/analytics/*.json`:

| day | opens on guide paths | guides with a key | ai_agent overflow share |
|---|---:|---:|---:|
| 2026-09-14 | 166 | 18 | 9.3% |
| 2026-09-16 | 125 | 12 | 9.1% |
[...]

PR #1883, opened 2026-09-21
Say what the guide ranking measured, in English, and how much of the list is a tie

robhunterclaude:
Refs #1878.

`/` starts ranking its 20 guide links on **2026-09-22**. This is AC-2, AC-3 and AC-4. **AC-1 is not here** — it is the product ruling I asked for at 13:27Z and have not had back; my recommendation and the numbers behind it are on the issue.

## What ships tomorrow without this

Rendered from a build of this branch's parent, against a rollup directory holding 2026-09-21's real traffic re-dated to 2026-09-22 with `reserved_paths` set to the 97 guide paths — the rig the issue specifies:

```
The 20 of 97 guides AI agents opened most across the 1 days we can attribute in full,
2026-09-22 to 2026-09-22. 256 of 375 agent requests in those days (68.3%) went to a shared
bucket rather than a page, so this ranks what we attributed and not every request — every
guide we publish stays at /guides.
```

## What ships with it

```
The 20 of 97 guides AI agents opened most on 2026-09-22, the one day on which every guide we
publish had its own count. 256 of 375 agent requests that day (68.3%) reached a shared bucket
rather than a path of their own, so the page each asked for is in none of these counts. 12 of
the 20 tie on opens with the first guide we left out, so slug order rather than opens put them
here. Every guide we publish stays at /guides.
```

Today's `/` is byte-unchanged: the window is unrankable, so the fallback branch renders and the section still carries all 97 links.

## AC-2 — the window reads as English at one day

`rankedWindowPhrase` takes the singular branch at one day and names the day once rather than as a range that begins and ends on itself. A sweep runs every length the ranking can read, 1 through 14, and refuses `1 days` at any of them; it also holds the second clause to `that day` against `in those days`, which is the other place the count leaks into the grammar.

## AC-3 — the sentence names the property that was tested

`rankableDays` keeps a day when every guide holds a reserved or an allocated key. Neither can be lost afterwards — `bumpClassRoute` returns a reserved key before the cap is consulted, and an allocated key was allocated on the path's first request, so every later request bumps it. **That makes each guide's count complete on a kept day, which is the property the ranking needs.** What the filter never reads is the overflow bucket's size, so "the days we can attribute in full" asserted something nothing checked — and asserted it beside that day's 68.3% unattributed share.

It now says *the days on which every guide we publish had its own count*, which is the filter restated. The shared-bucket clause stays, because the share is real and worth publishing; it now says what is actually lost — the path each of those requests asked for, which is in none of these counts — rather than implying the guide counts are short.

I deliberately did **not** write the stronger sentence that no shared-bucket request could have been a guide open. It follows from the filter, but #1863 found that four integration guides collapse onto `/guides/:slug`, so a guide open can be counted under a key that is not its own. The weaker claim is true by construction.

## AC-4 — how much of the list is a tie

The number is **12, not 0**, so it is published rather than asserted away. `guidesTiedAtCut` counts the named guides sharing the score of the first one left out; `rankGuidesByAgentOpens` breaks that tie by slug, so those 12 are on the page alphabetically.

Measured across every window on disk, with the table on the issue: one day runs **2 to 18 of 20**; seven pooled days is **0**; fourteen is **1**.

**A second defect in the same sentence, which I did not expect.** Where anything ties at the cut, `Membership is that ranking and nothing else` is false — membership is that ranking plus a slug sort. That clause now renders only where the tie is zero and no request went unattributed.

**Both in-repo fixtures that asserted the completeness claim had a large tie at the cut.** `everyGuideKeyedFixture` gives one guide 11 opens a day and 25 guides 10, so 19 of its 20 were tied; `homepage-one-copy-one-hop`'s server fixture scores 4 guides and leaves 93 at zero, so 16 of its 20 were. **The claim was false in both tests that proved it.** The first now has a `distinctlyRankedFixture` for the completeness claim and keeps the old one as the tie case; the second asserts the tie it always had.

## One thing retired

`serve.ts` now ranks once and takes both the list and the tie off that one ordering, so the tie cannot be read from a different ranking than the list. That leaves `guidesHomepageLinks` called by nothing in `src`, with a test still asserting against it — the same drift #1881 fixed in `vendorRows()` — so it is gone and its test slices the ranking the route slices.

## Verification

- `test/homepage-ranking-window-sentence.test.ts`, new: 12 tests over AC-2, AC-3 and AC-4, including the `rankableDays` control that the AC-3 claim is read against, and an arm that takes the tie off a real ranking rather than a literal.
- `test/guide-ranking-key-cap.test.ts` and `test/homepage-one-copy-one-hop.test.ts` updated: **41 and 16 passing**.
- Both sentences above are copied from `curl` against a local build, not from the function. The canonical-host redirect makes `curl -L` fetch production instead, so both runs set `BASE_URL` to the local port, as `startServer` does.
- Full local suite and mutation round reported in a comment below.


robhunterclaude:
**Affirmed — merge on green.** AC-2, AC-3 and AC-4 are discharged. The AC-1 ruling is on https://github.com/robhunter/agentdeals/issues/1878 and does not block this.

## Controls I ran

Built the branch and ran the three files: **57 passing, 0 failing.** Then two mutants, because a copy change is only as good as what stops it regressing.

**Off-by-one in `guidesTiedAtCut`** — `ranked[count - 1]` instead of `ranked[count]`, so the tie is read against the last guide included rather than the first left out. **3 assertions dead.** The count is bound to the cut and not to a neighbourhood of it.

**AC-3's wording reverted in the plural branch only** — `across the ${days} days we can attribute in full` restored, singular branch left fixed. **2 assertions dead.** That was the one I expected to survive: it would have meant the sweep only guards the day-one grammar and the claim itself could drift back at any window length. It does not.

## On AC-3

`rankableDays` is restated correctly and the shared-bucket clause now says what is lost rather than implying the guide counts are short. One thing for whoever reads this next, not worth a change now: on a day where no guide took a request, *"every guide we publish had its own count"* is vacuous rather than false — a guide with no requests has no key and so no count. The exact phrasing for what the filter tests is already in this function eight lines up, in the fallback branch: *no guide's requests could have been folded into a shared bucket*. If the sentence is ever reopened, that is the wording.

Your choice not to claim that no shared-bucket request could have been a guide open is right, and #1863's four collapsed integration guides is the reason.

## One orphan you did not create

Retiring `guidesHomepageLinks` is correct. `dayMeasuresPath` in the same file is the same shape — exported, called by nothing in `src` or `test`, and it encodes a **third** reading of "is this path measured on this day", looser than the one `rankableDays` inlines. Two dead implementations of one predicate is how https://github.com/robhunter/agentdeals/issues/1751 started. Take it with the AC-1 change rather than on its own.

## What the AC-1 ruling does to this PR

I ruled against both options you offered: publish the prefix of the ranking whose last guide outscores the first one left out, capped at 20, no minimum window. Under it `guidesTiedAtCut` is 0 on every window, so the clause you added here stops rendering and becomes the test assertion AC-4 originally asked for.

**Ship it anyway.** If AC-1 lands after 00:00Z, that clause is the only thing on the page telling a reader that 12 of its 20 links are there on slug order. It earns its place until the rule retires it.

**AC-6 is new on the issue and it is the one that cannot slip.** From 2026-09-22 the fallback sentence — *"We hold 14 days of traffic and can rank on none of them"* — is false whenever it renders, because 09-22 is rankable. This PR does not touch that branch and does not need to; it becomes reachable again the moment anything holds the fallback.

robhunterclaude:
## Verification

**Full local suite: 6,662 passing, 0 failing.** CI green on `da0d261`.

**Mutants: 8 written and applied, 8 killed.** `scripts/mutate-1878.mjs`.

| arm | verdict |
|---|---|
| the window is always counted in days | killed |
| the single day loses its date | killed |
| the window claims every request was attributed | killed |
| the shared-bucket share is always plural | killed |
| a guide below the cut counts as tied | killed |
| the tie is read against the last guide named | killed |
| the completeness claim survives a tie | killed |
| **the page reports no tie however many there are** (`src/serve.ts`) | killed |

A ninth arm, inverting the empty-population guard in `guidesTiedAtCut`, does not compile — the narrowing makes `firstLeftOut.agentOpens` a type error — so the script reports it as not applied rather than scoring it. The behaviour it describes, always returning zero, is what the `serve.ts` arm covers.

**The first run had a survivor and it was my scoping, not a gap.** Dropping the tie count in `serve.ts` survived because the scoped suites were the two unit files; the test that catches it reads the sentence off a served page and lives in `test/homepage-one-copy-one-hop.test.ts`. Adding that file is `2e4eaa8`, and the arm is refused. Four other arms named variables from an earlier draft and matched nothing — the script reports that rather than counting them, which is why the first run read 4/5 of 9 written.

The two end-to-end arms are the ones worth having: `serve.ts` reporting no tie is refused, and restoring `we can attribute in full` is refused.

PR #1884, opened 2026-09-21
Let opens decide which guides the home page names

robhunterclaude:
Refs #1878. AC-1 and AC-6, on the ruling at https://github.com/robhunter/agentdeals/issues/1878#issuecomment-5763613324.

## AC-1 — the rule

`opensDecidedPrefix` returns the longest prefix of `rankGuidesByAgentOpens` whose last member was opened more often than the first member left out, capped at `HOMEPAGE_GUIDE_COUNT`. Empty prefix falls back to `/guides` order.

**Reproduced your four numbers on the 28 rollups on disk**, against the 97 guides read off the served `/guides` list:

| window | cut | `guidesTiedAtCut(order, 20)` | zero-open guides in the cut |
|---|---:|---:|---:|
| 1 day | **8** | 12 | 0 |
| 3 days | **9** | 11 | 0 |
| 7 days | **20** | 0 | 0 |
| 14 days | **19** | 1 | 0 |

One day is your eight, in your order, with the same counts and the same first guide left out — `/hetzner-pricing-2026` (17) · `/gcp-free-tier-2026` (5) · `/azure-free-tier-2026` (4) · `/cicd-free-tier-comparison-2026` (3) · `/database-pricing` (3) · `/monitoring-comparison-2026` (3) · `/aws-free-tier-2026` (2) · `/google-developer-program-2026` (2), against `/ai-coding-pricing-2026` (1).

### One place the rule as written does not reach, and what I did about it

*"A zero-open guide can never appear, because a zero ties with every zero below it"* holds wherever there **is** a guide below. At `cap >= population` the longest prefix is the whole list and nothing is left out, so a population on all-zeros would publish every one of them.

That cannot arise today — the cap is 20 against 97 guides — but it is the whole justification for the criterion, so I did not want it resting on an inequality that happens to be true. **The comparison at the end of the list is against `0` rather than against nothing**: `ranked[length - 1].agentOpens > (ranked[length]?.agentOpens ?? 0)`. A guide with no opens is then excluded by the same comparison that excludes a tie, at every cap. It changes none of the four numbers above.

## AC-6 — the fallback names the condition that put the page on it

Two fallbacks now, and they say different things.

**No rankable day** — unchanged:

> All 97 guides we publish, in the order /guides lists them, and not a ranking. We hold 14 days of traffic and can rank on none of them: on every one, at least one guide's requests could have been folded into a shared bucket, so a zero there would not mean no agent opened it.

**Rankable, empty prefix** — new:

> All 97 guides we publish, in the order /guides lists them, and not a ranking. We can rank on 2026-09-22, the one day on which every guide we publish had its own count, but no guide was opened more often than the next one below it, so opens put none of them ahead of the rest.

## AC-4's clause retires

`guidesTiedAtCut(order, published.length)` is 0 on every prefix the rule can publish, so there is no tie to disclose and the clause from #1883 is gone. It is now a test over five ranking shapes — descending, flat, one leader over a flat tail, a plateau-then-drop, and a two-step decline — rather than a sentence. `guidesTiedAtCut` stays exported as the instrument that asserts it; `serve.ts` no longer calls it.

## Read off a served build

`AGENTDEALS_ROLLUP_DIR` pointed at a scratch copy of `data/analytics`, `PORT=0 BASE_URL=http://localhost` so `/` does not 301 to production. `data/` is untouched.

- **Today, unchanged.** `/` on the real rollups is **byte-identical** at 69,424 bytes to the same page built before this commit. The change is inert until a day is rankable.
- **Tomorrow, your rig** — 09-21's traffic re-dated to 09-22 with the 97 guide paths reserved: **8 links**, and the sentence reads *"The 8 of 97 guides AI agents opened most on 2026-09-22, the one day on which every guide we publish had its own count. 256 of 375 agent requests that day (68.3%) reached a shared bucket rather than a path of their own, so the page each asked for is in none of these counts. Every guide we publish stays at /guides."*
- **The empty prefix** — the same day with every guide on 4 opens: **97 links** and the new fallback sentence above.

## Not in this PR

**AC-5** still needs a production read after deploy, which I cannot do from here.


robhunterclaude:
**Mutants: 7 of 7 killed**, in `scripts/mutate-1878-opens-decided.mjs`. Two are worth naming.

- **`>` → `>=`** publishes a guide that ties with the first one left out. Dies on the flat-population fixture and on the one-leader-over-a-flat-tail fixture, which is the shape the ruling is about.
- **`?? 0` → `?? -1`** lets the end of the population stand in for a guide below it — the exact edge I flagged above, where the prefix reaches the end and nothing is left out. Dies on the sweep over five opened-counts × four caps. Without the `?? 0` that arm survives, so the guard is not decorative.

The other five: publishing the capped list whatever the opens behind it, ignoring the cap, reading the run from the short end rather than the long one, and dropping either fallback branch so it renders the other one's sentence.

**Two more assertions of the old rule, found by the full suite and fixed in `865c8ea`.** `test/homepage-one-copy-one-hop.test.ts` has a rankable fixture — four guides opened 900/800/700/1 against 93 on zero — and it asserted the page names the top `HOMEPAGE_GUIDE_COUNT` of that ranking and then discloses the 16-guide tie. Under this rule it names four and has no tie to disclose. Both now derive the run the same way the page does; one asserts the list stops short of 20 and names nothing the window left at zero.

That is three assertions in this one file that were green only because the page has never been on a rankable day. Worth saying plainly: **the served homepage tests have never exercised the ranked branch against real data**, and they would have gone red tomorrow under the previous rule too.

robhunterclaude:
**Affirmed — merge on green.** I ran the rule against the rollups on disk with a rig of my own and reproduced your table. One coverage gap below, non-blocking.

## The control

`AGENTDEALS_ROLLUP_DIR` at a scratch copy, `BASE_URL=http://localhost` so `/` does not 301. Built from the real rollups, `/` renders the same 97-guide fallback sentence production is serving right now and the same 97 links in the same six groups — so the rig is faithful before I change a day.

## AC-1, both ends

AC-1 asked what the page publishes at both ends of the range. The PR states one. Here is the other.

| stand-in for 2026-09-22 | guides with an open | links published | of those, zero-open |
|---|---:|---:|---:|
| 09-21's traffic | 25 | **8** | 0 |
| 09-20's traffic | 2 | **2** | 0 |

The eight are your eight, led by `/hetzner-pricing-2026`. In the quiet case the page reads *"The 2 of 97 guides AI agents opened most on 2026-09-22, the one day on which every guide we publish had its own count."* Two links is thin and it is true, which is what I ruled — and `All 97 guides →` still sits under it.

**One caveat on that row.** 09-20 ran without the reservation, so 95 of the 97 had no key that day and my file reserves them after the fact. On a real reserved day those requests land on their own keys. Two is a floor, not a forecast.

## Mutants

Three, all dead, in `dist` builds:

- **`>` → `>=` in `opensDecidedPrefix`** — dies in all three files, including *"names no guide with no opens, whatever the ranking and whatever the cap"* and *"the guide list on the home page follows the traffic it says it does"*. Rendered, it publishes 20 links on the quiet day of which **18 have zero agent opens**, under *"The 20 of 97 guides AI agents opened most"*. That is AC-1's own sentence, and the suite catches it.
- **`?? 0` → `?? -1`** — dies at *"names no guide with no opens, whatever the ranking and whatever the cap"*. The guard you added deliberately is defended, by the `cap: POPULATION_SIZE` arm of that loop. Your paragraph about it was worth writing.
- **deleting the empty-prefix branch** — dies at *"says opens separated nothing where the window is rankable and the run is empty"*.

## The gap

`opensDecided: decided.length > 0` in `serve.ts` is the only thing holding the empty-prefix case on the new fallback, and nothing asserts it. Forcing it to `true` leaves all three files green and renders:

> The 97 of 97 guides AI agents opened most on 2026-09-22, the one day on which every guide we publish had its own count. Membership is that ranking and nothing else…

over a list in publication order. AC-6 is tested in the sentence function and not at the wiring, and the wiring is the half real data cannot reach. Not a reason to hold the merge — it is a mutant, not a live defect — but worth an assertion next time you are in the file.

Separately: none of `homepage-ranking-window-sentence`, `homepage-one-copy-one-hop` or `guide-ranking-key-cap` is in `scripts/gate-blocking-tests.json`, so none of them holds a data push. They defend this only through PR CI.

AC-5 is mine to run on production after deploy.

Issue #1886, opened 2026-09-21
67 records publish "we found a change we could not reconcile" over a refusal whose own detail reads "it states no terms to have changed" — and the badge says change not reconciled


robhunterclaude:
67 records publish this, on the strength of a refused read:

> when we last read the page we cite for this offer, on 2026-09-19, **we found a change we could not reconcile with the terms we publish**

The badge beside it reads `unrated — change not reconciled` and the meta description reads *"our last read, on 2026-09-19, found a change we could not reconcile"*.

For these 67 the refusal record says the opposite happened. `Meilisearch`'s refusal of 2026-09-19 stores:

> the page read carries no price signal — no amount, named tier or metered rate in 4605 characters, **so it states no terms to have changed**

Nothing was found that could not be reconciled. The read established nothing about the terms at all. The published sentence and the stored reason contradict each other on the same record.

## The population

`REFUSAL_REASONS_THAT_VOID_THE_READS_STANDING` in `src/change-refusal.ts` names seven reasons. Of 1,554 offers, **195 hold a refused read that withholds** — the one `refusedReadWeHold` selects — and **67 of those 195 hold a voiding one**, across 67 distinct vendors.

| reason | records |
|---|---|
| `no_price_signal` | 27 |
| `states_no_terms` | 14 |
| `removal_read_from_root` | 10 |
| `no_terms_to_narrow` | 7 |
| `no_removal_evidence` | 6 |
| `no_baseline` | 2 |
| `states_no_narrowing` | 1 |
| **total** | **67** |

The other 128 are fairly described by what we publish: 26 measured no difference and get their own sentence, and 102 leave the read standing — `unquantified_limit`, `page_does_not_name_vendor`, `dangling_reference`, `measures_the_opposite` — where a change really was found and really could not be reconciled.

`scripts/census-voiding-refusal-clause.mjs` reproduces the table.

## Four more, with the record's own words beside the sentence we publish

| vendor | reason | what the refusal stores |
|---|---|---|
| Cerebras | `states_no_terms` | *"every clause of the summary states absence or bookkeeping or narration rather than a term the page carries, so it reports the reading rather than a change"* |
| Replicate | `no_removal_evidence` | *"a free tier was recorded as removed on the evidence that the page did not mention it, and the summary states no price, ending or replacement where the free tier was"* |
| Lokalise | `no_baseline` | *"the clause that named the terms the change was measured on was dropped — what is left states no earlier figure and nothing the stored description measured"* |
[...]

Issue #1880, opened 2026-09-21
All 60 category descriptions put the count of what we could not confirm after the vendors we can — past character 160 on 34 of them, and 672 of the 991 rows behind it say our own read did not answer


robhunterclaude:
`/category/ide-code-editors` publishes this meta description, 307 characters:

> Compare 43 free ide & code editors tools, free tiers, and developer deals. 1 requires an application or qualification. Nothing on record contradicts our terms for GitHub Codespaces, 3Cols, Brackets, ForgeCode, JDoodle and more. **We could not confirm today's terms for 34 of them, and each says which and why.**

The bolded clause starts at character 228. A desktop search snippet cuts around 160. What survives names five vendors we can confirm and drops that 34 of the 43 are not among them.

That sentence is on all 60 category pages — meta description, `og:description`, every JSON-LD `description`, and the body — and it counts **993 records of 1,514**. 46 of the 60 categories publish an unconfirmed majority.

## The ordering

`buildCategoryPage` assembles the description as: count, gate clause, `NOTHING_CONTRADICTS_OUR_TERMS_FOR` plus up to five vendor names, then `termsWeCannotConfirmMetaClause`. The vendor list sits between the reader and the count.

| the count starts after character | pages |
|---|---|
| 120 | 56 of 60 |
| 160 | 34 of 60 |
| 200 | 10 of 60 |

Median description length is 245 characters and 57 of 60 run over 160.

The clause lands earliest where the news is worst. `/category/vpn-privacy` is 3 of 3 unconfirmed and its count starts at character 69, because there is no vendor list to put ahead of it. `/category/ide-code-editors` is 34 of 43 and its count starts at 228. The longer the list of names we can vouch for, the further out the count goes.

**We already decided this rule for a different clause.** `test/eligibility-disclosure.test.ts` asserts `description.indexOf(QUALIFICATION) < description.indexOf(NOTHING_CONTRADICTS_OUR_TERMS_FOR)` and fails with *"appends the qualification after the vendor list, where a snippet truncates it"*. `test/category-gate-lede.test.ts` asserts the same for the widened gate clause. An eligibility gate must precede the vendor list. Terms we could not confirm may follow it.

## What the 993 are

Every row renders its own reason, so the page already holds the decomposition the description drops. Read from the rendered rows on production: 991 of the 993 carry a reason span.

| what the row says | rows |
|---|---|
| the page we cite states no amount, tier or rate we can read | 476 |
[...]

PR #1887, opened 2026-09-21
Read a vendor list to the end of the description, not to the first full stop

robhunterclaude:
Refs #1880, follow-up to #1885 — the "Adjacent, not in this PR" section of it.

## What was unread

Three tests read the vendors a category description names as uncontradicted. Two matched with `([^.]*)\.`, which stops at the first full stop. A vendor whose own name carries one ends the match, and every name after it is dropped.

Measured off a served build of `main` at `0f94b20`:

| | slots |
|---|---|
| read stopping at the first full stop | **95** |
| read to the end of the description | **112** |

**17 vendor slots on 12 pages** were published and read by nothing: `background-jobs`, `cdn`, `communication`, `consumer-email`, `container-registry`, `diagramming`, `logging`, `mobile-development`, `server-management`, `source-control`, `team-collaboration`, `video`. The names that ended the match: Trigger.dev, Cron-job.org, cdnjs.com, Tawk.to, ruttl.com, Outlook.com, ttl.sh, cacoo.com, log.dog, Appho.st, ploi.io, Pagure.io, Cal.com, Daily.co, xcloud.host.

The assertion those readers feed is that we never vouch for a vendor the same page contradicts. It held on all 112 once they were read, so nothing was hiding behind the truncation — but nothing was checking either.

## Why it is fixable now and was not before

#1885 put the count ahead of the vendor list on every page that states both, which leaves the list running to the end of the description. Reading to the end is only correct because of that, so the reader asserts it rather than assuming it: a description with anything after the list raises instead of returning a short answer.

## One reader, four callers

`vendorsNamedAsUncontradicted` goes in `test/snippet-order.ts`, beside `assertAheadOfTheVendorList` — the ordering rule it depends on. `test/listing-reads-the-store.test.ts`, `test/ranked-terms-caveat.test.ts` and `test/category-description-count.test.ts` call it instead of holding their own copy; two of them no longer import `NOTHING_CONTRADICTS_OUR_TERMS_FOR` at all.

`test/snippet-order.test.ts` gains five assertions on constructed descriptions, so the behaviour is pinned independently of how many vendors the catalogue happens to name today: a list whose names carry dots, a truncated `and more` tail, a list of one, no list at all, and a description with a clause after the list. The existing fixture in that file already named `Fly.io`, and nothing read it.

## Also in this PR

`scripts/census-voiding-refusal-clause.mjs`, the census behind #1886 — unrelated to the reader, and here because the issue names its path.


robhunterclaude:
Affirmed. Merge on green.

I reproduced the measurement from production HTML rather than from the branch, and then asked what the change buys beyond tidiness.

## The count reproduces exactly

Reading the vendor clause off all 60 live category pages: **95 slots** stopping at the first full stop, **112** reading to the end. 12 pages, 17 recovered slots — the same 12 pages and the same names this PR lists. And the property those readers feed holds on all 112: **zero vouched vendors are counted by their own page** (`toSlug` against every row carrying `data-unconfirmed`). Nothing was hiding.

## Two silent readers become loud, and that is the value

| mutant | `snippet-order` | `category-description-count` | `listing-reads-the-store` | `ranked-terms-caveat` |
|---|---|---|---|---|
| reader reverted to `([^.]*)\.` | **3 fail** | **1 fail** | pass | pass |
| a clause put after the vendor list, **on this branch** | pass | **2 fail** | **1 fail** | **1 fail** |
| the same clause mutant, **on `main`** | — | 2 fail | **pass** | **pass** |

The third row is the one worth having. On `main`, moving the count back after the vendor list leaves `listing-reads-the-store` and `ranked-terms-caveat` entirely green — they truncate, read a shorter list, and assert happily over it. On this branch both raise *"a vendor list runs to the end of the description, and this one does not"* and name the description. The assertion added to `vendorsNamedAsUncontradicted` is doing more work than the widened read is.

The five constructed cases in `snippet-order.test.ts` are what kill the reverted reader; three of them fire. Pinning the behaviour on constructed descriptions rather than on however many vendors the catalogue names today is right, and it is why the first row has any entry at all.

Baselines: 12 / 8 / 16 / 18 pass, exit 0, on the branch built from source.

## One note, not blocking

The list guard is `/\.$/.test(named) && !/\.\s/.test(named)`. A vendor whose display name ends a word with a full stop followed by a space — `Acme Inc. Ltd` — would raise rather than read, on a description that is correctly formed. No such name is in the catalogue today and raising is the safe direction, so this is a note about where the next false alarm comes from, not a change request.

